PR #195, opened 2022-03-08
feat(rn): release notes kots 1.66.0

DanStough:


docs/release-notes/rn-app-manager.md
```diff
@@ -5,6 +5,26 @@ For release notes earlier than v1.60.0, see
...
 
+## 1.66.0
+
+Released on March 8, 2022
+
+Support for Kubernetes: 1.21, 1.22, and 1.23
+
+### New Features
+* Adds the ability to exclude the applications or the KOTS admin console from full snapshot restores using the [kots restore](/reference/kots-cli-restore-index) CLI command.
+* Adds the ability to display the command to restore only the KOTS admin console from a full snapshot on the “Full Snapshots” page.
```
JustinN-Replicated, line 16: * Adds the ability to display the command to restore only the KOTS admin console from a full snapshot on the “[Full Snapshots](/kotsadm/snapshots/overview/#full-snapshots-recommended)” page.

docs/release-notes/rn-app-manager.md
```diff
@@ -5,6 +5,26 @@ For release notes earlier than v1.60.0, see
...
+
+### Improvements
+* Allows gitops to be disabled when it's enabled but has an invalid config. Before this change, you were required to fix the configuration and then disable.
```
JustinN-Replicated, line 19: Is this not a bug? this sounds like a bug. Would suggest something more like the following, though I don't think we have a lot of precedent for before/after in bugs?:

* Allows gitops to be disabled when enablement has failed due to an invalid config. Before this change, you were required to fix the configuration before you could disable this feature.

docs/release-notes/rn-app-manager.md
```diff
@@ -5,6 +5,26 @@ For release notes earlier than v1.60.0, see
...
+* Adds a new flag `--no-port-forward` to the CLI install command to disable automatic port-forwarding. The old flag `--port-forward` has been deprecated.
+
+### Bug Fixes
+* Corrects the placeholder prometheus URL in the admin console dashboard so that it is accurate for kURL clusters.
+* Fixes a bug where the version mismatch warning was printing incorrect messages in some cases.
+* Fixes a bug where the "See details" button in the support bundle analysis page wouldn't show any information about the unhealthy pod.
+* Allow user to re-upload license if application is not installed.
```
JustinN-Replicated, line 26: * Allows a user to re-upload license if an application is not yet installed.

docs/release-notes/rn-app-manager.md
ajp-io: This is referring to the full snapshots page in the admin console, not the docs, so this would be a more accurate update:
```suggestion
* Adds the ability to display the command to restore only the KOTS admin console from a full snapshot on the “Full Snapshots” page in the admin console.
```

ajp-io: Ok. Linking the the part where it says `"Full Snapshots" page` makes it look like that is the page we're talking about. So IMO it would be better to move the hyperlink to the first mention of "full snapshot" a few words earlier.

JustinN-Replicated: That works for me; I was figuring it just makes sense to give them some sort of guidance to where we're talking about (instead of presuming they're familiar enough with the admin console to know that is)

JustinN-Replicated: Just committed that change so resolving this thread.

ajp-io (approved):

docs/release-notes/rn-app-manager.md
```diff
@@ -12,18 +12,18 @@ Released on March 8, 2022
...
+* Adds the ability to display the command to restore only the admin console from a [full snapshot](/enterprise/snapshots-understanding#full-snapshots-recommended) on the Full Snapshots page in the admin console.
 
 ### Improvements
 * Adds the [`--no-port-forward`](/reference/kots-cli-install#usage) flag to the `kots install` command to disable automatic port-forwarding. The old `--port-forward` flag has been deprecated.
 
 ### Bug Fixes
 * Corrects the placeholder Prometheus URL in the admin console dashboard so that it is accurate for embedded installations.
-* Fixes a bug where the warning message indicating a mismatch between the KOTS API and CLI versions printed incorrect messages in some cases.
-* Fixes a bug where the "See details" button on the support bundle analysis page wouldn't show any information about an unhealthy pod.
+* Fixes a bug where the warning message indicated a mismatch between the Vendor API and CLI versions printed incorrect messages in some cases.
```
ajp-io, line 23: @jonquil2002 This isn't exactly accurate after your change. There is still a warning message that indicates whether there is a mismatch between the API and CLI versions. It's just that this message would previously print a wrong message sometimes. The change to "indicated" makes it sound like this message is no longer present/valid.

Also, I believe this is in reference to the KOTS API and not the Vendor API.

jonquil2002 (approved):
A few tweaks. lgtm

PR #200, opened 2022-03-08
44393 adds semver topics

paigecalvert:
SC cards:
- Enabling semver content: https://app.shortcut.com/replicated/story/44393/content-diff-update-semver-docs
- Adds semver replicated CLI commands: https://app.shortcut.com/replicated/story/44507/migrate-replicated-cli-semver-commands

https://deploy-preview-200--replicated-docs.netlify.app/vendor/releases-understanding

https://deploy-preview-200--replicated-docs.netlify.app/vendor/releases-semantic-versioning

https://deploy-preview-200--replicated-docs.netlify.app/reference/replicated-cli-channel-enable-semver and https://deploy-preview-200--replicated-docs.netlify.app/reference/replicated-cli-channel-disable-semver

### DIFF
--- a/docs/reference/replicated-cli-channel-disable-semver.md
+++ b/docs/reference/replicated-cli-channel-disable-semver.md
@@ -1,0 +1,20 @@
+# channel disable-semantic-versioning
+
+Disable semantic versioning for a channel.
+
+### Usage
+```bash
+replicated channel disable-semantic-versioning CHANNEL_ID
+```
+
+| Flag                 | Type | Description |
+|:----------------------|------|-------------|
+| `-h, --help`   |  |          Help for the command. |
+| `--app string` | |   The app slug or app id used in all calls. **Default**: Uses the `$REPLICATED_APP` environment variable. |
+| `--token string` | |  The API token used to access your app in the Vendor API. **Default**: Uses the `$REPLICATED_API_TOKEN` environment variable. |
+
+### Examples
+```bash
+replicated channel disable-semantic-versioning 1xyB2Mgbg9N7rExShfbdBYIuzeW
+Semantic versioning successfully disabled for channel 1xyB2Mgbg9N7rExShfbdBYIuzeW
+```
--- a/docs/vendor/releases-semantic-versioning.md
+++ b/docs/vendor/releases-semantic-versioning.md
@@ -12,7 +12,7 @@ For existing applications created before February 23, 2022, semantic versioning 
 
 When you enable semantic versioning on a channel, the version label for a release promoted to that channel is verified to ensure that it is a valid semantic version. For more information about valid semantic versions, see [Semantic Versioning 2.0.0](https://semver.org).
 
-For channels with semantic versioning, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
+For channels with semantic versioning enabled, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
 
 If releases that do not use a valid semantic version are already promoted to a channel, the admin console sorts the releases that do have semantic versions starting with the earliest version and proceeding to the latest. For example, assume that you promote these releases in the following order to a channel: 1.0.0, abc, 0.1.0, xyz, and 2.0.0. Then, you enable semantic versioning on that channel. The admin console would sequence the version history as follows for the channel: 0.1.0, 1.0.0, abc, xyz, 2.0.0.
 
--- a/docs/vendor/releases-semantic-versioning.md
+++ b/docs/vendor/releases-semantic-versioning.md
@@ -14,7 +14,7 @@ When you enable semantic versioning on a channel, the version label for a releas
 
 For channels with semantic versioning, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
 
-The admin console sorts releases with semantic versions starting with the earliest version and proceeding to the latest. For example, assume that you promote these releases in the following order to a channel: 1.0.0, abc, 0.1.0, xyz, and 2.0.0. Then, you enable semantic versioning on that channel. The admin console would sequence the version history as follows for the channel: 0.1.0, 1.0.0, abc, xyz, 2.0.0.
+If releases that do not use a valid semantic version are already promoted to a channel, the admin console sorts the releases that do have semantic versions starting with the earliest version and proceeding to the latest. For example, assume that you promote these releases in the following order to a channel: 1.0.0, abc, 0.1.0, xyz, and 2.0.0. Then, you enable semantic versioning on that channel. The admin console would sequence the version history as follows for the channel: 0.1.0, 1.0.0, abc, xyz, 2.0.0.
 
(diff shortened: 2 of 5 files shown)


docs/vendor/releases-semantic-versioning.md
```diff
@@ -0,0 +1,31 @@
+# Enabling Semantic Versioning
+
+Semantic versioning is available with the Replicated app manager v1.58.0 and later.
+
+For applications created in the vendor portal on or after February 23, 2022, semantic versioning is enabled by default on the Stable and Beta channels. Semantic versioning is disabled on the Unstable channel by default.
+
+For applications created before February 23, 2022, semantic versioning is disabled by default on new and existing channels.
```
ajp-io, line 7: The new channels part is actually true for all applications regardless of when they were created. Any new channel doesn't default to having semver enabled.

docs/vendor/releases-semantic-versioning.md
```diff
@@ -0,0 +1,31 @@
...
+
+## About Semantic Versioning
+
+When you enable semantic versioning on a channel, the version label for a release promoted to that channel is verified to ensure that it is a valid semantic version. For more information about valid semantic versions, see [Semantic Versioning 2.0.0](https://semver.org).
+
+For channels with semantic versioning, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
```
ajp-io, line 13: IMO it's important to say that semantic versioning is enabled, since a vendor could use semantic versioning for their version labels, but they still wouldn't get these benefits/behaviors unless they enabled our semantic versioning support for the channel.
```suggestion
For channels with semantic versioning enabled, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
```

docs/vendor/releases-semantic-versioning.md
```diff
@@ -0,0 +1,31 @@
...
+
+The admin console sorts releases with semantic versions starting with the earliest version and proceeding to the latest.
```
ajp-io, line 15: This was only added for the case where some versions don't use semantic versioning. I don't think this makes sense when stated more generally on its own like this. This is just in relation to how we sort the versions if some aren't semantic versions.

ajp-io (changes requested):

docs/reference/replicated-cli-channel-disable-semver.md
```diff
@@ -0,0 +1,20 @@
+# channel disable-semantic-versioning
+
+Disable semantic versioning for a channel.
```
jonquil2002, line 3: Disable versus disables?

jonquil2002 (dismissed):
One minor thought/question. But LGTM

docs/vendor/releases-semantic-versioning.md
paigecalvert: @ajp-io re-reading this blurb and the content on the corresponding kots.io topic, it looks like this blurb does state both cases: 1) what does the admin console do when semver is enabled, and 2) what does it do when there are some versions that don't have valid semver on the channel. ("The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.")

Could you clarify the specific suggested change or missing piece? Or, was this concern maybe already helped by your above commit to add the word "enabled" after "For channels with semantic versioning"? Thank you!

docs/vendor/releases-semantic-versioning.md
paigecalvert: Ah great catch, thanks for the clarification.

docs/vendor/releases-semantic-versioning.md
paigecalvert: One other thought is that I could change "For channels with semantic versioning enabled, the Replicated admin console sequences releases by their semantic versions instead of their creation dates." to "For channels with semantic versioning enabled, the Replicated admin console sequences releases **with valid semantic versions** by their ~~semantic versions~~ instead of their creation dates."

Does that perhaps address the concern?

jonquil2002 (approved):
lgtm

docs/vendor/releases-semantic-versioning.md
ajp-io: I was referring to this sentence.
>The admin console sorts releases with semantic versions starting with the earliest version and proceeding to the latest.

The earliest to latest part is just in relation to how releases are sorted if non-valid semvers are present as well. In that case (like the example I provided with abc and xyz labels), it sorts from the earliest to the latest. So this really goes with the example that is just below it.

ajp-io: These comments aren't always clear which line they refer to since they show preceding lines for context!

paigecalvert: Ahh, I see! So, you were hoping for the example to be part of the same paragraph as "The admin console sorts releases with semantic versions starting with the earliest ..." rather than there being a carriage return between them. I can definitely do that. 

I generally break up longer paragraphs to improve readability (our brains tend to struggle more when processing chunks of text). But I can certainly see how it wouldn't be quite as clear that the example corresponds specifically to the case described that sentence with a carriage return breaking them up.

docs/vendor/releases-semantic-versioning.md
```diff
@@ -0,0 +1,31 @@
...
+
+## About Semantic Versioning
+
+When you enable semantic versioning on a channel, the version label for a release promoted to that channel is verified to ensure that it is a valid semantic version. For more information about valid semantic versions, see [Semantic Versioning 2.0.0](https://semver.org).
+
+For channels with semantic versioning enabled, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
+
+The admin console sorts releases with semantic versions starting with the earliest version and proceeding to the latest.
+
+For example, assume that you promote these releases in the following order to a channel: 1.0.0, abc, 0.1.0, xyz, and 2.0.0. Then, you enable semantic versioning on that channel. The admin console would sequence the version history as follows for the channel: 0.1.0, 1.0.0, abc, xyz, 2.0.0.
```
ajp-io, line 17:
```suggestion
If releases that do not use a valid semantic version are already promoted to a channel, the admin console sorts the releases that do have semantic versions starting with the earliest version and proceeding to the latest.

For example, assume that you promote these releases in the following order to a channel: 1.0.0, abc, 0.1.0, xyz, and 2.0.0. Then, you enable semantic versioning on that channel. The admin console would sequence the version history as follows for the channel: 0.1.0, 1.0.0, abc, xyz, 2.0.0.
```

docs/vendor/releases-semantic-versioning.md
ajp-io: Yeah, removing the carriage return is one option. Or we can clarify the case for which this sentence applies. I added another suggestion below that does that, and you can choose which style is preferable.

docs/vendor/releases-semantic-versioning.md
paigecalvert: This commit got outdated, but added it in as part of https://github.com/replicatedhq/replicated-docs/pull/200/commits/72159fdca811e9d148cb7ba840d1a045373f517d

docs/vendor/releases-semantic-versioning.md
ajp-io: I'm going to commit this suggestion since it was marked as resolved but wasn't actually committed. I think maybe it was committed before but got overwritten in subsequent changes.

ajp-io (approved):

PR #201, opened 2022-03-08
44519: Update rn-app-manager.md

jonquil2002:
SC: https://app.shortcut.com/replicated/story/44519/fix-statement-in-app-manager-release-notes

Preview: https://deploy-preview-201--replicated-docs.netlify.app/release-notes/rn-app-manager

### DIFF
--- a/docs/release-notes/rn-app-manager.md
+++ b/docs/release-notes/rn-app-manager.md
@@ -20,7 +20,7 @@ Support for Kubernetes: 1.21, 1.22, and 1.23
 
 ### Bug Fixes
 * Corrects the placeholder Prometheus URL in the admin console dashboard so that it is accurate for embedded installations.
-* Fixes a bug where the warning message indicated a mismatch between the kots CLI version and the version of the admin console that is running in the cluster. This bug printed incorrect messages in some cases.
+* Fixes a bug where the warning message sometimes printed incorrectly when a mismatch was detected between the kots CLI version and the version of the admin console that is running in the cluster.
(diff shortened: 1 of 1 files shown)


docs/release-notes/rn-app-manager.md
```diff
@@ -20,7 +20,7 @@ Support for Kubernetes: 1.21, 1.22, and 1.23
 
 ### Bug Fixes
 * Corrects the placeholder Prometheus URL in the admin console dashboard so that it is accurate for embedded installations.
-* Fixes a bug where the warning message indicated a mismatch between the kots CLI version and the version of the admin console that is running in the cluster. This bug printed incorrect messages in some cases.
+* Fixes a bug where the warning message sometimes printed incorrectly when a mismatch was detected between the kots CLI version and the version of the admin console that is running in the cluster.
```
ajp-io, line 23: Do we want backticks on the kots CLI since it's code, or do we not style it that way elsewhere? And should the "is" be "was" since we use "was" earlier in the sentence?
```suggestion
* Fixes a bug where the warning message sometimes printed incorrectly when a mismatch was detected between the `kots` CLI version and the version of the admin console that was running in the cluster.
```

jonquil2002: Hey, Alex. Good call on the verb tense, they match now. The kots CLI code would use backticks if we were talking about the actual command, e.g. `kots version`. But the way it's written now, we're talking about the name of the CLI, which Marketing has determined is kots CLI (and replicated CLI). So I'll leave it as is, unless we need to document the actual command.

PR #204, opened 2022-03-08
Fix typos and use backticks for code

ajp-io:


jonquil2002 (approved):
lgtm - I wonder if we can globally misspell Kubernetes one more time ;-)

Thank you!!!

PR #211, opened 2022-03-09
44621 adds support lifecycle policy page

paigecalvert:
SC: https://app.shortcut.com/replicated/story/44621/new-support-policy-page

Preview: https://deploy-preview-211--replicated-docs.netlify.app/vendor/policies-support-lifecycle

### DIFF
--- a/docs/vendor/policies-support-lifecycle.md
+++ b/docs/vendor/policies-support-lifecycle.md
@@ -1,0 +1,23 @@
+# Support Lifecycle Policy
+
+Replicated will provide support for products per our terms and services until that product is noted as End of Life (EOL).
+
+| Product Phase                  | Definition           |
+|-----------------------|------------------------|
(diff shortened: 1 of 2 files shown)


paigecalvert:
@AmberAlston Just a note that the spacing in the table columns is not fabulous; we have not yet customized our table styles or come up with a method to handle column spacing/the ability to put carriage returns inside the table cells for markdown tables. Might be another request to consider for the future

AmberAlston (approved):

AmberAlston:
> @AmberAlston Just a note that the spacing in the table columns is not fabulous; we have not yet customized our table styles or come up with a method to handle column spacing/the ability to put carriage returns inside the table cells for markdown tables. Might be another request to consider for the future

@paigecalvert I think it's good enough to get the content up there, but we should have a nearer-term story to revisit the styling, as I agree it's not ideal.

AmberAlston (approved):

PR #216, opened 2022-03-10
[1.67.0] 44677 new info about automatic deploys and udpates

paigecalvert:
SC: https://app.shortcut.com/replicated/story/44677/automatic-updates-docs

Eng SC: https://app.shortcut.com/replicated/story/44111/automatic-upgrades-should-sequence-based-on-semvers-if-the-channel-is-semver-enabled-otherwise-they-should-sort-based-on

Preview: https://deploy-preview-216--replicated-docs.netlify.app/enterprise/updating-apps (specifically the https://deploy-preview-216--replicated-docs.netlify.app/enterprise/updating-apps#configure-automatic-updates section)


docs/enterprise/updating-apps.md
```diff
@@ -1,47 +1,96 @@
...
+   [![New Changes](/images/new-changes.png)](/images/new-changes.png)
+
+1. Click **Preflight results** to run the preflight checks defined by
+the application vendor.
+
+   [![Preflight Checks](/images/preflight-checks.png)](/images/preflight-checks.png)
+:::note
+Preflight failures and warnings do not prevent the upgrade to a new version. You can decide to ignore these failures and proceed with the upgrade.
+:::
+1. To update the application, click **Continue** on the preflight checks page. Or,
```
ajp-io, line 49: Is there a continue button there? The screenshot above shows a re-run button, which is what I recall.

docs/enterprise/updating-apps.md
```diff
@@ -1,47 +1,96 @@
...
+You cannot configure automatic updates for application installed in air gapped
+environments.
 
-[![Preflight Checks](/images/preflight-checks.png)](/images/preflight-checks.png)
+To configure automatic updates:
 
-Preflight failures and warnings do not preclude the upgrade to a new version.
-The installer may elect to ignore these failures and proceed with the upgrade.
+1. In the admin console, go to the Version History tab and click **Configure automatic updates**.
+1. In the Configure automatic updates dialog, under Automatically check for updates,
```
ajp-io, line 77: Do we not bold or put in quotations marks the headings? Just curious.

docs/enterprise/updating-apps.md
```diff
@@ -1,47 +1,96 @@
...
+select a cadence for the admin console to check for updates. To define a custom cadence,
+select **Custom** in the drop-down, then enter a cron expression into the text field.
+1. Under Automatically deploy new versions:
+   * **For applications that use semantic versioning**: Select an option in the dropdown
+   to specify the versions that the admin console automatically deploys. For example,
+   to automatically deploy only new major versions, select **Automatically deploy new major versions**.
```
ajp-io, line 83: I think it would be good to list the options. Also, you can't only deploy majors, so I changed it to patches.
```suggestion
   * **For applications that use semantic versioning**: Select an option in the dropdown
   to specify the versions that the admin console automatically deploys. You can choose to
   automatically deploy patch versions; patch and minor versions; or patch, minor, and
   major versions. For example, to automatically deploy only new patch versions, select 
   **Automatically deploy new patch versions**.
```

ajp-io (changes requested):

docs/enterprise/updating-apps.md
paigecalvert: Great catch; I did not edit any of the existing technical content in this section just for the scope of this story, but I just pulled up an app and it looks like there are a few things wrong here. There is no "Preflight results" button--there a little logo that says "View preflight checks" when you hover over it.

And you are absolutely correct that there is no "Continue" button, let alone one that would deploy the app. I wonder if this is a nod to the preflight checks that run when you first install an application.

I'll edit to remove "Preflight results" and also to remove the line about the Continue button all together

docs/enterprise/updating-apps.md
paigecalvert: This actually came up recently when Jonquil and I were adding to the style guide :) Here's what we decided: https://docs.google.com/document/d/1_TcMD7c2r89rrMxJ5FOmSLEZ7TZda-AIs_MczTsRcs0/edit?disco=AAAAVGwDXjk

docs/enterprise/updating-apps.md
paigecalvert: Nice--thank you.

For those options: I recommend that we don't list the options in the tech doc for a couple reasons: 1) it becomes an artifact that is just begging to get out of date and 2) since the user will be reading the options in the UI as they go through the procedure, it won't provide additional value to have it listed in the doc too.

I was thinking that providing that one example would be sufficient for conveying the logic of the options. But definitely let me know what you think and if you are ok with this approach

(If you are ok with it, I can go in and make the major --> patch edit that you caught)

ajp-io: Given that semantic versioning only has patch, minor, and major version components, it's not something that will change or get out of date IMO. That being said, the example is illustrative. I might choose to use patches and minor versions as the example, since that shows a combination rather than just a single type of version. That seems to convey it better to me. I edited my suggestion above if you want to review/commit it.

docs/enterprise/updating-apps.md
ajp-io: Yeah, the continue button is definitely present during the initial install, so it could have gotten confused there. I'll keep an eye out for your updates.

paigecalvert: Done! https://github.com/replicatedhq/replicated-docs/pull/216/commits/1c849d41833761ec2397defa0878851a2662607c

docs/enterprise/updating-apps.md
```diff
@@ -1,47 +1,96 @@
...
+
+   [![Diff Releases](/images/diff-releases.png)](/images/diff-releases.png)
+   [![New Changes](/images/new-changes.png)](/images/new-changes.png)
+
+1. Click the **View preflight checks** logo to view or re-run the preflight checks defined by
+the application vendor.
+
+   [![Preflight Checks](/images/preflight-checks.png)](/images/preflight-checks.png)
+:::note
+Preflight failures and warnings do not prevent the upgrade to a new version. You can decide to ignore these failures and proceed with the upgrade.
```
ajp-io, line 47: The blue team is about to merge docs on strict preflights, so we should amend this statement to reflect that work too.

ajp-io, line 47: The blue team is about to merge docs on strict preflights, so we should amend this statement to reflect that work too.

jonquil2002: The blue team has updated this topic to reflect that in their PR. My advice is to merge this, and then when the other PR is merged on Monday, they'll have to accept/pull your changes in. They're releasing theirs on Monday.

jonquil2002: The blue team has updated this topic to reflect that in their PR. My advice is to merge this, and then when the other PR is merged on Monday, they'll have to accept/pull your changes in. They're releasing theirs on Monday.

jonquil2002 (approved):
lgtm as is - without the strict preflight info bc it's included in the blue team's PR for Monday.

ajp-io:
This is not yet released, so we shouldn't merge these docs yet.

ajp-io (approved):

PR #228, opened 2022-03-15
Tutorial update disk space requirement

glaubinix:
Based on the host preflight check this disk space requirement should be 50GB instead of 40GB. I am not sure whether this is also an issue for installing it into an existing cluster

`[PASS] Longhorn Disk Usage: The disk containing directory /var/lib/longhorn has at least 50Gi of total space and is less than 60% full`

### DIFF
--- a/docs/vendor/tutorial-installing-without-existing-cluster.md
+++ b/docs/vendor/tutorial-installing-without-existing-cluster.md
@@ -119,7 +119,7 @@ To create the test server and install the app manager:
     * Ubuntu 18.04
     * At least 8 GB of RAM
     * 4 CPU cores
-    * At least 40GB of disk space
+    * At least 50GB of disk space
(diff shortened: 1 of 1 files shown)


paigecalvert (approved):
Thank you, @glaubinix!

While that particular failing preflight check does not prevent app deployment for the purposes of this tutorial, this seems like a reasonable edit to make 👍 

For the existing cluster tutorial: the same problem should not apply because you are not required to have a VM:
"Installing the app manager on existing clusters is similar to using a [Kubernetes installer cluster](https://docs.replicated.com/vendor/tutorial-installing-without-existing-cluster#installing-and-testing), except instead of bringing a plain virtual machine (VM), you will use a pre-built Kubernetes cluster and deploy your application into a namespace."

Thank again!

PR #233, opened 2022-03-17
Remove type from flag

PeterGerrard:


paigecalvert (approved):
Great catch, thanks @PeterGerrard

PR #237, opened 2022-03-18
Use "Kubernetes installer-created" terminology

ajp-io:
Rather then "embedded cluster" or "Kubernetes installer cluster"


jonquil2002 (approved):
lgtm. Thanks for catching these!

PR #272, opened 2022-04-04
.airgap is not a thing for kURL

laverya:
kurl airgap bundles have a .tar.gz extension, not .airgap


paigecalvert (approved):
Great catch, thank you @laverya !! Committed just a couple editorial changes. I'm going to merge this in after the checks complete.

PR #285, opened 2022-04-08
[1.69.0] release notes

sgalsaleh:


docs/release-notes/rn-app-manager.md
```diff
@@ -4,6 +4,28 @@ toc_max_heading_level: 2
...
+
+### Improvements
+* Adds an "Add a CA Certificate" field to the snapshot page.
+* Changes the KOTS login session timeout to 12 hours and validates expired sessions and logs out automatically.
+* Extends the login session timeout by an hour if the session's expiry date is at least less than one hour of session timeout.
+* Removes expired sessions from the store in a daily cleanup job
+* Two kotsadm airgap bundles are available on the KOTS release page, one of which does not include MinIO.
+
+### Bug Fixes
+* Fixes an issue that caused instances to run into DockerHub's rate limiting after running the [ensure-secret](/reference/kots-cli-docker-ensure-secret) command.
```
sgalsaleh, line 24: IMO this can be turned into an improvement and it also needs docs changes here https://docs.replicated.com/reference/kots-cli-docker-ensure-secret#usage

docs/release-notes/rn-app-manager.md
```diff
@@ -4,6 +4,27 @@ toc_max_heading_level: 2
...
+Support for Kubernetes: 1.21, 1.22, and 1.23
+
+### New Features
+* Adds the ability to switch from a community license to a different license from the same application.
+
+### Improvements
+* Adds an "Add a CA Certificate" field to the snapshot page.
+* User sessions will expire after 12 hours of inactivity.
+* Removes expired sessions from the store in a daily cleanup job
+* Two kotsadm airgap bundles are available on the KOTS release page, one of which does not include MinIO.
```
divolgin, line 20: I don't think this is kots release notes

docs/release-notes/rn-app-manager.md
divolgin: good point

docs/release-notes/rn-app-manager.md
ajp-io: That's an interesting point, and I think I agree. I'd like docs's opinion too.

@replicatedhq/replicated-docs We started producing two assets for KOTS: one that includes MinIO and one that doesn't. That lays the foundation for a feature that Blue team is building. This note isn't really relevant to the product at all at this point. We are just changing KOTS's release process a bit to produce two assets instead of one, and later our product will leverage those assets for different things. So Dmitriy's point, which I agree with, is that this isn't really an app manager release note. Does it also make sense to you to remove this?

docs/release-notes/rn-app-manager.md
ajp-io: Can you make an update to those docs, @divolgin?

docs/release-notes/rn-app-manager.md
paigecalvert: @ajp-io ah great question. If I'm understanding correctly, it sounds like there is no action a user could/would take based on this release note at this stage. If that's the case, then yes, makes sense to remove 👍 

(Related: To reduce noise, I even tend to leave out bug fixes when it's not something a user did notice or likely would have noticed in favor of only the noticeable bug fixes. Definitely can be selective in cases about what goes in the release notes)

sgalsaleh: it might interest the vendors to test installations with that bundle

docs/release-notes/rn-app-manager.md
ajp-io: We'll link these in later

ajp-io (dismissed):

ajp-io (approved):

paigecalvert (approved):

PR #286, opened 2022-04-08
46247: Update procedure.md

jonquil2002:
SC: https://app.shortcut.com/replicated/story/46247/updated-procedure-template-with-html-table-style


paigecalvert (approved):
Thank you!

PR #289, opened 2022-04-08
[1.69.0] Update ensure-secret command to better reflect when it actually does

divolgin:


ajp-io (approved):

paigecalvert (approved):
Thank you! Great updates. Just minor updates to align with the style guide:
- Removed the inline line per style guide (there is also already a link to this CLI reference topic at the bottom of this topic)
- Couple edits for active voice/present tense
- Swap utilize for use

PR #298, opened 2022-04-18
Update custom-resource-config.md

JustinN-Replicated:


JustinN-Replicated:
Screenshot of change below from preview; we may want to consider a different example since the existing example is specifically around cert files maybe? Though I do think it is odd that Stackstate's customer was providing a CA cert(s) as binary so maybe not.

image

paigecalvert:
Hi @JustinN-Replicated, just checking in on this since it's been open for awhile. Looks like it is waiting for a review from @jala-dx before the docs team reviews? Let me know!

JustinN-Replicated:
Going to close the PR. This was addressed with [this](https://docs.replicated.com/reference/template-functions-static-context#yamlescape)

PR #299, opened 2022-04-18
added namespace to command

grantmiller:


jonquil2002 (approved):
@grantmiller  I committed a description to the table for `-n` and I'm approving it so that you can merge it if you agree with the change - or lmk if further changes are needed. Also lmk if you prefer that I merge it. Thanks.

PR #304, opened 2022-04-19
46378-New: Edits to headings under Support Bundle CLI

jonquil2002:
SC: https://app.shortcut.com/replicated/story/46378/collect-a-support-bundle-using-cli

Updates to headings to make it cleaner.


paigecalvert (dismissed):
Looks great! Just fixed a typo. Also noticed that "using" was lowercased in the h2s and not sure if that was intentional. 

Q: would it make sense for the "install the plugin" procedure to be a prereq? Looking at it again it feels different from the other h3s it's grouped with

paigecalvert (approved):
This looks great

PR #308, opened 2022-04-20
46045 Clarify the difference between kotsadm bundle & airgap bundle

paigecalvert:
SC: https://app.shortcut.com/replicated/story/46045/incorrect-info-in-air-gap-install-instructions-for-existing-cluster

https://deploy-preview-308--replicated-docs.netlify.app/enterprise/installing-existing-cluster-airgapped#push-images-and-install-the-admin-console

### DIFF
--- a/docs/enterprise/installing-existing-cluster-airgapped.md
+++ b/docs/enterprise/installing-existing-cluster-airgapped.md
@@ -9,50 +9,66 @@ For information about how to install on a cluster created by the Kubernetes inst
 
 ## Push Images and Install the Admin Console
 
-You can install the admin console using the kots CLI plugin for the kubectl command-line tool.
+You can install the admin console using the kots CLI plugin for the kubectl command-line tool. To install the admin console, you use the admin console binary bundle, `kotsadm.tar.gz`.
 
-The `.airgap` package is named `kotsadm.tar.gz`. You can download `kotsadm.tar.gz` from the kots release page on GitHub. See [Releases](https://github.com/replicatedhq/kots/releases) in the kots GitHub repository.
+To push images and install the admin console:
 
-The asset version must match the kots CLI version, which can be determined by running:
+1. Download `kotsadm.tar.gz` from the kots release page on GitHub. See [Releases](https://github.com/replicatedhq/kots/releases) in the kots GitHub repository.
 
-```shell
-kubectl kots version
-```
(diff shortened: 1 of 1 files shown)


docs/enterprise/installing-existing-cluster-airgapped.md
```diff
@@ -9,50 +9,64 @@ For information about how to install on a cluster created by the Kubernetes inst
...
-The next step is to install the admin console using images pushed in the previous step.
-Registry credentials provided in this step only need to have read access, and they will be stored in a Kubernetes secret in the same namespace where admin console will be installed.
+   ```shell
+   kubectl kots admin-console push-images ./kotsadm.tar.gz private.registry.host/app-name \
+     --registry-username RW-USERNAME \
+     --registry-password RW-PASSWORD
+   ```
 
-These credentials will be used to pull the images, and will be automatically created as an imagePullSecret on all of the Admin Console pods.
+   Where:
```
jonquil2002, line 32: Some of the companies I worked for got rid of this "where" format and replaced it with a table with the values/definitions. I'm not sure if we want to consider that or if this format is still considered the prevalent form in the industry. This may be a good candidate to capture for the Style Guide.

jonquil2002 (approved):
lgtm

jonquil2002:
@paigecalvert  This looks so much better. I'm thinking that we should create another card to turn the remainder of the topic into task steps. I know that's outside the scope of this story.

docs/enterprise/installing-existing-cluster-airgapped.md
paigecalvert: I am totally opening to discussing a different format for this! Primarily using `Where` because it is familiar to me. Want to add it to sync notes for next time we get together? We can look at the table example you mention here

paigecalvert:
> @paigecalvert This looks so much better. I'm thinking that we should create another card to turn the remainder of the topic into task steps. I know that's outside the scope of this story.

Indeed indeed. I believe the work to do more of a style edit on these topics is already captured by the two stories we triaged on Monday:
- https://app.shortcut.com/replicated/story/43729/update-the-enterprise-install-in-an-air-gapped-environment-topic-guide
- https://app.shortcut.com/replicated/story/43736/update-the-enterprise-install-with-the-kubernetes-installer-topic-guide

PR #310, opened 2022-04-20
44514 replace unicode quotes with ascii

paigecalvert:
SC: https://app.shortcut.com/replicated/story/44514/documentation-examples-use-unicode-quotes

Preview: https://deploy-preview-310--replicated-docs.netlify.app/reference/template-functions-kurl-context

### DIFF
--- a/docs/reference/template-functions-kurl-context.md
+++ b/docs/reference/template-functions-kurl-context.md
@@ -10,7 +10,7 @@ The Kubernetes installer is based on the open source kURL project, which is main
 
 KurlBool, KurlInt, KurlString, and KurlOption all take a string yamlPath as a param.
 This path is the path from the manifest file, and is delineated between addon and subfield by a period ’.’.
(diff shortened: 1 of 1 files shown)


docs/reference/template-functions-kurl-context.md
```diff
@@ -28,7 +28,7 @@ func KurlBool(yamlPath string) bool
 Returns the value at the yamlPath if there is a valid boolean there, or false if there is not.
 
 ```yaml
-‘{{repl KurlBool “Docker.NoCEonEE” }}
+'{{repl KurlBool "Docker.NoCEonEE" }}'
```
paigecalvert, line 31: Previously, the KurlBool example only had a single quote at the start (it lacked a second single quote at the end). I don't think this was intentional, so I added a closing single quote.

cbodonnell (approved):

docs/reference/template-functions-kurl-context.md
cbodonnell: good catch

PR #311, opened 2022-04-20
Removed enterprise bucket references & added info

AmberAlston:
We updated the diagram but kept the references to enterprise buckets. That could infer enterprise vendors are treated differently. This isn't true, only 'add-on' vendors are treated differently.


paigecalvert:
@AmberAlston thank you! I noticed one more instance of "enterprise", so removed that as part of a direct commit.

paigecalvert (approved):

PR #316, opened 2022-04-21
 HOLD ON RELEASE - 46064: Docs for pin-latest-deployable flag for the get-versions cli command

sgalsaleh:


ajp-io (dismissed):

docs/reference/kots-cli-get-versions.md
```diff
@@ -13,14 +13,15 @@ kubectl kots get versions [app-slug] [flags]
...
+| Flag                      | Type   | Description                                                                                         |
+| :------------------------ | ------ | --------------------------------------------------------------------------------------------------- |
+| `-h, --help`              |        | Help for `get versions`.                                                                            |
+| `-n, --namespace`         | string | (Required) The namespace where the admin console is running.                                        |
+| `--current-page`          | int    | Offset, by page size, at which to start retrieving versions. **Default:** 0                         |
+| `--page-size`             | int    | Number of versions to return. **Default:** 20                                                       |
+| `--pin-latest`            | int    | When set to true, always returns the latest version at the beginning. **Default:** false            |
+| `--pin-latest-deployable` | int    | When set to true, always returns the latest deployable version at the beginning. **Default:** false |
```
ajp-io, line 23: @replicatedhq/replicated-docs This new flag is relevant when there are required releases, in which case the latest version isn't necessarily deployable. So the latest deployable version would return the required release that you have to deploy first. Maybe we should mention required releases here to make it clear what this is for, but let me know your thoughts. I don't know if it'll get too long.

paigecalvert: Oh that's a great question @ajp-io. I think it could definitely be helpful for the purpose of clarifying how latest and latest deployable might differ. What about something like this?

When set to true, always returns the latest version that is available [for upgrade/to be deployed]. The latest deployable version can differ from the latest version if the vendor prevents skipping one or more versions during upgrade.

paigecalvert: Or if we want to avoid "vendor", could also go with the passive "...if you are prevented from skipping..."

ajp-io:
```suggestion
| `--pin-latest-deployable` | int    | When set to true, always returns the latest version that can be deployed. The latest deployable version can differ from the latest version if a required version, which cannot be skipped, is present. **Default:** false |
```

ajp-io (approved):

PR #317, opened 2022-04-21
--use-restic flag when installing velero

cbodonnell:
The flag to use with the `velero install` command is `--use-restic` per [Velero docs](https://velero.io/docs/main/restic/#install-restic).

Otherwise, will get the error: `An error occurred: unknown flag: --restic`


ajp-io (approved):

paigecalvert (approved):
Nice, thanks @cbodonnell !

PR #319, opened 2022-04-22
43918: Uploading a Support Bundle to Replicated Support

jonquil2002:
SC: https://app.shortcut.com/replicated/story/43918/document-how-to-upload-a-support-bundle-to-vendor-portal-and-share-it-with-replicated

Preview: https://deploy-preview-319--replicated-docs.netlify.app/vendor/support-submit-request


docs/vendor/support-submit-request.md
```diff
@@ -0,0 +1,31 @@
...
+
+You can submit a support request with or without a support bundle using the Replicated vendor portal. Uploading a support bundle is secure and helps the Replicated support team troubleshoot your application faster.
+
+If you or your customer are unable to generate a support bundle, you can submit the request without a support bundle and the support team will assist you.
+
+For more information about how to generate a support bundle using either the Replicated admin console or the kubectl CLI, see [Troubleshooting an Application](https://docs.replicated.com/enterprise/troubleshooting-an-app).
+
+To submit a support request:
+
+1. From the [vendor portal](https://vendor.replicated.com), click **Support** in the upper right corner of the page.
```
paigecalvert, line 11: We can also link them directly to the support page with https://vendor.replicated.com/support.

I do see value in telling them how to navigate to the Support page in the way that you did, though.

Just throwing it out there if you thought it would be helpful to add a sentence like "Or, go to https://vendor.replicated.com/support."

Up to you

docs/vendor/support-submit-request.md
```diff
@@ -0,0 +1,31 @@
...
+
+  The Vendor Support page opens.
```
paigecalvert, line 13: I think we can do without "The Vendor Support page opens" and "The Support Request dialog opens", since they'll be seeing these things open in the UI as they follow along. And, this UI workflow is fairly straightforward as is.

I do see that there can be value in mentioning the name of the pages... Maybe one workaround would be something like "Click **Open a Support Request** to open the Support Request dialog." ?

paigecalvert (approved):
Looks great, thank you!

docs/vendor/support-submit-request.md
jonquil2002: Typically I would only document one method (and I forgot about that direct link, so thanks, I added that). But I kept in the navigation too, which I think is important because they may already be in the UI. Also, they would be confused if they were trying to open the request from the Troubleshoot page and they did not have a bundle to upload. From Troubleshoot, they have to upload the bundle first. So I designed the docs to compensate for that.

paigecalvert: I'm with you 👍

paigecalvert (approved):
lgtm!

PR #324, opened 2022-04-26
give an example of the sealedsecret certificate

laverya:


paigecalvert:
Hi @laverya, just checking in on this one since it's been open for awhile. 

Looks like this is still waiting on a review from @marccampbell ?

paigecalvert (approved):
@laverya approved--feel free to merge assuming there is no concern having this public key on the docs site 👍

PR #337, opened 2022-05-02
Update packaging-private-registry-security.md

amit-schnitzer:
fix wording (access seems more accurate describing the situation here)


paigecalvert (approved):
Thanks @amit-schnitzer :) 
And sorry for the delay--feel free to ping us in @product-docs or with @replicatedhq/replicated-docs if you don't see us merging your PR quickly

PR #339, opened 2022-05-02
39410: Update gitops workflows.md

jonquil2002:
SC: https://app.shortcut.com/replicated/story/39410/gitops-clarification-needed-for-enabling-gitops-application

Preview:

https://deploy-preview-339--replicated-docs.netlify.app/enterprise/gitops-workflow


jonquil2002:
@ajp-io  Will you review this topic and help make sure that it is technically accurate?

The SC story was originally just to clarify the Path information. In the process of it, I also did these tasks because it made sense to do so:

- Condensed the single-app and multi-app topics into one topic

- Updated the file name so that it will not be confusing to vendors if they look at the URL (I also requested redirects, so we can't merge this topic until Chase has that ready to go. We'll merge/redirect simultaneously).

- Rewrote and reformatted most of the content to make it compliant with our writing standards. 

In the process of rewriting the content, some additional points of confusion came up. Can you answer these questions for me?

- Step 4b: When I tested the UI, it seems that only the default is available and the user cannot choose something else for action to take for upgrades. Should I change the existing documentation to just say this is the only option available?

- Step 4c: For **What content will it contain?** it seems like there is no list to choose from, there is only the default setting, regardless of which GitOps provider I choose. Should I change the existing documentation to just say this is the only option available?

- For the First Commits section, it also talks about subsequent commits and updates "that have not been deployed".  That is leftover wording from before there were automatic updates available. What happens now when a user has GitOps enabled and also has automatic updates enabled? 

Thanks!

ajp-io:
@jonquil2002 GitOps isn't the most well understood part of our product :) I want to go through the workflow as part of answering the questions to be sure that I answer correctly. So I need a day or two to have a chance to do that to ensure that everything is accurate.

jonquil2002:
@ajp-io  No rush. I did a lot of testing myself and I know what you mean.  Thanks for helping with this :)

jonquil2002:
@dexhorthy  You had mentioned that some of the GitOps content was erroneous, especially the First Commit section. I revised the GitOps section and condensed single and multi into one procedure, and clarified the Path information. Even though Alex is giving a SME review, it would be helpful for you at look at it as well. Please :)

docs/enterprise/gitops-workflow.md
```diff
@@ -0,0 +1,78 @@
...
+
+    1. For **What content will it contain?**, select what type of asset to deliver to the Git repository. Using the **Rendered YAML** option results in a single, rendered YAML file being committed to the repository.
+
+      ![GitOps Action](/images/gitops-action.png)
+
+    1. Click **Finish setup** if this is a first time setup. For subsequent edits, click **Update settings**.
+
+      The Gitops subtab opens and displays a public deployment key. The private key is stored securely in the admin console.
+
+1. Click **Copy key** to copy the key. Click **repo settings page** and add the public deployment key in the repository settings page. Enable the write access permissions for this key, otherwise the commit fails when KOTS pushes the commit to the repository.
```
ajp-io, line 65: Lgtm, but this wording felt better to me.
```suggestion
1. Click **Copy key** to copy the key. Click **repo settings page** and add the public deployment key in the repository settings page. Enable the write access permissions for this key, otherwise KOTS will be unable to push the commit to the repository.
```

jonquil2002: Yep, I hear ya. Adapting your change, just using present tense instead, e.g. "cannot" instead of "will be unable to".

jonquil2002:
@ajp-io  I think I've resolved everything:

4b/c: No matter which GitOps provider I choose, there are no options other than the default. I asked Grayson where I should open a story/bug to get the UI wording to remove the reference to "it's full output" since only a single rendered YAML file is possible. No SME responded otherwise.

4c/d: I created a new install on a new VM, and this is the latest version of app manager that I can muster. The Cancel button still shows in mine, but maybe that also something that shows only on an update existing GitOps setting thinger versus a first time GitOps setup. So, I'm not going to worry about replacing the screenshot in this PR, due to time constrictions.

5: Agreed with your changes and add one more minor one of my own, as stated above: "cannot push the commit".

First Commits section: Ada said there use to be a PR option there but she didn't know when it was removed. I searched the newer and old Release Notes and there was a mention of removing UI wording for features that aren't supported yet. I think the PR may have been one of those features, though they aren't specifically named. (This is also why I think the rendered YAML/full output wording exists.) Anyway, I removed references to pull requests in that section.

Please see if you can approve or feel free to suggest other changes. Thanks so much!!!!!

jonquil2002:
@ajp-io @paigecalvert  I committed Alex's last suggestion. Looks like maybe you two can approve this, please PTAL.

ajp-io (approved):
LGTM

paigecalvert (approved):
LGTM!

PR #342, opened 2022-05-03
Add missing 1.70.0 release notes

sgalsaleh:


docs/release-notes/rn-app-manager.md
```diff
@@ -27,6 +27,8 @@ Support for Kubernetes: 1.21, 1.22, and 1.23
 * Fixes an issue where ingress status informers always reported as "Missing" in Kubernetes 1.22+.
 * Fixes an issue that caused image garbage collection in Kubernetes installer-created clusters (embedded clusters) to remove images outside of the application's dedicated registry namespace.
 * Fixes an issue where a newer version might not have a **Deploy** button after the configuration is updated from the currently deployed version.
+* Fixes an issue where the legend on the dashboard graphs were blank.
```
JustinN-Replicated, line 30: I am presuming that this was only for the default graphs included from prometheus on embedded installs - is that correct? If so maybe we should be more explicit (reference: https://docs.replicated.com/vendor/admin-console-prometheus-monitoring)

jonquil2002: Also, I think it's one legend, yes? If so, it should be "was great" instead.

sgalsaleh: @JustinN-Replicated i think this affects any prometheus configuration both on existing and embedded clusters and not just the default graphs.

@jonquil2002 there are multiple legends displayed on the graph and they were all blank.

JustinN-Replicated: Got it, ok thanks!

JustinN-Replicated (approved):
lgtm

jonquil2002 (approved):
lgtm with minor changes made.

PR #345, opened 2022-05-05
44385 Improve documentation on customizing the config screen

paigecalvert:
Story: https://app.shortcut.com/replicated/story/44385/packaging-an-application-should-have-an-intro-to-the-config-object-early-on (There is more to be done here: I added some necessary spin-off stories in the story comments)

Previews:

(new topic) About the configuration screen: https://deploy-preview-345--replicated-docs.netlify.app/vendor/config-screen-about
(re-purposes existing topic) Creating and editing configuration values: https://deploy-preview-345--replicated-docs.netlify.app/vendor/admin-console-customize-config-screen
(new topic) Mapping user-supplied values: https://deploy-preview-345--replicated-docs.netlify.app/vendor/config-screen-map-inputs


docs/vendor/admin-console-customize-config-screen.md
```diff
@@ -1,57 +1,86 @@
...
+
+To include a configuration screen in the Replicated admin console for your application, you add a Config custom resource manifest file to the release in the Replicated vendor portal. You define the fields that appear on the configuration screen as an array of `groups` and `items` in the Config custom resource manifest.
+
+There are several types of `items` supported in the Config manifest that allow you to collect different types of user inputs. For example, you can use the `password` input type to create a text field on the configuration screen that hides user input.
+
+For more information about the syntax of the Config custom resource manifest, see [Config](../reference/custom-resource-config) in the _Custom Resources_ section.
+
+## Add Fields to the Configuration Screen
+
+Applications can include a configuration screen in the Replicated admin console to collect required or optional values from your users that are used to start your application.
```
dexhorthy, line 15:
```suggestion
Applications can include a configuration screen in the Replicated admin console to collect required or optional values from your users that are used to run your application.
```

docs/vendor/config-screen-about.md
```diff
@@ -0,0 +1,31 @@
+# About the Configuration Screen
+
+Applications can include a configuration screen in the Replicated admin console to collect required or optional values from your users that are used to start your application.
```
dexhorthy, line 3:
```suggestion
Applications can include a configuration screen in the Replicated admin console to collect required or optional values from your users that are used to run your application.
```

docs/vendor/config-screen-about.md
```diff
@@ -0,0 +1,31 @@
...
+
+For more information about how to add custom fields to the configuration screen, see [Adding and Editing Configuration Fields](admin-console-customize-config-screen).
+
+If you use a Helm chart for your application in Replicated, your users provide any values specific to their environment from the configuration screen, rather than in the Helm chart `values.yaml` file.
+
+By allowing your users to provide configuration values in the admin console rather than in the Helm `values.yaml` file directly, you can control which options you expose to your users. It also makes it easier for your users to provide their inputs through a user interface, rather than having to edit a YAML file.
```
dexhorthy, line 9: This got too wordy as I wrote the alternative, but I'd lead with the primary benefit which is a UI, and then have the "you get to control what you expose" as kind of an "also..." sort of thing (as it's less of a universal benefit than the improved user experience of using a UI). Anyways the below is in the right direction but feel free to slide/dice/iterate

```suggestion
By allowing your users to provide configuration values in the admin console rather than in a Helm `values.yaml` file directly or `--set` CLI commands, you make it easier for your users to provide their inputs through a user interface, rather than having to edit a YAML file or construct complex CLI invocations. It also allows you to control which options you expose to your users .
```

docs/vendor/config-screen-about.md
```diff
@@ -0,0 +1,31 @@
...
+
+For example, you can use the configuration screen to provide database configuration options for your application. Your users could connect your application to an external database by providing required values in the configuration screen, such as the host, port, and a username and password for the database. For a tutorial of this use case, see [Adding Database Configuration Options for your Application](tutorial-adding-db-config).
```
dexhorthy, line 11:
```suggestion
For example, you can use the configuration screen to provide database configuration options for your application. Your users could connect your application to an external database by providing required values in the configuration screen, such as the host, port, and a username and password for the database. Replicated's config screen also allows you to provide an alternative batteries-included option where the database runs in-cluster as part of the application. For a tutorial of this use case, see [Adding Database Configuration Options for your Application](tutorial-adding-db-config).
```

docs/vendor/config-screen-about.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Viewing the Configuration Screen
+
+Users of your application can access the configuration screen from the admin console:
+* During application installation if you included required fields. See [Application Installation](#application-installation) below.
+* At any time after application installation on the Config tab. See [Admin Console Config Tab](#admin-console-config-tab) below.
+
+### Application Installation
+
+If you add any required fields, the admin console displays the configuration screen when the user installs the application, after they upload their license file.
```
dexhorthy, line 21:
```suggestion
If you add a KOTS `kind: Config` object to your application manifests, the admin console displays the configuration screen when the user installs the application, after they upload their license file.
```

docs/vendor/config-screen-about.md
```diff
@@ -0,0 +1,31 @@
...
+
+![configuration screen that displays during application install](../../static/images/initial-config.png)
+
+### Admin Console Config Tab
+
+For applications that do not have any required fields, the admin console does not display the configuration screen during application installation.
```
dexhorthy, line 27: False, AFAIK the config screen is always shown during setup. I will just make the one comment here and not call it out everywhere I see it from here on out.

```suggestion
```

docs/vendor/admin-console-customize-config-screen.md
```diff
@@ -1,57 +1,86 @@
...
+After you create fields in the Config manifest file, you then map the fields to other manifest files in your application to apply the user-supplied values. If you use a Helm chart for your application, you then map the fields that you add to the `values.yaml` file. For more information, see [Next Steps](#next-steps) below.
+
+To add fields to the admin console configuration screen:
+
+1. In the [vendor portal](https://vendor.replicated.com/apps), click **Releases**. Then, either click **Create release** to create a new release, or click **View YAML** to edit an existing release.
+1. Create or open the Config custom resource manifest file in the desired release. A Config custom resource manifest file has `kind: Config`.
+1. In the Config custom resource manifest file, define custom user-input fields in an array of `groups` and `items`:
+   * `groups`: A set of `items`. Each group must have a `name`, `title`, `description`, and `items`.
+   * `items`: An array of user input fields. Each array under `items` must have a `name`, `title`, and `type`. You can also include several optional properties.
+
```
dexhorthy, line 28: I would love to see a teeny bit more content on why its split into groups vs items -- what should I put in a group? What items can all be in the same group. 

A screenshot showing the group nav items on the left side might help to articulate this.

Also open to pushback that this level of detail is better suited to Reference, but my first impression as a user reading this trying to get started is "well what do I use for which one"

docs/vendor/admin-console-customize-config-screen.md
```diff
@@ -1,57 +1,86 @@
...
+             type: password
+             required: true
+   ```
+
+   The example above includes a single group with the name `smtp_settings`.
+
+   The `items` array for the `smtp_settings` group includes four user-input fields: `enable_smtp`, `enable_smtp`, `smtp_username`, and `smtp_password`.
+
+1. (Optional) Add default values for the fields. You can add default values using one of the following properties:
+   * **With the `default` property**: When you include the `default` key, the Replicated app manager uses this value when rendering the manifest files for your application. The value then displays as a placeholder on the configuration screen in the admin console for your users. The app manager only uses the default value if the user does not provide a different value.
```
dexhorthy, line 67: also, if a later release of your app changes the `default` field for a configuration option, end-user instances will receive the updated value if and only if they have not changed the field from the `default` when they initially installed.

docs/vendor/admin-console-customize-config-screen.md
```diff
@@ -1,57 +1,86 @@
...
+   **Example**:
+
+   ```yaml
+   - name: smtp_password
+     title: SMTP Password
+     type: password
+     required: true
+    ```      
+
+1. Save and promote the release to a development environment to test your changes.
```
dexhorthy, line 80: Love that we're writing these as little mini tutorials, reminding folks when and how to ship & test their changes.

docs/vendor/config-screen-map-inputs.md
```diff
@@ -0,0 +1,144 @@
+# Mapping User-Supplied Values
+
+This topic describes how to map the values that your users provide in the Replicated admin console configuration screen to your application.
+
+This topic assumes that you have already added custom fields to the admin console configuration screen by editing the Config custom resource. For more information, see [Creating and Editing Configuration Fields](admin-console-customize-config-screen).
+
+## Overview of Mapping Values
+
+You use the values that your users provide in the admin console configuration screen to render YAML in the custom resource manifest files for your application.
```
dexhorthy, line 9: small clarification - not all user-supplied yaml manifests are Custom Resources

```suggestion
You use the values that your users provide in the admin console configuration screen to render YAML in the manifest files for your application.
```

docs/vendor/config-screen-map-inputs.md
```diff
@@ -0,0 +1,144 @@
...
+
+For example, if you provide an embedded database with your application, you might add a field on the admin console configuration screen where users input a password for the embedded database. You can then map the password that your user supplies in this field to the Secret manifest file for the database in your application.
+
+Similarly, you can include fields on the configuration screen where your users can enable a custom ingress controller for the cluster. You then map these user-supplied values to the Ingress custom resources in your application.
+
+For a tutorial of mapping database configuration options in a sample application, see [Adding Database Configuration Options for your Application](tutorial-adding-db-config).
+
+For an example of adding custom Ingress resources based on user-supplied configuration, see [Configuring Cluster Ingress](packaging-ingress).
```
dexhorthy, line 17: Nice! This is a great guide that doesn't get enough looks. Glad we're linking to it.

docs/vendor/config-screen-map-inputs.md
```diff
@@ -0,0 +1,144 @@
...
+   ```
+
+   In the example above, the field name to map is `example_config_screen_field_name`.
+
+1. In the same release in the vendor portal, open the custom resource manifest file where you want to map the value for the field that you selected.
+
+   Use the ConfigOption template function to map the user-supplied value in a key value pair:
+
+   ```yaml
+   example_key: '{{repl ConfigOption "CONFIG_SCREEN_FIELD_NAME"}}'
```
dexhorthy, line 69: Why not use the field name from the example? `example_config_screen_field_name`

docs/vendor/config-screen-map-inputs.md
```diff
@@ -0,0 +1,144 @@
...
+   apiVersion: v1
+   kind: Secret
+   metadata:
+     name: tls-secret
+   type: kubernetes.io/tls
+   data:
+     tls.crt: '{{repl ConfigOption "tls_certificate_file" }}'
+     tls.key: '{{repl ConfigOption "tls_private_key_file" }}'
+   ```
+
```
dexhorthy, line 87: This is another great example. I might add a link back to the ingress guide because the last option where we re-use the `kotsadm-tls` secret should be considered before a user goes in marking up their own custom TLS ConfigOption Fields.

```suggestion
 **Note**: While the example above is perfectly valid, if you're specifically looking to handle TLS secrets, check out the last example in [Configuring Cluster Ingress](packaging-ingress). Depending on your use case, there are some shortcuts you can take to re-use the certificates uploaded for the App Manager console itself.
```

docs/vendor/config-screen-map-inputs.md
```diff
@@ -0,0 +1,144 @@
...
+   For more information about the ConfigOption template function, see [Config Context](../reference/template-functions-config-context#configoption) in the _Template Functions_ section.
+
+1. Save and promote the release to a development environment to test your changes.
+
+### Map Values to a Helm Chart
+
+The `values.yaml` file in a Helm chart contains values that are specific to the end-user environment. With Replicated, your users provide these values through the configuration screen in the admin console. You customize the configuration screen based on the required and optional configuration fields that you want to expose to your users.
```
dexhorthy, line 94:
```suggestion
The `values.yaml` file in a Helm chart defines parameters that are specific to the end-user environment. With Replicated, your users provide these values through the configuration screen in the admin console. You customize the configuration screen based on the required and optional configuration fields that you want to expose to your users.
```

dexhorthy: i dunno how you wanna wordsmith these but a few tiny tweaks

dexhorthy (approved):
A few small comments and suggestions, but overall a great and much-needed addition. I really like where it lands in the "packaging your application" section, as well as how the three topics are split up. Two thumbs up. 👍 👍

docs/vendor/config-screen-map-inputs.md
```diff
@@ -0,0 +1,144 @@
...
+
+To map the values that your users provide in the admin console configuration screen to your Helm chart `values.yaml` file, you create a Replicated HelmChart custom resource.
+
+To follow a tutorial that maps values from the configuration screen to a Helm chart, see [Example: Mapping the Configuration Screen to Helm Values](helm-mapping-example).
+
+To map user inputs from the configuration screen to the `values.yaml` file:
+
+1. In the [vendor portal](https://vendor.replicated.com/apps), click **Releases**. Then, click **View YAML** next to the desired release.
```
dexhorthy, line 102: @cremerfc @tdensmore @jdewinne do you folks think it will be obvious to advanced users that they can do this in their GH repo or local CLI env as well?

docs/vendor/admin-console-customize-config-screen.md
paigecalvert: That's a great note-- this is lacking a nice example to help people conceptualize

docs/vendor/config-screen-map-inputs.md
paigecalvert: Very nice

docs/vendor/config-screen-map-inputs.md
paigecalvert: You know, I was thinking about that as I was writing these... there are probably a few good ways to address this in the docs (that is, that vendors can accomplish tasks through a few different mediums, including the vendor portal, the cli, their gh repo). 

I do think we should be documenting these different paths. Curious to hear what you all think. If you agree, this could be another effort to bring these different paths in across the procedures in the docs.

docs/vendor/config-screen-about.md
paigecalvert: Ahh got it, ok. I'll re-work this section to reflect that

paigecalvert:
Woo! Thanks @dexhorthy. All great corrections and suggestions. I'll get your notes incorporated, then ping again when it's ready for a final approval/merge.

docs/vendor/config-screen-map-inputs.md
dexhorthy: I wonder if there's a generic way we can say "edit your manifests" without specifying which of the three levels of advanced-ness?

paigecalvert: Yeah, totally--one idea would be that we can have a single topic that describes how to do these three methods. Then, we can link to that topic from procedures like this, when there are few options for how to access/edit/release the manifests

docs/vendor/config-screen-about.md
paigecalvert: You know, it was probably too wordy to begin with. Re-wrote:

>If you use a Helm chart for your application in Replicated, your users provide any values specific to their environment from the configuration screen, rather than in a Helm chart `values.yaml` file.
>
>This means that your users can provide configuration values through a user interface, rather than having to edit a YAML file or use `--set` CLI commands. The admin console configuration screen also allows you to control which options you expose to your users.

docs/vendor/admin-console-customize-config-screen.md
paigecalvert: Update: moved this description up to the Overview section so as not to clutter the procedure, and added examples for each. I also updated the example yaml to copy what we use for Sentry Enterprise, and included a screenshot of how the SMTP Settings section of the config screen looks in the admin console when installing Sentry Enterprise (so: here's the yaml, and here how the yaml displays)

docs/vendor/admin-console-customize-config-screen.md
paigecalvert: I wrote:

>If you change the `default` value in a later release of your application, installed instances of your application receive the updated value only if your users did not change the default from what it was when they initially installed the application.
>
>If a user did change a field from its `default`, the admin console does not overwrite the value they provided.

AmberAlston:
@paigecalvert Changes look good. How are we handling redirects on the existing page (https://docs.replicated.com/vendor/admin-console-customize-config-screen) since the url changed in the new section updates? (I'm thinking the kots to new site mapping, and anyone that linked to the new site page more recently)

paigecalvert:
@AmberAlston the URL did not change: we are just adding two new topics. So, this won't create any 404s. Should not need a redirect here

AmberAlston (approved):

PR #351, opened 2022-05-09
Make commands work when copy/pasting

ajp-io:


ajp-io:
At some point we should probably go through all the commands, make sure they work, and standardize on where/when we use curl brackets. We were using them every before, but now I'm removing them here since they didn't work. Maybe we should remove them everywhere, but I don't have time right now to test it all out.

paigecalvert (approved):
thank you @ajp-io

PR #353, opened 2022-05-11
kurl release notes for v2022.05.10-0 and v2022.05.11-0

laverya:


ajp-io:
We should add the AWS add-on doc so that we can link to it in these release notes. That can be separate from this PR as long as it's done sooner than later.

ajp-io (approved):

paigecalvert (approved):
Thanks @laverya, lgtm.

And agreed @ajp-io. I created a draft of the AWS add-on topic here: https://github.com/replicatedhq/kurl.sh/pull/784. It just needs SME input on the requirements and limitations.

(Alternatively, we could remove that Requirements and Limitations section for now in order to get the doc out with the release, then add it in when Jalaja or another SME is able to fill it out cc @JustinN-Replicated )

PR #359, opened 2022-05-18
42804: Update gitops-workflow.md

jonquil2002:
SC: https://app.shortcut.com/replicated/story/42804/gitops-workflow-topic-is-incorrect

Preview: https://deploy-preview-359--replicated-docs.netlify.app/enterprise/gitops-workflow

### DIFF
--- a/docs/enterprise/gitops-workflow.md
+++ b/docs/enterprise/gitops-workflow.md
@@ -79,4 +79,4 @@ Subsequently, the admin console makes separate commits with any available update
 
 ## Automatic updates
 
-If you configure [automatic updates](updating-apps#configure-automatic-updates), any updates from your vendor are automatically committed to your Git repository and deployed.
+If you configure [automatic updates](updating-apps#configure-automatic-updates), any updates from your vendor are automatically committed to your Git repository.
--- a/docs/enterprise/gitops-workflow.md
+++ b/docs/enterprise/gitops-workflow.md
@@ -1,6 +1,6 @@
 # Pushing Updates to a GitOps Workflow
 
-The Replicated admin console default workflow is configured to receive updates, show the changes, and deploy the updates to the cluster. You can enable a GitOps workflow instead. When using a GitOps workflow, changes from the admin console are pushed to a private Git repository, where an existing CI/CD process can execute the delivery of manifests to the cluster. Admin console changes can include local configuration changes such as password, upstream updates, license updates, and so on.
+The Replicated admin console default workflow is configured to receive updates, show the changes, and deploy the updates to the cluster. You can enable a GitOps workflow instead. When using a GitOps workflow, changes from the admin console are pushed to a private Git repository, where an existing CI/CD process can execute the delivery of manifests to the cluster. Admin console changes can include local configuration changes (such as password), and upstream updates (such as application and license updates) from your vendor.
 
 If you have more than one application installed, you can selectively enable a GitOps workflow for each application.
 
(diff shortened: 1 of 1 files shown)


docs/enterprise/gitops-workflow.md
```diff
@@ -1,6 +1,6 @@
 # Pushing Updates to a GitOps Workflow
 
-The Replicated admin console default workflow is configured to receive updates, show the changes, and deploy the updates to the cluster. You can enable a GitOps workflow instead. When using a GitOps workflow, changes from the admin console (configuration changes, upstream updates, license updates) are pushed to a private Git repository, where an existing CI/CD process can execute the delivery of manifests to the cluster.
+The Replicated admin console default workflow is configured to receive updates, show the changes, and deploy the updates to the cluster. You can enable a GitOps workflow instead. When using a GitOps workflow, changes from the admin console are pushed to a private Git repository, where an existing CI/CD process can execute the delivery of manifests to the cluster. Admin console changes can include local configuration changes (such as password), and upstream updates from your vendor (such as application and license updates).
```
ajp-io, line 3: What password is this talking about? I assume it's an example of a config item rather than the admin console password. IMO it's more confusing to mention it since it's ambiguous, so I think leaving it at "local configuration changes" is fine, rather than giving an example.

docs/enterprise/gitops-workflow.md
```diff
@@ -76,3 +76,7 @@ To enable pushing updates to a GitOps workflow:
 
 After converting to GitOps, the admin console makes your first commit with the current version that is deployed.
 Subsequently, the admin console makes separate commits with any available updates that have not been deployed from the admin console.
+
+## Automatic updates
+
+If you configure [automatic updates](updating-apps#configure-automatic-updates), any release updates from your vendor are automatically committed to your Git repository and deployed.
```
ajp-io, line 82: The word "deployed" at the end is not accurate. I'm not an expert on this feature, but configuring automatic updates should automatically commit any new versions to the Git repository, but that's it. Deployment is driven by the end user's process, which we can't control.

I don't know how I feel about the wording "release updates". Is this meant to say the same thing as "Admin console changes" above?

docs/enterprise/gitops-workflow.md
jonquil2002: Good suggestion, will implement, thanks!

docs/enterprise/gitops-workflow.md
jonquil2002: @ajp-io I'm not sure about either of these. "Release updates" and "deployed" since both were mentioned in the demo video.

@amit-schnitzer  Would you please clarify what is meant in your GitOps demo video for Alex's questions regarding "deployed" and "release updates"? Being reasonably new to this product still, I am not sure. I was under the impression that a release update could include a change to the actual application manifests, not just changes to the admin console.

ajp-io: Yeah, it does mean changes to the actual application manifests, if I understand the usage here correctly. I actually think "Admin console changes" above is ambiguous and should just be "Changes" and get rid of the admin console thing. Because it isn't changes to the admin console, it's changes that originate in the admin console. And we already specified that in an earlier sentence.

jonquil2002: @ajp-io  Thanks for that clarification - that makes sense! And I'll wait on @amit-schnitzer  response to "deployed". Because my understanding is that this is the end-user's Git repository so if they have commits configured to automatically build as part of their workflow, then it will deploy. If that's true, then maybe that clarification needs to be added as well.

jonquil2002: @amit-schnitzer  Can you confirm whether "deployed" is accurate in this proposed docs change for GitOps workflow? "If you configure [automatic updates](updating-apps#configure-automatic-updates), any release updates from your vendor are automatically committed to your Git repository and deployed."

I based this off of what you said in your demo, but I may have misunderstood. Thanks.

paigecalvert (approved):

PR #361, opened 2022-05-18
48533 How to add a helm chart to a release

paigecalvert:
**This PR changes a filename and requires a redirect before it can be merged.** Redirect story for SRE team: https://app.shortcut.com/replicated/story/48726/redirect-for-helm-docs

Shortcut: https://app.shortcut.com/replicated/story/48533/better-document-how-vendors-add-helm-charts-to-a-release

Preview: https://deploy-preview-361--replicated-docs.netlify.app/vendor/helm-release

For reference, here's the old topic that this replaces: https://docs.replicated.com/vendor/helm-installing-replicated-helm

### DIFF
--- a/docs/vendor/helm-release.md
+++ b/docs/vendor/helm-release.md
@@ -1,3 +1,5 @@
+> This topic applies to both native Helm and Replicated Helm installations.
+> 
 # Adding Helm Charts to a Release
 
 This topic describes how to add a Helm chart to a release in the Replicated vendor portal. Adding a Helm chart to a release is useful if you want to distribute a Kubernetes application with Replicated that is already packaged using Helm.
@@ -20,7 +22,7 @@ The following table provides more information about these files:
   <th>Description</th>
 </tr>
 <tr>
-  <td>HelmChart custom resource manifest file</td>
+  <td>HelmChart CRD <br> <i>chart-name.yaml</i></td>
   <td>A HelmChart custom resource is a YAML file with <code>kind: HelmChart</code>. Replicated automatically creates a HelmChart custom resource for each Helm chart that you add to a release.
   <br/>
   <br/>
--- a/docs/vendor/helm-release.md
+++ b/docs/vendor/helm-release.md
@@ -21,10 +21,10 @@ The following table provides more information about these files:
 </tr>
 <tr>
   <td>HelmChart custom resource manifest file</td>
-  <td>The HelmChart custom resource manifest is a YAML file with <code>kind: HelmChart</code>. Replicated automatically creates the HelmChart custom resource when you add a Helm chart to a release.
+  <td>A HelmChart custom resource is a YAML file with <code>kind: HelmChart</code>. Replicated automatically creates a HelmChart custom resource for each Helm chart that you add to a release.
   <br/>
   <br/>
-  You must include a HelmChart custom resource to enable the Replicated app manager to process and deploy Helm charts. The HelmChart custom resource references the <code>.tgz</code> export of the Helm chart and provides the necessary instructions for processing and preparing the chart for deployment.
+  The HelmChart custom resource references the <code>.tgz</code> export of the Helm chart and provides the necessary instructions to the Replicated app manager for processing and preparing the chart for deployment.
   <br/>
   <br/>
   For more information, see <a href="../reference/custom-resource-helmchart">HelmChart</a> in the <em>Custom Resources</em> section.</td>
@@ -49,11 +49,13 @@ To include a Helm chart in a release, first you package the Helm chart, includin
 
 There are different steps for creating the Helm chart package depending on if the Helm chart source is in a remote chart repository such as Artifact Hub, or in a local directory.
 
+For more information about the Helm CLI commands in this procedure, see the [Helm Commands](https://helm.sh/docs/helm/helm/) section in the Helm documentation.
+
 To package a Helm chart and add it to a release:
 
 1. If the Helm chart source is in a chart repository, do the following:
 
-   1. Update your local directory with the latest available Helm chart information from the chart repositories. Run:
+   1. Update your local directory with the latest available Helm chart information from your chart repositories. Run:
 
       ```
       helm repo update
@@ -61,7 +63,7 @@ To package a Helm chart and add it to a release:
       :::note
       You can also pass the names of a specific repository or repositories that you want to update in the `helm repo update` command. For more information, see [Helm Repo Update](https://helm.sh/docs/helm/helm_repo_update/) in the Helm documentation.
       :::
-   1. Download the latest copy of the desired Helm chart from the repository. Run:
+   1. Download the latest copy of the desired Helm chart from a repository. Run:
 
       ```
       helm fetch REPO_NAME/CHART_NAME
@@ -112,6 +114,8 @@ To package a Helm chart and add it to a release:
           ```
           Replace:
           * `SEQUENCE` with the release sequence number. This identifies the existing release to be updated.
-          * `PATH-TO-APP-DIRECTORY` with the path to the local directory where you copied the Helm chart `.tgz`.     
+          * `PATH-TO-APP-DIRECTORY` with the path to the local directory where you copied the Helm chart `.tgz`.
+
+    For more information, see [release create](../reference/replicated-cli-release-create) and [release update](../reference/replicated-cli-release-update) in the replicated CLI documentation.     
 
 1. Save and promote the release to a development environment to test your changes.
--- a/docs/vendor/helm-release.md
+++ b/docs/vendor/helm-release.md
@@ -1,10 +1,8 @@
-> This topic applies to both native Helm and Replicated Helm installations.
-> 
 # Adding Helm Charts to a Release
 
 This topic describes how to add a Helm chart to a release in the Replicated vendor portal. Adding a Helm chart to a release is useful if you want to distribute a Kubernetes application with Replicated that is already packaged using Helm.
 
-For more information, see [Helm Overview](helm-overview).
+This procedure applies for Replicated Helm installations and native Helm installations. For more information, see [Helm Overview](helm-overview).
 
 ## About Releasing with Helm Charts
 
@@ -12,7 +10,7 @@ You can add one or more Helm charts to a release in the vendor portal by uploadi
 
 When you add a Helm chart to a release, Replicated adds a copy of the `Chart.yaml` file and the `values.yaml` file from the Helm chart to the release.
 
-Replicated also automatically creates a HelmChart custom resource manifest file in the release for each Helm chart that you add. You use this HelmChart custom resource to configure the Helm chart with your application.
+You must also have a HelmChart custom resource manifest file in the release for each Helm chart that you add. You use this HelmChart custom resource to configure the Helm chart with your application.
 
 The following table provides more information about these files:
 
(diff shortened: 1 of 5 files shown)


docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+      ```
+      :::note
+      You can also pass the names of a specific repository or repositories that you want to update in the `helm repo update` command. For more information, see [Helm Repo Update](https://helm.sh/docs/helm/helm_repo_update/) in the Helm documentation.
+      :::
+   1. Download the latest copy of the desired Helm chart from a repository. Run:
+
+      ```
+      helm fetch REPO_NAME/CHART_NAME
+      ```
+      Replace:
```
paigecalvert, line 71: @jonquil2002 I tried using our new table style idea for this, but it ended up feeling a little cumbersome to right out the HTML each time, and looked a little noisy in the rendered markdown. What do you think about something like this with the bulleted list using Replace/With?

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
+# Adding Helm Charts to a Release
+
+This topic describes how to add a Helm chart to a release in the Replicated vendor portal. Adding a Helm chart to a release is useful if you want to distribute a Kubernetes application with Replicated that is already packaged using Helm.
```
jonquil2002, line 3: I think we're still supporting the two types of Helm - Replicated Helm and native Helm? Do you want to clarify that this topic applies to both types? Not sure what you and Justin discussed.

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+
+For more information, see [Helm Overview](helm-overview).
+
+## About Releasing with Helm Charts
+
+You can add one or more Helm charts to a release in the vendor portal by uploading each Helm chart as a `.tgz` package. This allows you to use Replicated to distribute applications that are packaged with Helm.
+
+When you add a Helm chart to a release, Replicated adds a copy of the `Chart.yaml` file and the `values.yaml` file from the Helm chart to the release.
+
+Replicated also automatically creates a HelmChart custom resource manifest file in the release for each Helm chart that you add. You use this HelmChart custom resource to configure the Helm chart with your application.
```
jonquil2002, line 13: Fernando mentioned something about this being created automatically only if they're using the vendor portal, but if they're using the CLI they have to create it manually. You might want to verify that with Justin.

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+</tr>
+</table>
+
+For example, the following screenshot shows how a Postgres Helm chart displays in the file tree of a release in the vendor portal:
+
+![Postgres Helm Chart](/images/postgres-helm-chart.png)
+
+## Add a Helm Chart to a Release
+
+To include a Helm chart in a release, first you package the Helm chart, including any of its dependencies, as a `.tgz` file. Then, you add the `.tgz` Helm chart package to your release.
```
jonquil2002, line 48: It's a personal preference, but you could consider removing "you" from either of these sentences, perhaps the second sentence.

docs/vendor/helm-release.md
jonquil2002: This looks good. Would we still want to do the table with a longer list, or always use this bulleted list no matter how many items are in it? Food for thought. Still reviewing the rest of it :)

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+   1. Download the latest copy of the desired Helm chart from a repository. Run:
+
+      ```
+      helm fetch REPO_NAME/CHART_NAME
+      ```
+      Replace:
+      * `REPO_NAME` with the name of the repository where the Helm chart is located.
+      * `CHART_NAME` with the name of the Helm chart as it appears in the repository.
+
+      The Helm chart, including any dependencies, is packaged and copied to your current directory in a `.tgz` file. The file uses the naming convention: `chartname-version.tgz`. For example, `postgresql-8.1.2.tgz`.
```
jonquil2002, line 75: s/ `chartname-version.tgz` / `CHART_NAME-VERSION.tgz`

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+
+      1. In the [vendor portal](https://vendor.replicated.com), click **Releases**. Then, either click **Create release** to create a new release, or click **View YAML** to edit an existing release.
+
+      1. Drag and drop the Helm chart `.tgz` package from your local directory to the file tree in the release in the vendor portal.
+
+   * **From the replicated CLI**:
+
+      1. Copy the Helm chart `.tgz` package to the local directory that contains the manifest files for the release.
+
+      1. Create a new release or update an existing release:
```
jonquil2002, line 105: This substep could be broken into two substeps, which might be easier to scan.

1. To create a new release, run:

          ```
          replicated release create --yaml-dir PATH-TO-APP-DIRECTORY
          ```
          Replace `PATH-TO-APP-DIRECTORY` with the path to the local directory where you copied the Helm chart `.tgz`.

1. To update an existing release, run:
          ```
          replicated release update SEQUENCE --yaml-dir PATH-TO-APP-DIRECTORY
          ```
          Replace:
          * `SEQUENCE` with the release sequence number. This identifies the existing release to be updated.
          * `PATH-TO-APP-DIRECTORY` with the path to the local directory where you copied the Helm chart `.tgz`.

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+          ```
+          Replace:
+          * `SEQUENCE` with the release sequence number. This identifies the existing release to be updated.
+          * `PATH-TO-APP-DIRECTORY` with the path to the local directory where you copied the Helm chart `.tgz`.
+
+    For more information, see [release create](../reference/replicated-cli-release-create) and [release update](../reference/replicated-cli-release-update) in the replicated CLI documentation.     
+
+1. Save and promote the release to a development environment to test your changes.
+
+## Next Steps
```
jonquil2002, line 123: Since the first recommended step is to connect to a private registry, might you want to add this here?

docs/vendor/helm-release.md
JustinN-Replicated: So this step is the same for both (IE any vendor who wants to install with either method must get the helm-chart.tgz bundle and get it into vendor portal either via drag and drop or via cli).

docs/vendor/helm-release.md
JustinN-Replicated: That is correct

JustinN-Replicated: Actually, I am not 100% sure now. Looking in my lab, it does not look like the crd is created
image
though I do know that my app (https://vendor.replicated.com/apps/justin-nginx/) deploys successfully. I may need to dig into this more....

JustinN-Replicated: My bad, the `nginx-ingress.yaml` is the helmchart crd created for me when I drag/dropped into vendor portal. This file is not created automatically when using cli.

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+
+After you add one or more Helm charts to a release, Replicated recommends that you update the HelmChart custom resource manifest files to continue configuring your application:
+
+* For information about collecting user-supplied values from the Replicated admin console configuration screen and mapping those values to your Helm chart, see [Creating and Editing Configuration Fields](admin-console-customize-config-screen).
```
jonquil2002, line 127: Since these are next steps, do we want to shorten the wording like this?

* To collect user-supplied values from the..., see [Creating and Editing Configuration Fields](admin-console-customize-config-screen)

* To do blah blah blah, see [crosss-reference]

Or just like this:

* [Creating and Editing Configuration Fields](admin-console-customize-config-screen)

* [Installing with Native Helm](helm-installing-native-helm)

docs/vendor/helm-release.md
```diff
@@ -0,0 +1,130 @@
...
+          * `SEQUENCE` with the release sequence number. This identifies the existing release to be updated.
+          * `PATH-TO-APP-DIRECTORY` with the path to the local directory where you copied the Helm chart `.tgz`.
+
+    For more information, see [release create](../reference/replicated-cli-release-create) and [release update](../reference/replicated-cli-release-update) in the replicated CLI documentation.     
+
+1. Save and promote the release to a development environment to test your changes.
+
+## Next Steps
+
+After you add one or more Helm charts to a release, Replicated recommends that you update the HelmChart custom resource manifest files to continue configuring your application:
```
jonquil2002, line 125: Word list decision time: do we want to go with "Replicated recommends" or "we recommend"? I have no preference, just want us to be consistent.

jonquil2002 (approved):
A few suggestions and comments. But LGTM overall.

docs/vendor/helm-release.md
jonquil2002: Yes, I understand. But do we want to spell that out for the vendor, so that they understand that it applies to both Helm types? Perhaps it's not obvious enough, given how confusing having two types might be?

JustinN-Replicated: Will push a commit and see what ya'll think - I could see us mentioning something that alludes to this though (IE you must/should do this regardless of installation type)

docs/vendor/helm-release.md
paigecalvert: Good question! We should add this to the our style guide decision spreadsheet

docs/vendor/helm-release.md
paigecalvert: Updated this to clarify it's only automatically created when you do the drag and drop method. Also clarified the naming convention that Replicated uses when we automatically create it. (Edited your commit a bit Justin because it doesn't always have to use that naming convention i.e., if they create it themselves or change it)

docs/vendor/helm-release.md
paigecalvert: Good call

docs/vendor/helm-release.md
paigecalvert: I think I'll leave it out because those recommended topics are intentionally unordered. There's lots you can do with the HelmChart custom resource from here, so wanted to group the options together in no particular order so that people can choose what they do next (depending on what they've already done).

Maybe it would be good for me to revisit this list to consider only recommending the config screen step, since I don't think there's any way they would have done that before adding this Helm chart, but not sure.

docs/vendor/helm-release.md
paigecalvert: Good call, will do 👍 I'll keep some context as in your first suggestion, because I don't think the topic names alone express all the info that would be valuable for someone deciding what they want to do next

docs/vendor/helm-release.md
paigecalvert: Ooo yes yes, this is a good one to decide on. Want to add it to the style decision spreadsheet?

docs/vendor/helm-release.md
paigecalvert: Darn I keep doing this! I even thought this time "remember not to use 'you'" 😄

jonquil2002 (approved):
LGTM

JustinN-Replicated (dismissed):
lgtm

paigecalvert:
Chase is pushing out the redirect, so I'll merge this shortly. Thanks, everyone!

PR #362, opened 2022-05-18
48530: New Application Button

jonquil2002:
SC: 

Preview: https://deploy-preview-362--replicated-docs.netlify.app/vendor/releases-creating-releases

### DIFF
--- a/docs/vendor/releases-creating-releases.md
+++ b/docs/vendor/releases-creating-releases.md
@@ -10,7 +10,7 @@ We recommend that you bookmark the Replicated [vendor portal](https://vendor.rep
 
     - If you are logging in with an existing account and want to create a new application, click **Create new app** from the Applications drop-down list.
 
-    - If you want to update an existing application, you can use the search filter from the Applications drop-down list to find the application.
+    - If you want to update an existing application, select it from the Applications tab. If you have multiple applications, a search filter appears at the top of the list or you can select the application from the drop-down list.
 
   The Channels page for the release opens and displays a list of your release channels, which are logical stacks for you to stage and promote releases to your customers.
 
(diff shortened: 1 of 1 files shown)


paigecalvert (approved):
lgtm -- just added one suggestion

ajp-io (changes requested):

jonquil2002:
@paigecalvert  @ajp-io  Ready for another review pass. Good ideas, thanks!

ajp-io (approved):
@miaawong These docs should get merged when we ship the updated application dropdown and new create app button.

docs/vendor/releases-creating-releases.md
```diff
@@ -2,21 +2,23 @@
...
+1. Log in (or create a new team) on the vendor portal and do one of the following:
 
-  **Note:** If you are logging in with an existing account, you are not prompted to create a new application. Instead, click https://vendor.replicated.com/new-application to create a new application.
+    - If you are activating a new account: In the prompt, name your application and click **Create application**.
 
-  ![Create Application](/images/guides/kots/create-application.png)
+    ![Create Application](/images/guides/kots/create-application.png)
 
-  The Channels page opens and displays a list of your release channels, which are logical stacks for you to stage and promote releases to your customers.
+    - If you are logging in with an existing account: Click **New Application** or select an existing application.
```
ajp-io, line 11: We decided against a dedicated New Application button, so we moved this back into the dropdown. So creating a new app is in the dropdown again.

ajp-io (commented):
Thanks for resurfacing this PR. We decided against a dedicated New Application button, so that part needs to change to mention how it's in the dropdown (as before). But if you think the rest of the changes here are worth making, then we can definitely get this approved and merged.

jonquil2002:
Closing this PR because the feature was archived and the other non-feature changes are not needed due to another PR from last week.

ajp-io:
@jonquil2002 One thing I liked in this PR was that instead of referencing a random link in the first step [here](https://docs.replicated.com/vendor/releases-creating-releases), we talk about how to find the New Application button. We. can make this change separately, but I would rather tell people to use the standard button instead of this random link. This link isn't accessible anywhere in the product—like you can't click something in the product to get to that. So I'd rather show people how they can do this in their normal workflow, rather than using a random link they have to bookmark.

PR #381, opened 2022-06-01
47495: Update Enterprise Installing an Embedded Cluster

jonquil2002:
SC: https://app.shortcut.com/replicated/story/47495/style-procedure-improvements-to-installing-with-the-kubernetes-installer

Preview: https://deploy-preview-381--replicated-docs.netlify.app/enterprise/installing-embedded-cluster

### DIFF
--- a/docs/enterprise/installing-embedded-cluster.md
+++ b/docs/enterprise/installing-embedded-cluster.md
@@ -13,7 +13,7 @@ Complete the following items before you perform this task:
 
 ## Install in an Online Environment
 
-To install the admin console on a cluster created by the Kubernetes installer, run the installation script provided by the application vendor.
+To install an application and the admin console on a cluster created by the Kubernetes installer, run the installation script provided by the application vendor.
 
 **Example:**
 
@@ -28,10 +28,10 @@ With KOTS v1.67.0 and later, you can install a specific version of the applicati
 
 ## Install in an Air Gapped Environment
 
-To install in an air gapped environment:
+To install an application and the admin console in an air gapped environment:
 
 1. Download and extract the kURL air gap `.tar.gz` file. The air gap `.tar.gz` includes only the admin console components, which are required to install the application.
-1. Run the install.sh script:
+1. Run the `install.sh` script:
 
     ```bash
     curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
@@ -42,7 +42,7 @@ To install in an air gapped environment:
     Replace `FILENAME` with the name of the kURL air gap `.tar.gz` file.
 
     :::note
-    You can construct the URL for the air gap bundle by prefixing the URL path for online installations, as described in [Install in an Online Environment](#install-in-an-online-environment) above, with `/bundle` and adding `.tar.gz` to the end.
+    You can construct the URL for the air gap bundle by prefixing the URL path for online installations with `/bundle` and adding `.tar.gz` to the end. For more information, see [Install in an Online Environment](#install-in-an-online-environment).
     :::
 
 1. Install the application with the application `.airgap` bundle:
--- a/docs/enterprise/installing-embedded-cluster.md
+++ b/docs/enterprise/installing-embedded-cluster.md
@@ -117,7 +117,7 @@ To add primary and secondary nodes:
   curl -sSL https://kurl.sh/my-test-app-2-unstable/join.sh | sudo bash -s \
   kubernetes-master-address=192.0.2.0:6443 \
   kubeadm-token=8z0hjv.s9wru9z \
-  kubeadm-token-ca-hash=sha256:289f5c0d61775edec20d4e980602deeeeeeeeeeeeeeeeeffffffffffggggggg \
+  kubeadm-token-ca-hash=sha256:289f5c0d61775edec20d4e980602deeeeeeeeeeeeeeeeffffffffffggggggg \
   docker-registry-ip=198.51.100.3 \
   kubernetes-version=v1.19.16 \
   primary-host=203.0.113.6
--- a/docs/enterprise/installing-embedded-cluster.md
+++ b/docs/enterprise/installing-embedded-cluster.md
@@ -1,6 +1,6 @@
 # Installing with the Kubernetes Installer
 
-This sections explains how to install an application and the Replicated admin console on a cluster created by the Replicated Kubernetes installer.
+This section explains how to install an application and the Replicated admin console on a cluster created by the Replicated Kubernetes installer.
 
 The Kubernetes installer is based on the open source kURL project, which is maintained by Replicated. For more information about installing with kURL, including advanced installation options, see the [kURL documentation](https://kurl.sh/docs/introduction/).
 
--- a/docs/enterprise/installing-embedded-cluster.md
+++ b/docs/enterprise/installing-embedded-cluster.md
@@ -90,7 +90,10 @@ To install with high availability in an online environment:
       load-balancer-address=HOST:PORT
       ```
 
-      Replace `HOST:PORT` with the host and port number of the external load balancer.
+      Replace:
+
+      - `HOST` with the external load balancer host.
+      - `PORT` with the port number of the external load balancer.
 
 1. Using the admin console URL and password that displays when the installation script finishes, upload the license file and install the application.
 
@@ -122,7 +125,10 @@ To install with high availability in an air gapped environment:
       load-balancer-address=HOST:PORT
       ```
 
-      Replace `HOST:PORT` with the host and port number of the external load balancer.
+      Replace:
+
+      - `HOST` with the external load balancer host.
+      - `PORT` with the port number of the external load balancer.
 
 1. Install the application with the application `.airgap` bundle:
 
--- a/docs/enterprise/installing-embedded-cluster.md
+++ b/docs/enterprise/installing-embedded-cluster.md
@@ -147,7 +147,7 @@ To install with high availability in an air gapped environment:
 
 You can generate commands in the admin console to join additional primary and secondary nodes to the cluster. Primary nodes run services that control the cluster. Secondary nodes run services that control the pods that host the application containers. Adding nodes can help manage resources to ensure that your application runs smoothly.
 
-For high availability clusters, Kubernetes recommends using at least 3 primary nodes, and that you use an odd number to help with leader selection if machine or zone failure occurs. For more information, see [Creating Highly Available Clusters with kubeadm](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/) in the Kubernetes documentation.
+For high availability clusters, Kubernetes recommends using at least 3 primary nodes, and that you use an odd number of nodes to help with leader selection if machine or zone failure occurs. For more information, see [Creating Highly Available Clusters with kubeadm](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/) in the Kubernetes documentation.
 
 To add primary and secondary nodes:
 
--- a/docs/enterprise/installing-embedded-cluster.md
+++ b/docs/enterprise/installing-embedded-cluster.md
@@ -1,6 +1,6 @@
 # Installing with the Kubernetes Installer
 
-This section explains how to install an application and the Replicated admin console on a cluster created by the Replicated Kubernetes installer.
+This document explains how to install an application and the Replicated admin console on a cluster created by the Replicated Kubernetes installer.
 
 The Kubernetes installer is based on the open source kURL project, which is maintained by Replicated. For more information about installing with kURL, including advanced installation options, see the [kURL documentation](https://kurl.sh/docs/introduction/).
 
@@ -8,25 +8,26 @@ The Kubernetes installer is based on the open source kURL project, which is main
 
 Both online and air gap installations can be installed in high availability (HA) mode with the Kubernetes installer.
 
-When installing a cluster with high availablity, the script prompts for a load balancer address. In the absence of a load balancer, all traffic is routed to the first primary node.
+We recommend that you use a load balancer. If your vendor has provided an internal load balancer with the EKCO add-on, you do not need to provide your own external load balancer. An external load balancer can be preferred when clients outside the cluster need access to the cluster's Kubernetes API.
 
-If you decide to use a load balancer, the load balancer must be a TCP forwarding load balancer. For more information, see [Prerequisites](#prerequisites).
+In the absence of any load balancer, all traffic is routed to the first primary node.
 
-The health check for an apiserver is a TCP check on the port that the kube-apiserver listens on. The default value is `:6443`.
+If you decide to use a external load balancer, the external load balancer must be a TCP forwarding load balancer. For more information, see [Prerequisites](#prerequisites).
 
-For more information about the kube-apiserver load balancer, see [Create load balancer for kube-apiserver](https://kubernetes.io/docs/setup/independent/high-availability/#create-load-balancer-for-kube-apiserver) in the Kubernetes documentation.
+The health check for an apiserver is a TCP check on the port that the kube-apiserver listens on. The default value is `:6443`. For more information about the kube-apiserver external load balancer, see [Create load balancer for kube-apiserver](https://kubernetes.io/docs/setup/independent/high-availability/#create-load-balancer-for-kube-apiserver) in the Kubernetes documentation.
 
 ## Prerequisites
 
 Complete the following before you perform this task:
 
-- Meet the general system requirements. See [General System Requirements](installing-general-requirements)
-- Meet the system requirements for the Kubernetes installer. See [Kubernetes Installer Requirements](installing-embedded-cluster-requirements)
+- Meet the general system requirements. See [General System Requirements](installing-general-requirements).
+- Meet the system requirements for the Kubernetes installer. See [Kubernetes Installer Requirements](installing-embedded-cluster-requirements).
+- (Optional) To preconfigure the internal load balancer, pass the `ekco-enable-internal-load-balancer` flag.
+- (Optional) To preconfigure an external load balancer, pass the `load-balancer-address=HOST:PORT` flag.
 
-Additionally, if you are installing with high availability and want to use a load balancer, ensure that your load balancer is:
-- A TCP forwarding load balancer
-- Configured to distribute traffic to all healthy control plane nodes in its target list
-- (Optional) Preconfigured by passing in the `load-balancer-address=<host:port>` flag
+Additionally, if you are installing with high availability and want to use an external load balancer, ensure that your load balancer is:
+- A TCP forwarding external load balancer.
+- Configured to distribute traffic to all healthy control plane nodes in its target list.
 
 ## Install in an Online Environment
 
@@ -35,20 +36,34 @@ To install an application and the admin console in an online environment, do one
 - To install without high availability, run the installation command provided by the application vendor:
 
   ```bash
-  curl -sSL https://k8s.kurl.sh/APP-SLUG | sudo bash
+  curl -sSL https://k8s.kurl.sh/APP_SLUG | sudo bash
   ```
   Replace `APP-SLUG` with the unique slug for the application. The application slug is included in the installation command provided by the vendor.
 
-- To install with high availability, run the installation command with the `-s ha` flag:
+- To install with high availability:
 
-  ```bash
-  curl -sSL https://k8s.kurl.sh/APP-SLUG | sudo bash -s ha
-  ```
-  Replace `APP-SLUG` with the unique slug for the application. The application slug is included in the installation script provided by the vendor.
+    1. Run the installation command with the `-s ha` flag:
 
-:::note
-With KOTS v1.67.0 and later, you can install a specific version of the application. Use the `app-version-label` flag and the version label for a particular version of your vendor's application. For example, `curl https://kurl.sh/supergoodtool | sudo bash -s app-version-label=3.1.0`.
-:::
+      ```bash
+      curl -sSL https://k8s.kurl.sh/APP_SLUG | sudo bash -s ha
+      ```
+      Replace `APP-SLUG` with the unique slug for the application. The application slug is included in the installation script provided by the vendor.
+
+    1. If you did not preconfigure a load balancer, you are prompted during the installation. Do one of the following:
+
+        - If you are using the internal load balancer, pass the following flag:
+
+          ```bash
+          ekco-enable-internal-load-balancer
+          ```
+
+        - If you are using an external load balancer, pass the following flag:
+
+          ```bash
+          load-balancer-address=HOST:PORT
+          ```
+
+          Replace `HOST:PORT` with the host and port number of the external load balancer.
 
 ## Install in an Air Gapped Environment
 
@@ -67,7 +82,9 @@ To install an application and the admin console in an air gapped environment:
 
       Replace `FILENAME` with the name of the kURL air gap `.tar.gz` file.
 
-    - To install with high availability, run:
+    - To install with high availability:
+
+    1. Run the following commands:
 
       ```bash
       curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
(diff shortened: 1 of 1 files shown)


docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
-This article refers to installing an application and the Replicated admin console on a cluster created by the Replicated Kubernetes installer.
+This section explains how to install an application and the Replicated admin console on a cluster created by the Replicated Kubernetes installer.
 
 The Kubernetes installer is based on the open source kURL project, which is maintained by Replicated. For more information about installing with kURL, including advanced installation options, see the [kURL documentation](https://kurl.sh/docs/introduction/).
 
-## Requirements
+## Prerequisites
 
-Before you install, ensure that you meet the system requirements. For more information, see [Kubernetes installer requirements](installing-embedded-cluster-requirements).
+Complete the following items before you perform this task:
```
paigecalvert, line 9: s/ could remove the word "items"

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
-When installing on a highly available cluster, the script will prompt for a load balancer address.
-The load balancer can be preconfigured by passing in the `load-balancer-address=<host:port>` flag.
+Both online and air gap installations can be installed in high availability (HA) mode with the Kubernetes installer.
 
-This load balancer should be configured to distribute traffic to all healthy control plane nodes in its target list.
-This should be a TCP forwarding load balancer.
+When installing on a highly available cluster, the script prompts for a load balancer address. In the absence of a load balancer, all traffic is routed to the first primary node.
 
-The health check for an apiserver is a TCP check on the port the kube-apiserver listens on (default value :6443).
+If you decide to use a load balancer, the load balancer should be:
```
paigecalvert, line 73: should --> must ?

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
-### Online
+For more information about the kube-apiserver load balancer, see [Create load balancer for kube-apiserver](https://kubernetes.io/docs/setup/independent/high-availability/#create-load-balancer-for-kube-apiserver) in the Kubernetes documentation.
 
-To install with high availability in an online environment, run:
+### Install with HA in an Online Environment
+
+To install an application and the admin console with high availability in an online environment, run:
 
 ```bash
 curl -sSL https://kurl.sh/APP-SLUG | sudo bash -s ha
```
paigecalvert, line 88: An idea as I'm reading these procedures back to back: Since the commands for installing with HA are the same just with the added `-ha` flag, maybe these can be turned into optional steps in the main procedures.

For the optional load balancer part, maybe that could be added to the prerequisites section. For example, "If you are installing on a high availability cluster and want to use a load balancer, ensure the your load balancer meets these requirements..."

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
+1. Run the following command:
+
+    ```bash
+    cat install.sh | sudo bash -s airgap ha
+    ```
 
 ## Join Primary and Secondary Nodes
 
-Visit the `/cluster/manage` page in the admin console to generate scripts for joining additional secondary and primary nodes.
+You can generate commands in the admin console to join additional primary and secondary nodes.
```
paigecalvert, line 106: Do you know what the context for this procedure is? It looks like the only other place we mention nodes on this page is where we say that if you don't use a load balancer, traffic goes to the primary node. Could be helpful to clarify the "why" on this section; I wasn't able to sort out how it fits in with installing.

(This could be a spin-off story if you think that's better. I think it would be addressing more of a content gap in terms of the "why")

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
-The health check for an apiserver is a TCP check on the port the kube-apiserver listens on (default value :6443).
+If you decide to use a load balancer, the load balancer should be:
+- A TCP forwarding load balancer
+- Configured to distribute traffic to all healthy control plane nodes in its target list
 
-For more information on the kube-apiserver load balancer see [Create load balancer for kube-apiserver](https://kubernetes.io/docs/setup/independent/high-availability/#create-load-balancer-for-kube-apiserver) in the Kubernetes documentation.
+You can preconfigure the load balancer by passing in the `load-balancer-address=<host:port>` flag.
 
-In the absence of a load balancer, all traffic will be routed to the first primary.
+The health check for an apiserver is a TCP check on the port that the kube-apiserver listens on (the default value is `:6443`).
```
paigecalvert, line 79: s/ The parenthetical phrase could be its own sentence

paigecalvert (approved):
Looks good @jonquil2002 ! Just small notes, and one suggestion around the separate HA procedures

paigecalvert (approved):
Looks great!

One thing I noticed in the air gap HA procedure: 

For air gap, I believe you'll still need to do the `curl` and `tar` commands before the `cat` install command. (I.e., the commands should look identical, with the only difference being that one of them has the -ha flag). The kurl docs have more info: https://kurl.sh/docs/install-with-kurl/#airgapped-usage

jonquil2002:
@paigecalvert  - Yep, that's what I was thinking too and was planning to verify that. It just makes sense that they would, not sure why it wasn't documented originally.

jonquil2002:
@JustinN-Replicated @ajp-io  Would one of you please do a technical review of this updated topic? In addition to heavy reformatting, I added a business reason for joining additional nodes and added the missing command from high availability air gap, so I want to make sure that it's technically correct. Thanks.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
+- Meet the general system requirements. See [General System Requirements](installing-general-requirements)
+- Meet the system requirements for the Kubernetes installer. See [Kubernetes Installer Requirements](installing-embedded-cluster-requirements)
 
-```bash
-curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
-tar xvzf FILENAME.tar.gz
-cat install.sh | sudo bash -s airgap
-```
-Where `FILENAME` is the name of the kURL air gap `.tar.gz` file.
+Additionally, if you are installing with high availability and want to use a load balancer, ensure that your load balancer is:
```
ajp-io, line 26: Throughout this doc, I would refer to this as an external load balancer to distinguish it from the internal load balancer. Since they will almost certainly use a load balancer if they're installing in HA, but these requirements will only be pertinent if they bring their own, not if they use the internal one.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
+- A TCP forwarding load balancer
+- Configured to distribute traffic to all healthy control plane nodes in its target list
+- (Optional) Preconfigured by passing in the `load-balancer-address=<host:port>` flag
```
ajp-io, line 29: Oh great, you mention this here (I mentioned this flag in a comment above). Maybe we should similarly mention pre-configuring the internal load balancer.

ajp-io (changes requested):

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
-You can construct the URL for the air gap bundle by prefixing the URL path for online installations as described in [Install in an Online Environment](#install-in-an-online-environment) above with `/bundle` and adding `.tar.gz` to the end.
-:::
+## Install in an Online Environment
+
+To install in an application and the admin console in an online environment, do one of the following:
+
+- To install without high availability, run the installation command provided by the application vendor:
+
+  ```bash
+  curl -sSL https://kurl.sh/APP-SLUG | sudo bash
```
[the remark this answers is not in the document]
jonquil2002: Thanks! This is an old tutorial, and I think there are many old kurl.sh references throughout the docs. I'll create an SC story to scrub the rest of the docs. But I fixed it for this tutorial.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,124 @@
...
+- To install with high availability, run the installation command with the `-s ha` flag:
+
+  ```bash
+  curl -sSL https://kurl.sh/APP-SLUG | sudo bash -s ha
+  ```
+  Replace `APP-SLUG` with the unique slug for the application. The application slug is included in the installation script provided by the vendor.
 
 :::note
-The air gap `.tar.gz` includes only the admin console components, which are required to install the application.
+With KOTS v1.67.0 and later, you can install a specific version of the application. Use the `app-version-label` flag and the version label for a particular version of your vendor's application. For example, `curl https://kurl.sh/supergoodtool | sudo bash -s app-version-label=3.1.0`.
```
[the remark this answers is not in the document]
jonquil2002: Good call.

jonquil2002:
@ajp-io  Hey, Alex. I made some substantial changes re: the load balancer info and also separated out the previous versions installation info.  I'll put a hold on merging this topic until the other version topic gets merged. Please review this again. Thanks.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
+
+          ```bash
+          ekco-enable-internal-load-balancer
+          ```
+
+        - If you are using an external load balancer, pass the following flag:
+
+          ```bash
+          load-balancer-address=HOST:PORT
+          ```
```
ajp-io, line 63: The flags would be passed to the install command above (e.g., `| sudo bash -s ha ekco-enable-internal-load-balancer`). That's the pre-configuring option. But if they don't preconfigure the load balancer (i.e., they don't pass one of those flags), then they are prompted for a load balancer address. Leaving it blank uses the internal load balancer, or you can pass the address of an external load balancer.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
-To install in an air gapped environment, download the kURL air gap `.tar.gz`, extract it, and run the install.sh script:
+To install an application and the admin console in an air gapped environment:
 
-```bash
-curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
-tar xvzf FILENAME.tar.gz
-cat install.sh | sudo bash -s airgap
-```
-Where `FILENAME` is the name of the kURL air gap `.tar.gz` file.
+1. Download and extract the kURL air gap `.tar.gz` file. The air gap `.tar.gz` includes only the admin console components, which are required to install the application.
```
ajp-io, line 71: I assume this is trying to say that the .tar.gz file doesn't include the application, but it feels a bit incorrect to say that it only includes the admin console components. It includes all the Kubernetes cluster components and the admin console. So if my assessment of what this is trying to say is correct, this might be more accurate.
```suggestion
1. Download and extract the kURL air gap `.tar.gz` file. The air gap `.tar.gz` file only includes the Kubernetes and admin console components. It doesn't include the application.
```

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
-You can construct the URL for the air gap bundle by prefixing the URL path for online installations as described in [Install in an Online Environment](#install-in-an-online-environment) above with `/bundle` and adding `.tar.gz` to the end.
-:::
+    - To install without high availability, run:
 
-:::note
-The air gap `.tar.gz` includes only the admin console components, which are required to install the application.
-:::
+      ```bash
+      curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
+      tar xvzf FILENAME.tar.gz
```
ajp-io, line 78:
```suggestion
      tar -xvzf FILENAME.tar.gz
```

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
-* `PASSWORD` is a shared password.
+    - To install with high availability:
 
-For more information about the `kots install` command, see [install](../reference/kots-cli-install) in the kots CLI documentation.
+      1. Run the following commands:
 
-## Install with High Availability
+        ```bash
+        curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
+        tar xvzf FILENAME.tar.gz
```
ajp-io, line 90:
```suggestion
        tar -xvzf FILENAME.tar.gz
```

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
+        :::note
+        For both installation types, you can construct the URL for the air gap bundle by prefixing the URL path for online installations with `/bundle` and adding `.tar.gz` to the end. For more information, see [Install in an Online Environment](#install-in-an-online-environment).
+        :::
 
-This load balancer should be configured to distribute traffic to all healthy control plane nodes in its target list.
-This should be a TCP forwarding load balancer.
+      1. If you did not preconfigure a load balancer, you are prompted during the installation. Do one of the following:
 
-The health check for an apiserver is a TCP check on the port the kube-apiserver listens on (default value :6443).
+          - If you are using the internal load balancer, pass the following flag:
```
ajp-io, line 102: Same comment here about the flags as above.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
+          ekco-enable-internal-load-balancer
+          ```
+
+        - If you are using an external load balancer, pass the following flag:
+
+          ```bash
+          load-balancer-address=HOST:PORT
+          ```
+
+          Replace `HOST:PORT` with the host and port number of the external load balancer.
```
ajp-io, line 65: At the end, do we need to mention how to install the application? You would just open the admin console and upload the license, but since the air gap instructions mention the `kots install` command for installing the app, should we mention the app here too? Or is this only about installing the cluster and we shouldn't mention the app?

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
+        ```
 
-You can include the `ha` option to install with high availability. Both online and air gapped installations can be configured in high-availability mode.
+        Replace `FILENAME` with the name of the kURL air gap `.tar.gz` file.
 
-When installing on a highly available cluster, the script will prompt for a load balancer address.
-The load balancer can be preconfigured by passing in the `load-balancer-address=<host:port>` flag.
+        :::note
+        For both installation types, you can construct the URL for the air gap bundle by prefixing the URL path for online installations with `/bundle` and adding `.tar.gz` to the end. For more information, see [Install in an Online Environment](#install-in-an-online-environment).
+        :::
```
ajp-io, line 98: This note ends up nested under the "To install with high availability section", which to me makes it unclear that "both installation types" refers to HA and non-HA. It would make more sense if the note wasn't nested within one of those two types.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
+    For more information about the `kots install` command, see [install](../reference/kots-cli-install) in the kots CLI documentation.
 
 ## Join Primary and Secondary Nodes
 
-Visit the `/cluster/manage` page in the admin console to generate scripts for joining additional secondary and primary nodes.
+You can generate commands in the admin console to join additional primary and secondary nodes to the cluster. Primary nodes run services that control the cluster. Secondary nodes run services that control the pods that host the application containers. Adding nodes can help manage resources to ensure that your application runs smoothly.
+
+To add primary and secondary nodes:
+
+1. (Air gap only) Download and extract the `.airgap` bundle on the remote node before running the join command.
```
ajp-io, line 141: I don't think this is actually referring to the application `.airgap` file. I think it is the `.tar.gz` file. The reason is, for air gap, you can't pull the join script from the internet (see how the command below is curling k8s.kurl.sh?). So instead, you have to have the `.tar.gz` file present on the remote node because it contains the `join.sh` script you're going to run.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,178 @@
...
+  curl -sSL https://k8s.kurl.sh/my-test-app-unstable/join.sh | sudo bash -s \
+  kubernetes-master-address=192.0.2.0:6443 \
+  kubeadm-token=8z0hjv.s9wru9z \
+  kubeadm-token-ca-hash=sha256:289f5c0d61775edec20d4e980602deeeeeeeeeeeeeeeeffffffffffggggggg \
+  docker-registry-ip=198.51.100.3 \
+  kubernetes-version=v1.19.16 \
+  primary-host=203.0.113.6
+  ```
+
+## Install Previous Versions in an Online Environment
```
ajp-io, line 157: I'm curious about this part here. This looks like it's more related to the other PR we have in progress about installing previous versions of the application. Are you also adding some context on that as part of this PR? I wasn't sure if this got mixed up with this one or if it's intentional.

ajp-io (changes requested):
One thing we might need a more technical SME for is, when should HA mode be used? Should I use it for single-node installations, or is it only relevant for multi-node installs? I also think there are rules about how many primary nodes you need (maybe it has to be an odd number), so that's also something it would be good to document here. Could be a separate iteration though once we nail this content. Up to you.

docs/enterprise/installing-embedded-cluster.md
jonquil2002: Intentional, as we discussed, and we'll postpone making these changes in a future PR. For now, I restored the notes about --app-version-label

Thank you so much for meeting with me.

docs/enterprise/installing-embedded-cluster.md
ajp-io: Perfectly fine, thanks!

jonquil2002:
@paigecalvert  @ajp-io  The combined HA and non-HA content got confusing to read after adding the load balancing info, so I separated them again. Please approve if possible. Thanks for all of your great feedback!

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,174 @@
...
 
-## Install in an Online Environment
+- Meet the general system requirements. See [General System Requirements](installing-general-requirements).
+- Meet the system requirements for the Kubernetes installer. See [Kubernetes Installer Requirements](installing-embedded-cluster-requirements).
 
-To install the admin console on a cluster created by the Kubernetes installer, run the installation script provided by the application vendor.
+Additionally, if you are installing in high availability mode:
 
-For example:
+  - (Optional) If your vendor has enabled the internal load balancer, you can preconfigure it by passing `| sudo bash -s ha ekco-enable-internal-load-balancer`.
```
ajp-io, line 20: Small thing… The vendor doesn't have to enable it—the flag _is_ what enables it. So it's more like, "If you are going to use the internal load balancer…"

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,174 @@
...
 
-```bash
-curl -sSL https://kurl.sh/APP-SLUG | sudo bash
-```
-Where `APP-SLUG` is the unique slug for the application. The application slug is included in the installation script provided by the vendor.
+  - For an external load balancer, ensure that your load balancer is:
+    - A TCP forwarding external load balancer.
+    - Configured to distribute traffic to all healthy control plane nodes in its target list.
+    - (Optional) Preconfigured with the load balancer address by passing the `load-balancer-address=HOST:PORT` flag.
```
ajp-io, line 25: If you combine the preamble before the colon and this bullet as is, it would read, "ensure that your load balancer is preconfigured with the load balancer address." But the LB doesn't have to be preconfigured with the LB address—the installer does. So I think this is clearer.
```suggestion
    - (Optional) Preconfigured by passing the `load-balancer-address=HOST:PORT` flag.
```

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,174 @@
...
+## Install in Air Gapped Environments
+
+To install the admin console and an application in an air gapped environment:
+
+1. Download and extract the kURL air gap `.tar.gz` file. The file only includes the Kubernetes and admin console components. It does not include the application.
+1. Run the following commands:
+
+    ```bash
+    curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
+    tar -xvzf FILENAME.tar.gz
```
ajp-io, line 50: The `curl` command does the download and the `tar` command here does the extraction, so that is a bit repetitive since step 1 says to download and extract. Maybe these should just be one step, or however you want to resolve that.

ajp-io: Down lower we just say, "Run the following commands," so we could do that again.

docs/enterprise/installing-embedded-cluster.md
jonquil2002: @ajp-io Ok, oh, I thought that the vendor had to include the ECKO add-on and configure it in the manifest on their end. But it sounds like you're saying the internal load balancer is included by default for all Kubernetes installers?

jonquil2002:
@paigecalvert Ready for final review/approval when you get a chance. Thanks!

docs/enterprise/installing-embedded-cluster.md
jonquil2002: I made your change anyway. My question is really to help me understand this feature better. Thanks!

paigecalvert (approved):
Lgtm!

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,173 @@
...
-```bash
-curl -LO https://k8s.kurl.sh/bundle/FILENAME.tar.gz
-tar xvzf FILENAME.tar.gz
-cat install.sh | sudo bash -s airgap
-```
-Where `FILENAME` is the name of the kURL air gap `.tar.gz` file.
+  ```bash
+  curl -sSL https://k8s.kurl.sh/APP_SLUG | sudo bash
+  ```
+  Replace `APP-SLUG` with the unique slug for the application. The application slug is included in the installation command provided by the vendor.
```
ajp-io, line 36:
```suggestion
  Replace `APP_SLUG` with the unique slug for the application. The application slug is included in the installation command provided by the vendor.
```

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,173 @@
...
-In the absence of a load balancer, all traffic will be routed to the first primary.
+    - If you are using the internal load balancer, leave the prompt blank and proceed with the installation.
 
-### Online
+    - If you are using an external load balancer, pass the load balancer address:
 
-To install with high availability in an online environment, run:
+      ```bash
+      load-balancer-address=HOST:PORT
+      ```
```
ajp-io, line 90: So to pass the load balancer address in this case, there is just a prompt that asks you to input the address. So you're not passing a flag like is indicated here. I think we can just remove these three lines (and the lines below about replacing stuff) and tell them to pass the LB address. Let me know if that makes sense. They should know how to do this when they see it—it's just a standard input.

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,173 @@
...
+    - If you are using an external load balancer, pass the load balancer address:
+
+      ```bash
+      load-balancer-address=HOST:PORT
+      ```
+
+      Replace:
+
+      - `HOST` with the external load balancer host.
+      - `PORT` with the port number of the external load balancer.
```
ajp-io, line 130: Same as above. This stuff isn't really necessary because it's not a flag they're passing. They're literally just typing in the address at the prompt.

jonquil2002:
@ajp-io Changes made - ready for approval?

docs/enterprise/installing-embedded-cluster.md
```diff
@@ -1,100 +1,155 @@
...
 
-## Install in an Online Environment
+- Meet the general system requirements. See [General System Requirements](installing-general-requirements).
+- Meet the system requirements for the Kubernetes installer. See [Kubernetes Installer Requirements](installing-embedded-cluster-requirements).
 
-To install the admin console on a cluster created by the Kubernetes installer, run the installation script provided by the application vendor.
+Additionally, if you are installing in high availability mode:
 
-For example:
+  - (Optional) If you are going to use the internal load balancer, you can preconfigure it by passing `| sudo bash -s ha ekco-enable-internal-load-balancer`.
```
ajp-io, line 20: Rereading it, I question this line since `| sudo bash` will already be there, and you really should just pass `-s ha ekco-enable-internal-load-balancer`, assuming they know how to do that. But that's not a dealbreaker, so I'm approving. Can handle separately if we want to.

ajp-io (approved):

docs/enterprise/installing-embedded-cluster.md
ajp-io: Just to clarify this, because I don't know if it's clear as I reread it: the default command includes the `| sudo bash` part, so it might make more sense to tell people to just pass `-s ha ekco-enable-internal-load-balancer`, since that's the only net new thing they need to add. But like I said before, I can live with it either way if we want to just merge this.

jonquil2002: Typically, we are supposed to document the entire command line entry, even if something is already there. So something like this would be correct formatting (even though this is not a full command, just showing this for demo purposes):

`$ curl install`

Not sure if this helps?

ajp-io: If we're supposed to show the whole entry, then should we be showing all the way back to `curl` instead of starting in the middle with `| sudo bash`?

jonquil2002: Yep, I think so. I could open another PR to add the missing parts. Do you want to supply the full command here, s I'm not sure what it should be?

PR #386, opened 2022-06-03
Remove code block syntax in command

red-patience:
This PR should fix an issue where the triple backtick syntax to create code blocks is displayed in the command for creating a support bundle. Using the copy button currently copies these backticks along with the command.


jonquil2002 (approved):
Nice! Thank you so much!

PR #389, opened 2022-06-07
48427: Removing "Understanding Releases" because it's moot

jonquil2002:
SC: https://app.shortcut.com/replicated/story/49427/fix-move-content-for-semantic-versioning-and-understanding-releases


jonquil2002:
@paigecalvert  I need to set up the redicte request in the morning. So no rush, but please review this when you can. I only needed to remove the topic "Understanding Releases" and didn't have to copy the content to "Enabling Semantic Versioning" because the content was redundant.

paigecalvert (approved):

PR #403, opened 2022-06-16
Update admin-console-display-app-status.md

JustinN-Replicated:
adding important note for status informers

### DIFF
--- a/docs/vendor/admin-console-display-app-status.md
+++ b/docs/vendor/admin-console-display-app-status.md
@@ -18,7 +18,7 @@ You can target resources of the supported types that are deployed in any of the 
 
 * Deployed directly by the Replicated app manager.
 * Deployed by a Kuberentes Operator that is deployed by the app manager. For more information, see [About Packaging a Kubernetes Operator Application](operator-packaging-about).
-* Deployed by Helm v3 using native Helm. For more information, see [Helm Overview](helm-overview).
+* Deployed by Helm. For more information, see [Helm Overview](helm-overview).
 
 Replicated recommends that you add at least one resource.
 
--- a/docs/vendor/admin-console-display-app-status.md
+++ b/docs/vendor/admin-console-display-app-status.md
@@ -4,15 +4,25 @@ You can configure the Application custom resource to display application status 
 
 ![Application Status](/images/kotsadm-dashboard-appstatus.png)
 
-It is necessary to target specific Kubernetes resources for the dashboard to accurately report status.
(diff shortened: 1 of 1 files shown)


JustinN-Replicated:
I had also considered the following text in the note; not sure which reads better:

> Status informers are not limited to resources created by app manager; any supported Kubernetes resource may be used to help reflect the applications current status.

This came up in discussions from [this internal slack thread](https://replicated.slack.com/archives/C131GHQ83/p1655393665097069?thread_ts=1655337749.013779&cid=C131GHQ83)

paigecalvert:
@JustinN-Replicated thank you! this is a great addition. 

I pulled this info out of the note and into a list of the different ways these supported resources can de deployed in order to use them as status informers.

I also couldn't resist a little extra editing to this topic while reviewing your changes.

Not going to merge because I want you to have a change to review my changes and make sure it's accurate

paigecalvert (dismissed):

docs/vendor/admin-console-display-app-status.md
```diff
@@ -4,11 +4,25 @@ You can configure the Application custom resource to display application status
...
+* StatefulSet
+* Service
+* Ingress
+* PersistentVolumeClaims
+
+You can target resources of the supported types that are deployed in any of the following ways:
+
+* Deployed directly by the Replicated app manager.
+* Deployed by a Kuberentes Operator that is deployed by the app manager. For more information, see [About Packaging a Kubernetes Operator Application](operator-packaging-about).
+* Deployed by Helm v3 using native Helm. For more information, see [Helm Overview](helm-overview).
```
JustinN-Replicated, line 21: @paigecalvert I wasn't sure why we are mentioning the helm version and native Helm in particular; it may be challenging for users to parse here that Native helm uses help to deploy vs replicated helm deploys the yaml directly w/o helm. 

maybe we could just simplify like this?
* Deployed by Helm. For more information, see [Helm Overview](helm-overview).

paigecalvert: Sounds good! I thought to include this detail because if you are not using _native_ Helm, then the resources are deployed directly by kots (right?), which would put it under the category of bullet #1. But I think the idea to just say "deployed by helm" totally works

JustinN-Replicated:
@pavansokkenagaraj could I have you review real quick? Just want to make sure things are accurate.

pavansokkenagaraj (approved):

PR #409, opened 2022-06-17
50882 add list of distribution return values

paigecalvert:
https://app.shortcut.com/replicated/story/50882/update-the-static-context-docs-with-full-list-of-distros

Preview: https://deploy-preview-409--replicated-docs.netlify.app/reference/template-functions-static-context#distribution


cremerfc (approved):
Looks great! Thank you very much!

PR #412, opened 2022-06-21
all lowercase values for distributions

cbodonnell:
fixing a few values for distribution that should be all lowercase


jonquil2002 (approved):
lgtm

PR #415, opened 2022-06-22
38776: Benefits of including Velero add-on with Kubernetes installer

jonquil2002:
SC: https://app.shortcut.com/replicated/story/38776/snapshots-add-on-for-embedded-cluster-can-be-included-in-kubenetes-installer-spec

Previews:

- https://deploy-preview-415--replicated-docs.netlify.app/vendor/packaging-embedded-kubernetes
- https://deploy-preview-415--replicated-docs.netlify.app/vendor/snapshots-overview

### DIFF
--- a/docs/vendor/snapshots-configuring-backups.md
+++ b/docs/vendor/snapshots-configuring-backups.md
@@ -63,7 +63,7 @@ The snapshots feature is a backup and restore option that lets you define a mani
 
     ```
 
-1. Optional: Configure manifest exclusions. By default, Velero also includes snapshots of all of the Kubernetes objects in the namespace.
+1. (Optional) Configure manifest exclusions. By default, Velero also includes snapshots of all of the Kubernetes objects in the namespace.
 To exclude any manifest, add a `velero.io/exclude-from-backup` label to the manifest to be excluded.
 
     Example:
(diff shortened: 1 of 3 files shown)


JustinN-Replicated (approved):
lgtm

docs/vendor/snapshots-overview.md
```diff
@@ -5,16 +5,24 @@ Snapshots is the backup and restore feature for applications. This is an optiona
...
-**Note**: The restore process is managed through the kots CLI only.
+:::note
+The restore process is managed through the kots CLI only.
+:::
 
 The app manager also exposes hooks that can be used to inject scripts to execute with Snapshots both [before and after a backup](snapshots-configuring-backups) and [before and after a restore](../enterprise/snapshots-understanding).
 
+## About the Velero Add-On for Kubernetes Installers
+
+If you are distributing your application with the Kubernetes installer, we recommend that you include the Velero add-on so that customers do not have to manually install Velero on their cluster. For more information about the Kubernetes installer, see [Creating a Kubernetes Installer Specification](packaging-embedded-kubernetes).
```
paigecalvert, line 16: A couple notes on this section, for your consideration:

- Because this is an important step for a vendor to consider when configuring snapshots, what do you think about  instead including the info as an optional step in the Configuring procedure? https://deploy-preview-415--replicated-docs.netlify.app/vendor/snapshots-configuring-backups

  As is, this content appearing in a conceptual topic under an "About" heading might indicate that there is no action to be taken. Making this a step in the procedure might better convey the idea that "adding the velero add-on is something you do as part of properly configuring snapshots"
  
  (Having it be a step in the procedure could help avoid adding a specific heading for the info, too)

- One piece of conceptual info that's def missing from this topic is that not only do we use velero, but it has to be installed on the cluster for snapshots to work. Could consider just adding a sentence to the existing mention of velero on this topic to call this out? (main suggestion here is just that we explain the connection that velero has to be installed on the cluster for snapshots to work)

paigecalvert (commented):
Thank you for adding this info @jonquil2002 ! Added a couple suggestions

docs/vendor/snapshots-configuring-backups.md
```diff
@@ -79,6 +79,8 @@ To exclude any manifest, add a `velero.io/exclude-from-backup` label to the mani
       uri: Secret To Not Include
 
     ```
+1. Optional: If you are distributing your application with the Kubernetes installer, we recommend that you include the Velero add-on so that customers do not have to manually install Velero on their cluster. For more information about the Kubernetes installer, see [Creating a Kubernetes Installer Specification](packaging-embedded-kubernetes).
```
paigecalvert, line 82: I've been using the (Optional) format--have a preference?

jonquil2002: I'm good with (Optional), I actually prefer that. I think this was an RH carry-over.

paigecalvert (approved):

PR #422, opened 2022-06-24
50499: document the releaseName field for the HelmChart CRD

sgalsaleh:


docs/reference/custom-resource-helmchart.md
```diff
@@ -81,11 +81,11 @@ This must match the `version` field from a `Chart.yaml` in a `.tgz` chart archiv
 
 > Introduced in Replicated app manager 1.73.0
 
-The release name to be used when installing this instance of the Helm chart.
+Specifies the release name to be used when installing this instance of the Helm chart.
 Defaults to the chart name.
 The release name must be unique across all charts deployed in the namespace.
-Specifying a release name allows you to deploy multiple instances of the same Helm chart. Otherwise, the chart name will be used for each instance, causing a collision.
-Must be a valid Helm release name (must match regex `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` and be no longer than 53).
+Specifying a release name allows you to deploy multiple instances of the same Helm chart. If a unique release name is not used, then the chart name is used for each instance instead, causing a collision.
```
sgalsaleh, line 87: this implies that if a release name is used but it's not unique, the chart name will be used, which is not accurate. the specified release name will always be used even if it's not unique.

jonquil2002: So would it be more accurate to say "If a release name is not used, then the chart name is used for each instance instead, causing a collision."?

sgalsaleh: the release name is specified for each instance separately. if the release name for a particular instance isn’t used, then the chart name will be used only for that instance.

i think the point we’re trying to make here is:
when deploying multiple instances of the same helm chart, a unique release name must be used for each instance.

how about
"Specifying a unique release name allows you to deploy multiple instances of the same Helm char"

jonquil2002 (dismissed):
lgtm, if you think my latest commit is ok to pull in.

docs/reference/custom-resource-helmchart.md
```diff
@@ -77,6 +78,15 @@ This must match the `name` field from a `Chart.yaml` in a `.tgz` chart archive t
 The version of the chart.
 This must match the `version` field from a `Chart.yaml` in a `.tgz` chart archive that's also included in the release.
 
+### `chart.releaseName`
+
+> Introduced in Replicated app manager 1.73.0
+
+Specifies the release name to be used when installing this instance of the Helm chart.
+Defaults to the chart name.
+The release name must be unique across all charts deployed in the namespace if you want to deploy multiple instances of the same Helm chart.
```
sgalsaleh, line 87: it has to be unique across all charts, not just when deploying multiple instances of the same chart

jonquil2002 (approved):
lgtm

PR #425, opened 2022-06-25
Tags & digests can't be used together currently

grantmiller:


paigecalvert (approved):
thanks @grantmiller !

PR #430, opened 2022-06-28
Update create a PR and issue Github icons

miaawong:
Screen Shot 2022-06-29 at 9 31 54 AM
Screen Shot 2022-06-29 at 9 31 50 AM

In hover state: 
Screen Shot 2022-06-29 at 9 31 58 AM


paigecalvert (commented):
Definitely like the change on the feedback icon 👍 

For the other text, I think we should keep it as the pencil icon for now. I say this because Mark just requested here that I change the text to say "Suggest a change", which deprioritizes the fact that it's a PR. It might be more easily understood as a pencil icon in that case: https://replicated.slack.com/archives/C02BPJQQP28/p1656441996109449?thread_ts=1654613201.802639&cid=C02BPJQQP28. LMK if you or @breezybaldwin agrees on that, or if you think it would still make more sense as the PR icon

PR #431, opened 2022-06-28
48569: Clarify Single Line String is Required

jonquil2002:
SC: https://app.shortcut.com/replicated/story/48569/conditional-resource-declaration-multiline-configuration-is-inconsistent

### DIFF
--- a/docs/vendor/helm-optional-charts.md
+++ b/docs/vendor/helm-optional-charts.md
@@ -72,11 +72,13 @@ stringData:
 ```
 
 Let's edit this to add a conditional statement, rendering either a connection string to the embedded postgres chart, or the user supplied instance, as needed.
-The logic for this can be thought of as:
+The logic for this can be thought of as follows:
+
+:::important
+This multiple line string must be a single line in your application, but it is displayed here on multiple lines for readability. Continue reading to see how this is combined into a single line.
+:::
 
 ```shell
-## NOTE: This must be all on one line in your application, but it's displayed here on multiple lines for readability.
-## Continue reading to see how this is combined into a single line
 repl{{ if ConfigOptionEquals "postgres_type" "embedded_postgres" }}
   postgres://myapplication:repl{{ ConfigOption "embedded_postgres_password" }}@postgres:5432/mydatabase?sslmode=disable
 repl{{ else }}
--- a/docs/vendor/tutorial-adding-db-config.md
+++ b/docs/vendor/tutorial-adding-db-config.md
@@ -556,7 +556,11 @@ psql: could not translate host name "postgres" to address: Name or service not k
 ### Mapping User Inputs
 
 To map the user-supplied configuration, we'll start by expanding our secret we created before, adding fields for additional variables, using `{{repl if ... }}` blocks to switch between embedded/external contexts.
-To start, we'll add a field for hostname, using the yaml `>-` to collapse the multiline string into a single line:
+To start, we'll add a field for hostname:
+
+:::important
+This DB_HOST multiple line string must be a single line in your application, but it is displayed here on multiple lines for readability.
+:::
 
 ```yaml
 apiVersion: v1
@@ -565,7 +569,7 @@ metadata:
   name: postgres
 data:
   DB_PASSWORD: '{{repl ConfigOption "embedded_postgres_password" | Base64Encode }}'
-  DB_HOST: >-
+  DB_HOST:
     {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" }}
       {{repl Base64Encode "postgres" }}
     {{repl else}}


docs/vendor/tutorial-adding-db-config.md
```diff
@@ -565,14 +565,18 @@ metadata:
...
+  DB_HOST:
     {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" }}
       {{repl Base64Encode "postgres" }}
     {{repl else}}
       {{repl ConfigOption "external_postgres_host" | Base64Encode }}
     {{repl end}}
 ```
 
+:::important
+The DB_HOST multiple line string above must be a single line in your application, but it is displayed here on multiple lines for readability.
```
ajp-io, line 577: IMO the way this is handled on the other page makes more sense, where we say "think of it like this" but then show it all on one line in the example. If it's supposed to be written on one line, I think our example should show it that way. If we want to show it separated out for readability, I prefer the approach above where we explicitly do it twice so that the readable version and the actual example can be seen.

docs/vendor/helm-optional-charts.md
```diff
@@ -72,19 +72,20 @@ stringData:
...
 repl{{ else }}
   repl{{ ConfigOption "external_postgres_uri" }}
 repl{{ end }}
 ```
+:::important
+The multiple line string above must be a single line in your application, but it is displayed here on multiple lines for readability.
+:::
 
-But we need to write it all on a single line.
+However, we need to write it all on a single line.
```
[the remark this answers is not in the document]
jonquil2002: I know, it flows terribly, but then it did to begin with. I wasn't sure what I could cut since I don;t know this part of the technology well. Let's see what you think about my rewrite.

docs/vendor/tutorial-adding-db-config.md
jonquil2002: Ok, tried to mirror it, though the two are different - one uses the Bsse64Encode, one does not. Let's see if this rewrite helps.

jonquil2002: Changed it, I trust your perspective. Fingers crossed.

docs/vendor/helm-optional-charts.md
```diff
@@ -72,21 +72,20 @@ stringData:
 ```
 
 Let's edit this to add a conditional statement, rendering either a connection string to the embedded postgres chart, or the user supplied instance, as needed.
-The logic for this can be thought of as:
+The logic for this can be thought of as follows:
```
paigecalvert, line 75: This stem sentence feels off, given that's it's leading into a YAML sample and not really explaining logic. (I know it was there before, but just as part of making this multi-line configuration documentation more clear to help avoid vendors running into this problem again)

docs/vendor/tutorial-adding-db-config.md
```diff
@@ -565,14 +567,28 @@ metadata:
...
+The following example shows the same YAML file with the `DB_HOST` field as multiple lines for better readability. However, it must be a single line in your application, as shown above.
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: postgres
+data:
+  DB_PASSWORD: '{{repl ConfigOption "embedded_postgres_password" | Base64Encode }}'
+  DB_HOST:
```
paigecalvert, line 583: Since the goal here is just to show the DB_HOST template functions as multiple lines for readability, suggestion to remove the rest of the YAML from this one (lines 577 - 583).

That could cut down on the noise, and help avoid anyone copying this whole thing and using it as is. 

This might be something like...

The following displays the template functions in the `DB_HOST` key as multiple lines for better readability. These template functions must be a single line in your application, as shown in the example above.

```
{{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" }}
      {{repl Base64Encode "postgres" }}
{{repl else}}
      {{repl ConfigOption "external_postgres_host" | Base64Encode }}
{{repl end}}
```

docs/vendor/helm-optional-charts.md
```diff
@@ -72,21 +72,20 @@ stringData:
...
 repl{{ end }}
 ```
+:::important
+The multiple line string above must be a single line in your application, but it is displayed here on multiple lines for readability.
+:::
 
-But we need to write it all on a single line.
-For readability, we are using the `stringData` field of the Kubernetes Secret object, which allows us to not base64 encode the value.
-Replicated has a Base64Encode function available that you can pipe a string through, if desired.
+The following example shows the same YAML file with the `stringData` field of the Kubernetes Secret object as a single line. Optionally, you can use the Replicated Base64Encode function to pipe a string through.
```
paigecalvert, line 88: +1 to Alex's comment that I think it makes sense to swap these and show the example that we really want them to use first and foremost (like how you've done in the second file). Then the step of showing "multiple lines for readability" can be an added bonus, if we feel it's necessary.

Can probably borrow the same language/structure that yo've used in the other topic for this. Making this change also helps to avoid the need for the attention-drawing Important note

paigecalvert (approved):
@jonquil2002 left a few suggestions!

paigecalvert (approved):
lgtm!

PR #439, opened 2022-07-05
add missing paren

dexhorthy:
do we want a period at the end of that sentence? Unsure where it should go so I removed it 🙂 will defer to y'all on that part


paigecalvert:
@dexhorthy thank you!! Decided to go the other way and remove the parens all together and add the period. Either way, the stray paren had got to go

paigecalvert (approved):

PR #440, opened 2022-07-05
Fix badly formatted note

ajp-io:


jonquil2002 (approved):
lgtm, thanks for catching that.

PR #449, opened 2022-07-08
Change installer feature to beta and add latest and .x requirement

ajp-io:


ajp-io:
The reason I didn't is because it's recommended in the other case (hence the note) but actually required in this case. You cannot use anything but specific versions with this new feature. The old way lets you use other things, we just don't recommend it. If there's a better way to present it, let me know.

PR #471, opened 2022-07-19
add version for required releases

cbodonnell:
Adding a note to the required releases section to indicate the KOTS version that this was released in.  I did not see an area where this was explicitly specified outside of the release notes.

### DIFF
--- a/docs/vendor/releases-promoting.md
+++ b/docs/vendor/releases-promoting.md
@@ -28,7 +28,11 @@ To promote a release:
 1. Edit the fields:
     * **Channel**: Select the channel where you want to promote the release. The defaults are Stable, Beta, and Unstable. If you created custom channels using the CLI, they are listed here also.
     * **Version label**: Enter a version label. If semantic versioning is enabled for the channel, you must use a valid semantic version. For more information, see [Enabling Semantic Versioning](releases-semantic-versioning).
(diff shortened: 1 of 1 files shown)


ajp-io:
Thanks for noticing this, Craig.

More of a comment to docs when they review, but we might want to consider rewording the note a bit. This section is about the vendor portal and marking a release as required in the vendor portal. It's not about the app manager. So mentioning app manager support without more details might be confusing.

The point Craig is trying to make is that even if you mark a release as required, only app manager 1.68+ will respect the required release. Prior to that, you can skip required releases since the support wasn't yet added.

So we might want the note to say something a bit more descriptive, like "Required releases are only respected in app manager v1.68.0+." Not necessarily perfect wording, but that's the gist of what I think we want to say here.

cbodonnell:
> So we might want the note to say something a bit more descriptive, like "Required releases are only respected in app manager v1.68.0+." Not necessarily perfect wording, but that's the gist of what I think we want to say here.

I updated the wording to "Required releases are only supported in app manager v1.68.0+."

Marking this one as ready for review now.

ajp-io (dismissed):

paigecalvert (dismissed):
Good call @cbodonnell , thanks for adding this.

I pushed one edit to move the info out of a note and into the text itself so this bullet point now begins with: "In the app manager v1.68.0 and later, ...". This just helps to avoid "note overload" for readers :)

paigecalvert:
Ah wait, reviewing your convo with Alex above, I'm realizing that this change I made might not be the most accurate. Going to re-review.

paigecalvert (approved):
@cbodonnell OK, put it back in the note with just minor text changes. :)

PR #478, opened 2022-07-21
52621: Changes to Preflight & Support Bundle Custom Resource Reference Content

jonquil2002:
SC: https://app.shortcut.com/replicated/story/52621/better-reference-content-on-the-preflights-features

Preview: https://deploy-preview-478--replicated-docs.netlify.app/reference/custom-resource-preflight


docs/reference/custom-resource-preflight.md
```diff
@@ -22,4 +22,59 @@ spec:
...
+
+The following fields are supported on all optional collectors for preflights and support bundles:
+
+<table>
+  <tr>
+    <th width="30%">Field Name</th>
+    <th width="70%">Description</th>
+  </tr>
+  <tr>
+    <td>`collectorName`</td>
```
paigecalvert, line 47: Since this is in an HTML table, we have to use `<code></code>` instead of the backticks to get this to display properly as code

docs/reference/custom-resource-preflight.md
```diff
@@ -22,4 +22,59 @@ spec:
...
+    <td>(Optional) A collector can specify the `collectorName` field. In some collectors, this field controls the path where result files are stored in the support bundle.</td>
+  </tr>
+  <tr>
+    <td>`exclude`</td>
+    <td>(Optional) Based on the runtime available configuration, a conditional can be specified in the `exclude` field. This is useful for deployment techniques that allow templating for Replicated app manager and the optional Helm component. When this value is `false`, the collector is not included.</td>
+  </tr>
+</table>
+
+
+## Analyzer Global Fields
```
paigecalvert, line 57: Would it make sense to include in these sections some sample YAML that shows the format of adding these global fields? I found myself wondering how the information in these tables would be used as is. 

I think it does make sense to stick to the global fields here, but I'm not sure it's useful without a bit more context

docs/reference/custom-resource-preflight.md
```diff
@@ -22,4 +22,59 @@ spec:
...
+    <td>`exclude`</td>
+    <td>(Optional) Based on the runtime available configuration, a conditional can be specified in the `exclude` field. This is useful for deployment techniques that allow templating for Replicated app manager and the optional Helm component. When this value is `true`, the analyzer is not included.</td>
+  </tr>
+  <tr>
+    <td>`strict`</td>
+    <td>(Optional) An analyzer can be set to `strict: true` so that `fail` outcomes for that analyzer prevent the release from deploying until the vendor-specified requirements are met. When `exclude: true` is also specified, `exclude` overrides `strict` and the analyzer is not executed.</td>
+  </tr>
+</table>
+
+For a list of collectors and analyzers, see [All Collectors](https://troubleshoot.sh/docs/collect/all/) and [Analyzing Data](https://troubleshoot.sh/docs/analyze/) in the Troubleshoot documentation.
```
paigecalvert, line 80: As I was trying to understand the global fields and why/how I'd use them, I was looking for a links for more information and I almost missed this. Maybe this could moved more to the top of these sections (I could see for example an H2 Global Fields heading with a quick intro and links for more info that these are nested under as H3s)

docs/reference/custom-resource-preflight.md
```diff
@@ -1,16 +1,16 @@
...
+Preflight checks use collectors and analyzers to provide cluster operators with clear feedback for any missing requirements or incompatibilities in the target environment before an application is deployed.
 
-For more information, see [Introduction](https://troubleshoot.sh/docs/preflight/introduction/) in the Troubleshoot documentation.
+Support bundles collect, redact, and analyze troubleshooting data from a cluster and help diagnose problems with application deployments.
 
-This enables the app manager and the Replicated admin console to provide cluster Operators with clear feedback on any missing requirements or incompatibilities in the target environment before an application is deployed.
+Default preflight checks and support bundle features are automatically included with releases. To customize these features uniquely for your application release, add any of the optional collectors and analyzers to `kind: Preflight` and `kind: SupportBundle` custom resource manifest files. For more information about these troubleshoot features and how to customize them, see [Customizing Preflight Checks and Support Bundles](/vendor/preflight-support-bundle-creating/).
```
paigecalvert, line 7: I think this information is really valuable for understanding the "Basic Manifest" section below. Might it make sense to move it under that heading? I think the context about how these are included by default is particularly useful, as well as the link for more info.

As is, I don't feel like I am able to take action after reading the Basic Manifest section because there is no direction as to what I do next (a xref to the global fields section below could also be helpful in this)

paigecalvert (approved):
Much improved!!!

I added a few comments, the summary being that I think there could be a bit more context/direction of how to format the fields to make this reference information more useful. As is, I wasn't sure how to include the global fields provided.

I know there is a balance between reference content and procedural/conceptual content, so we might need to be careful about crossing that line too much, but I think just a YAML sample to show the format of adding fields, as well as links to other fields in the troubleshooting docs, would help a lot!

docs/reference/custom-resource-preflight.md
jonquil2002: Moved it, let's see what we think.

docs/reference/custom-resource-preflight.md
jonquil2002: Moved it, let's see what we think of the flow.

docs/reference/custom-resource-preflight.md
jonquil2002: Yep

jonquil2002:
@paigecalvert  Great suggestions. Please take a look at these adjustments and see if this is better.

docs/reference/custom-resource-preflight.md
```diff
@@ -22,4 +30,110 @@ spec:
...
+    <td><code>collectorName</code></td>
+    <td>(Optional) A collector can specify the <code>collectorName</code> field. In some collectors, this field controls the path where result files are stored in the support bundle.</td>
+  </tr>
+  <tr>
+    <td><code>exclude</code></td>
+    <td>(Optional) Based on the runtime available configuration, a conditional can be specified in the <code>exclude</code> field. This is useful for deployment techniques that allow templating for Replicated app manager and the optional Helm component. When this value is <code>false</code>, the collector is not included.</td>
+  </tr>
+</table>
+
+**Example: Collector Definition for a Support Bundle**
```
jonquil2002, line 72: @ajp-io I'm wondering how/if we can add an example of the `exclude` field to this example. The Troubleshoot site mentions the use of templating, and I'm not sure how that all works together or whether there are simpler examples and use cases of `exclude` that we could use instead.

docs/reference/custom-resource-preflight.md
```diff
@@ -1,16 +1,24 @@
...
 
-For more information, see [Introduction](https://troubleshoot.sh/docs/preflight/introduction/) in the Troubleshoot documentation.
+Support bundles collect and analyze troubleshooting data from a cluster and help diagnose problems with application deployments. Default support bundles are automatically included with releases, and can be customized.
 
-This enables the app manager and the Replicated admin console to provide cluster Operators with clear feedback on any missing requirements or incompatibilities in the target environment before an application is deployed.
+Collectors and analyzers are configured in the Preflight and Support Bundle custom resource manifest files.
 
-The Preflight custom resource is optional.
+:::note
+Built-in redactors run by default for preflight checks and support bundles to protect customers' sensitive information. To add custom redactors, use the Redactor custom manifest file. For more information about the Redactor custom manifest file, see [Redactor](custom-resource-redactor).
```
jonquil2002, line 10: @ajp-io Is it correct that if a vendor defines a preflight check that our built-in redactors will run automatically? Or does that only apply to support bundles. Please lmk if a Zoom meeting would be easier.

ajp-io: Looks like you got some answers to this in Slack. Let me know if you need anything else.

jonquil2002: Yep, I just clarified that they'd customize the redactors for the support bundle.

docs/reference/custom-resource-preflight.md
jonquil2002: @sgalsaleh  Would you please take a look at this and see whether it's worth adding an example of an exclude field to this example - and how to do it? I only saw references to templating as a use case but no examples of how that all fits together. Or perhaps the templating example is not worth adding here and requires something more expansive on the Troubleshoot site instead?

jonquil2002:
@sgalsaleh  Would you please review this for technical accuracy and also see my question above?

docs/reference/custom-resource-preflight.md
```diff
@@ -1,16 +1,24 @@
...
+## Basic Manifest Files
 
-A Preflight custom resource manifest file is a collector and an analyzer specification in one manifest. The following is an example manifest file for the Preflight custom resource:
+To define preflight checks or customize the default support bundle settings, add the corresponding custom resource manifest file to your release. Then add custom collector and analyzer specifications to the custom resource manifest file. For more information about these troubleshoot features and how to configure them, see [Configuring Preflight Checks and Support Bundles](/vendor/preflight-support-bundle-creating/).
+
+The following sections show basic manifest files for the Preflight and Support Bundle custom resources that you can start with.
+
+### Preflight
+
+The Preflight manifest file uses `kind: Preflight`. All of the default collectors and analyzers are included, indicated by `[]`.
```
sgalsaleh, line 21: there are no default analyzers for the preflight manifest

docs/reference/custom-resource-preflight.md
ajp-io: Yeah, giving an example of `exclude` would be great. A good example is excluding a preflight when the cluster is or is not a kURL cluster. [Here's](https://replicated.slack.com/archives/CKXRFP7GA/p1624932639164900?thread_ts=1624931849.161300&cid=CKXRFP7GA) an example from Swimlane:
```
- textAnalyze:
        exclude: 'repl{{ (not IsKurl) }}'
        checkName: Weave Version
        fileName: versions/weave.txt
        # The KOTS data collectors erroneously inject some whitespace so the (?m) at the start is necessary until they fix that
        # This regex checks if the patch version is greater than or equal to 2.6.5.
        regex: '(?m)^(2)\.(6)\.([5-9]+|\d{2,})$'
        outcomes:
          - fail:
              message: Weave version ({{repl KurlString "Weave.Version"}}) is incompatible with this release. Please see the release notes about upgrading.
          - pass:
              message: Weave version ({{repl KurlString "Weave.Version"}}) is compatible with this release.
```
Probably don't want to copy that exactly, but you can see how the `exclude` field words there to detect whether the cluster is from kURL or not. This excludes the preflight is the cluster is _not_ kURL.

jonquil2002: Thanks, @ajp-io! Let's see what you think about the updated example with `exclude` added in.

ajp-io: Yeah, I think that can work!

jonquil2002:
@ajp-io  Please approve if you can. Thanks.

paigecalvert (approved):

PR #480, opened 2022-07-21
Add procedure for delivering admin console

paigecalvert:
https://app.shortcut.com/replicated/story/52694/document-how-to-deliver-the-admin-console-when-offering-direct-helm-install

I based this new section and procedure off these instructions that Marc wrote: https://docs.google.com/document/d/1P4oCUtkM8IileNNw83NepEsyq14rTTS9w-64T-uwmeA/edit#heading=h.lfu2nj4y975g

Preview: [/vendor/helm-install#deliver-admin-console](https://deploy-preview-480--replicated-docs.netlify.app/vendor/helm-install#deliver-admin-console)

### DIFF
--- a/docs/vendor/helm-install.md
+++ b/docs/vendor/helm-install.md
@@ -60,7 +60,7 @@ This procedure shows how to add the `admin-console` Helm chart as a conditional 
 
 It also shows how to conditionally exclude the `admin-console` Helm chart when your customers install with either the kots CLI or with the Kubernetes installer. It is important to exclude the `admin-console` chart in this scenario because both the kots CLI and the Kubernetes installer already install the admin console by default.
 
-To conditionally include or exclude the `admin-console` Helm chart:
+To conditionally include and exclude the `admin-console` Helm chart:
 
 1.  In the Helm chart `Chart.yaml` file, add the `admin-console` Helm chart to the `dependencies` field:
 
@@ -121,7 +121,7 @@ To conditionally include or exclude the `admin-console` Helm chart:
 
    The Replicated HelmChart custom resource allows Replicated to process and deploy Helm charts when customers install with the kots CLI or with the Kubernetes installer.
 
-   For more information, see [HelmChart](reference/custom-resource-helmchart) in the _References_ section.
+   For more information, see [HelmChart](/reference/custom-resource-helmchart) in the _References_ section.
 
 1. Edit the HelmChart manifest file to exclude the `admin-console` Helm chart when customers install your application with the kots CLI or with the Kubernetes installer:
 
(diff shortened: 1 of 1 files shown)


docs/vendor/helm-install.md
```diff
@@ -35,24 +35,171 @@ The `helm install` method is Alpha and has the following limitations:
...
+        chart:
+          name: samplechart
+          chartVersion: 3.1.7
+          releaseName: samplechart-release-v1
+        values:
+          admin-console:
+           enabled: false  
+      ```     
+
+   1. If the HelmChart custom resource includes an `optionalValues` field, set `recursiveMerge` to `true`. This prevents the `admin-console` field in `values` from being overwritten by the fields in `optionalValues`.
```
paigecalvert, line 155: I added this step, so need to verify that it is technically accurate. I believe if recursiveMerge is not set to true (or if the vendor doesn't copy and paste the `admin-console` field to their opitonalValues section as well), then this could be overwritten by the optionalValues section

docs/vendor/helm-install.md
```diff
@@ -35,24 +35,171 @@ The `helm install` method is Alpha and has the following limitations:
...
+1. In the `values.yaml` file, add the following fields that map to the `condition` field in the `Chart.yaml` file:
+
+   ```yaml
+   admin-console:
+    enabled: true
+   ```
+
+1. Package your Helm chart, and add the packaged chart to a release in the Replicated vendor portal. For more information, see [Add a Helm Chart to a Release](helm-release#add-a-helm-chart-to-a-release).
+
+1. In the release, create or open the HelmChart custom resource manifest file. A HelmChart custom resource manifest file has `kind: HelmChart`.
```
ajp-io, line 109: The API version and kind are the unique identifier, so I'd probably add the kot.io/v1beta1 API version here too. Likely there isn't another HelmChart custom resource that they include, but this would be the true unique identifier.

ajp-io (commented):
I'd add a sentence or two somewhere to say that the admin console works differently in this mode than it normally does, since Helm, and not the admin console, manages the lifecycle of the application. So, for example, it provides the correct `helm upgrade` commands rather than a button you can click to upgrade the application from the admin console.

paigecalvert:
> I'd add a sentence or two somewhere to say that the admin console works differently in this mode than it normally does, since Helm, and not the admin console, manages the lifecycle of the application. So, for example, it provides the correct `helm upgrade` commands rather than a button you can click to upgrade the application from the admin console.

Oh good call. I added this to the Overview with the upgrade example that you provided. (We can add to this with screenshots and such when available).

ajp-io (approved):

ajp-io (approved):

PR #485, opened 2022-07-26
Add procedure for using private registry

paigecalvert:
Story: https://app.shortcut.com/replicated/story/54157/add-procedure-for-using-private-images-to-alpha-topic

Preview: https://deploy-preview-485--replicated-docs.netlify.app/vendor/helm-install#private-images

I based this procedure on the [Example: Delivering Image Pull Secrets for Private Images](https://docs.replicated.com/vendor/helm-install#example-delivering-image-pull-secrets-for-private-images) content that is in the currently-published version of this topic. I also used the generic procedure provided in the [Helm Installation (Alpha)](https://docs.google.com/document/d/1P4oCUtkM8IileNNw83NepEsyq14rTTS9w-64T-uwmeA/edit#heading=h.l11cvyaz20lx) google doc.

### DIFF
--- a/docs/vendor/helm-install.md
+++ b/docs/vendor/helm-install.md
@@ -18,7 +18,9 @@ When a new version of a chart is uploaded, Replicated renders and cache that ver
 
 When a license field changes for a customer, Replicated invalidates all rendered and cached charts. This causes the charts to be rebuilt the next time they are pulled.
 
-The `values.yaml` in your Chart is rendered with the customer provided license. This creates a way to pass custom license fields and other customer-specific data into the `values.yaml` as defaults, and consume them in Helm templates. When Replicated renders the `values.yaml`, it is not assumed to be a valid YAML format in order to allow flow-control and conditional template functions to optionally write some fields to the `values.yaml` file.
+The `values.yaml` in your Chart is rendered with the customer provided license. Each customer logs in with unique credentials, and our registry is able to identify which customer is pulling the chart. This creates a way to pass custom license fields and other customer-specific data into the `values.yaml` as defaults, and consume them in Helm templates.
+
+When Replicated renders the `values.yaml`, it is not assumed to be a valid YAML format in order to allow flow-control and conditional template functions to optionally write some fields to the `values.yaml` file. Replicated renders template functions in the `values.yaml` for each customer, as the Helm chart is served to the customer.
 
 ## Requirement
 
@@ -47,7 +49,7 @@ To deliver the admin console with your application when users install with the h
 
 When you include the `admin-console` Helm chart as a dependency, Replicated injects values into the Helm chart `values.yaml` file when the chart is pulled by your users. These values provide the license ID for the customer, which enables the admin console to authenticate with the image registry and check for updates.
 
-The following shows the formatting of the `replicated` and `license_id` fields that  Replicated adds to the `values.yaml` file:
+The following shows the formatting of the `replicated` and `license_id` fields that Replicated adds to the `values.yaml` file:
 
 ```
 replicated:
@@ -203,58 +205,100 @@ To conditionally include and exclude the `admin-console` Helm chart:
       ```
   1. Save and promote the release to a development environment to test your changes.      
 
+## Deliver Image Pull Secrets for a Private Registry {#private-images}
 
-## About Customer Values and Private Images
+Using a private image registry for your application requires that the customer has an image pull secret to authenticate to the registry. When users install with the kots CLI, Replicated automatically generates this image pull secret.
 
-Replicated renders template functions in the `values.yaml` for each customer, as the Helm chart is served to the customer. Each customer license logs in with unique credentials, and our registry is able to identify which customer is pulling the chart. The registry does not render the chart, but it replaces your `values.yaml` with a rendered version.
+For users who install with the helm CLI, Replicated cannot automatically inject the image pull secret into the Helm chart for your application.
 
-### Example: Delivering Image Pull Secrets for Private Images
+To support private images for helm CLI installations, create a `pullSecrets` field in the Helm chart `values.yaml` file where you can use a Replicated template function to pass the unique customer license. This allows you to then consume this customer-specific pull secret from the `values.yaml` file in the templates for your Helm chart.
 
-`values.yaml`
+To deliver customer-specific image pull secrets for a private registry:
 
-```
-images:
-  pullSecrets:
-    replicated:
-      dockerconfigjson: {{repl LicenseDockerCfg }}
-  myapp:
-    repository: registry.replicated.com/my-app/my-image
-    tag: 0.0.1
-    pullPolicy: IfNotPresent
-    pullSecret: replicated
-```
+1. Add the following to the `values.yaml` file for your Helm chart:
 
-`templates/imagepullsecret.yaml`
+   ```yaml
+   images:
+    pullSecrets:
+      replicated:
+        dockerconfigjson: "repl{{ LicenseDockerCfg }}"
+   ```
 
-```
-{{ if .Values.images.pullSecrets.replicated.dockerconfigjson }}
-apiVersion: v1
-kind: Secret
-metadata:
-  name: replicated
-type: kubernetes.io/dockerconfigjson
-data:
-  .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
-{{ end }}
-```
+   Replicated renders the `LicenseDockerCfg` template function on `helm install` or `helm pull`. This allows the customer's license-specific credentials to be injected into the `values.yaml`.
 
-`templates/deployment.yaml`
+1. In the `templates` directory of your Helm chart, create a  `imagepullsecret.yaml` file that writes the pull secret that you added to the `values.yaml` into a Secret on the cluster:
 
-```
-        ...
-        image: {{ .Values.images.myapp.repository }}{{ .Values.images.myapp.tag }}
-        imagePullPolicy: {{ .Values.images.myapp.pullPolicy }}
+   ```yaml
+   {{ if .Values.images.pullSecrets.replicated.dockerconfigjson }}
+   apiVersion: v1
+   kind: Secret
+   metadata:
+    name: replicated
+   type: kubernetes.io/dockerconfigjson
+   data:
+    .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
+   {{ end }}
+   ```
+
+1. Add the following to the Pod `spec`:
+
+   ```yaml
+   spec:
+    template:
+      spec:
         {{ if .Values.images.pullSecrets.replicated }}
         imagePullSecrets:
           - name: replicated
         {{ end }}
-        name: myapp
-        ports:
-        - containerPort: 3000
-          name: http
-```          
+   ```
 
-### Example: Delivering Custom License Fields
+### Example: Delivering Image Pull Secrets for Private Images
+
+   `values.yaml`
+
+   ```yaml
+   images:
+     pullSecrets:
+       replicated:
+         dockerconfigjson: {{repl LicenseDockerCfg }}
+     myapp:
+       repository: registry.replicated.com/my-app/my-image
+       tag: 0.0.1
+       pullPolicy: IfNotPresent
+       pullSecret: replicated
+   ```
+
+   `templates/imagepullsecret.yaml`
+
+   ```yaml
+   {{ if .Values.images.pullSecrets.replicated.dockerconfigjson }}
+   apiVersion: v1
+   kind: Secret
+   metadata:
+     name: replicated
+   type: kubernetes.io/dockerconfigjson
+   data:
+     .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
+   {{ end }}
+   ```
+
+   `templates/deployment.yaml`
+
+   ```yaml
+           ...
+           image: {{ .Values.images.myapp.repository }}{{ .Values.images.myapp.tag }}
+           imagePullPolicy: {{ .Values.images.myapp.pullPolicy }}
+           {{ if .Values.images.pullSecrets.replicated }}
+           imagePullSecrets:
+             - name: replicated
+           {{ end }}
+           name: myapp
+           ports:
+           - containerPort: 3000
+             name: http
+   ```          
+
+## Example: Delivering Custom License Fields
 
 `values.yaml`
 
--- a/docs/vendor/helm-install.md
+++ b/docs/vendor/helm-install.md
@@ -211,7 +211,7 @@ Using a private image registry for your application requires an image pull secre
 
 When users install with the kots CLI or the Kubernetes installer, Replicated automatically uses the customer license to create and inject an image pull secret that grants proxy access to a private registry. Instances of your application can then pull private images through the Replicated proxy service. For more information, see [About Connecting to an External Registry](packaging-private-images#about-connecting-to-an-external-registry).
 
-For installations with the helm CLI, Replicated cannot automatically inject an image pull secret into the Helm chart for your application. To support the use of private images for helm CLI installations, use the Replicated `LicenseDockerCfg` template function to render a value based on the unique customer license file. Write the value to a pull secret, then reference the pull secret in the `templates/deployment.yaml` file for the Helm chart.
+For installations with the helm CLI, Replicated cannot automatically inject an image pull secret into the Helm chart for your application. To support the use of private images for helm CLI installations, use the Replicated `LicenseDockerCfg` template function to render a value based on the unique customer license file. Write the value to a pull secret, then reference the pull secret in the necessary template files for the Helm chart.
 
 For more information about the `LicenseDockerCfg` template function, see [LicenseDockerCfg](/reference/template-functions-license-context#licensedockercfg) in _License Context_.
 
--- a/docs/vendor/helm-install.md
+++ b/docs/vendor/helm-install.md
@@ -207,17 +207,24 @@ To conditionally include and exclude the `admin-console` Helm chart:
 
 ## Deliver Image Pull Secrets for a Private Registry {#private-images}
 
-Using a private image registry for your application requires an image pull secret to access the registry. The unique license file for each customer can grant access to the private registry without exposing registry credentials to the customer.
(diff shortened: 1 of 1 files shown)


docs/vendor/helm-install.md
```diff
@@ -18,7 +18,9 @@ When a new version of a chart is uploaded, Replicated renders and cache that ver
...
 
-The `values.yaml` in your Chart is rendered with the customer provided license. This creates a way to pass custom license fields and other customer-specific data into the `values.yaml` as defaults, and consume them in Helm templates. When Replicated renders the `values.yaml`, it is not assumed to be a valid YAML format in order to allow flow-control and conditional template functions to optionally write some fields to the `values.yaml` file.
+The `values.yaml` in your Chart is rendered with the customer provided license. Each customer logs in with unique credentials, and our registry is able to identify which customer is pulling the chart. This creates a way to pass custom license fields and other customer-specific data into the `values.yaml` as defaults, and consume them in Helm templates.
+
+When Replicated renders the `values.yaml`, it is not assumed to be a valid YAML format in order to allow flow-control and conditional template functions to optionally write some fields to the `values.yaml` file. Replicated renders template functions in the `values.yaml` for each customer, as the Helm chart is served to the customer.
```
ajp-io, line 23: I need some clarity on a few parts of this.

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,73 @@ To conditionally include and exclude the `admin-console` Helm chart:
       ```
   1. Save and promote the release to a development environment to test your changes.      
 
+## Deliver Image Pull Secrets for a Private Registry {#private-images}
 
-## About Customer Values and Private Images
+Using a private image registry for your application requires an image pull secret to access the registry. The unique license file for each customer can grant access to the private registry without exposing registry credentials to the customer.
```
ajp-io, line 210: As an example, here's what my pull secret credentials look like when `LicenseDockerCfg` is rendered. So this works for both proxied images and images storied in our Replicated private registry. We should clarify throughout that this is for both types of images. We're talking only about proxied images in external registries here.
```
{"auths":{"proxy.replicated.com":{"auth":"redacted"},"registry.replicated.com":{"auth":"redacted"}}}
```

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,73 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
-type: kubernetes.io/dockerconfigjson
-data:
-  .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
-{{ end }}
-```
+   ```yaml
+   images:
+     pullSecrets:
+       replicated:
+         dockerconfigjson: {{repl LicenseDockerCfg }}
```
ajp-io, line 226: This needs quotes, but there's also some weird issue where `repl` has to be outside of the curly braces for this to work. It should technically work both ways though, it just doesn't right now.
```suggestion
         dockerconfigjson: "repl{{ LicenseDockerCfg }}"
```

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,73 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+   **Example:**
 
-```
+   The following example shows the `LicenseDockerCfg` template function in a `values.yaml` that also defines the URL for the image registry, the image tag, and a pull policy.    
+
+   ```yaml
+   images:
+     pullSecrets:
+       replicated:
+         dockerconfigjson: {{repl LicenseDockerCfg }}
```
ajp-io, line 238:
```suggestion
         dockerconfigjson: "repl{{ LicenseDockerCfg }}"
```

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,73 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+When users install with the kots CLI or the Kubernetes installer, Replicated automatically uses the customer license to create and inject an image pull secret that grants proxy access to a private registry. Instances of your application can then pull private images through the Replicated proxy service. For more information, see [About Connecting to an External Registry](packaging-private-images#about-connecting-to-an-external-registry).
 
-### Example: Delivering Image Pull Secrets for Private Images
+For installations with the helm CLI, Replicated cannot automatically inject an image pull secret into the Helm chart for your application. To support the use of private images for helm CLI installations, use the Replicated `LicenseDockerCfg` template function to render a value based on the unique customer license file. Write the value to a pull secret, then reference the pull secret in the `templates/deployment.yaml` file for the Helm chart.
```
ajp-io, line 214: Could be a deployment, but it could be other types of Kubernetes resources too.
```suggestion
For installations with the helm CLI, Replicated cannot automatically inject an image pull secret into the Helm chart for your application. To support the use of private images for helm CLI installations, use the Replicated `LicenseDockerCfg` template function to render a value based on the unique customer license file. Write the value to a pull secret, then reference the pull secret in the necessary template files for the Helm chart.
```

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,73 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
-  .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
-{{ end }}
-```
+   ```yaml
+   images:
+     pullSecrets:
+       replicated:
+         dockerconfigjson: {{repl LicenseDockerCfg }}
+   ```   
+   Replicated renders the `LicenseDockerCfg` template function on `helm install` or `helm pull`. This allows the customer's license-specific credentials to be injected into `values.yaml`.
```
ajp-io, line 228: `helm upgrade` does too, so we can just say "when the chart is pulled" instead of listing all commands.
```suggestion
   Replicated renders the `LicenseDockerCfg` template function when the chart is pulled. This allows the customer's license-specific credentials to be injected into `values.yaml`.
```

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,73 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
-apiVersion: v1
-kind: Secret
-metadata:
-  name: replicated
-type: kubernetes.io/dockerconfigjson
-data:
-  .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
-{{ end }}
-```
+   ```yaml
```
ajp-io, line 222: The exact format here is up to the vendor. Marc came up with `images.pullSecrets.replicated.dockerconfigjson`, but it could really be anything. The important thing is that there is _some_ value that is populated with the `LicenseDockerCfg` template function. What its path is in the values YAML is irrelevant, as long as the correct path is then used below in the secret file.

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,73 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+   kind: Secret
+   metadata:
+    name: replicated
+   type: kubernetes.io/dockerconfigjson
+   data:
+    .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
+   {{ end }}
+   ```
+
+1. Add the following to the `templates/deployment.yaml` to inject the pull secret that you created in the previous step when the `.Values.images.pullSecrets.replicated` field is present:
```
ajp-io, line 260: I don't know if my wording is perfect, but I'm trying to generalize these things. There may be lots of files that reference private images, and any of them would need this. Also, like I said above, the names we give to these things don't have to be hardcoded. So we named the secret `replicated` in our example, but it could be any name. It just has to match when it is referenced in the manifests.
```suggestion
1. Add the following to any manifests in the `templates` directory that reference private images, to inject the pull secret that you created in the previous step when the `.Values.images.pullSecrets.replicated` field is present. Ensure that the name of the pull secret matches.
```

docs/vendor/helm-install.md
```diff
@@ -18,7 +18,9 @@ When a new version of a chart is uploaded, Replicated renders and cache that ver
 
 When a license field changes for a customer, Replicated invalidates all rendered and cached charts. This causes the charts to be rebuilt the next time they are pulled.
 
-The `values.yaml` in your Chart is rendered with the customer provided license. This creates a way to pass custom license fields and other customer-specific data into the `values.yaml` as defaults, and consume them in Helm templates. When Replicated renders the `values.yaml`, it is not assumed to be a valid YAML format in order to allow flow-control and conditional template functions to optionally write some fields to the `values.yaml` file.
+The `values.yaml` in your Chart is rendered with the customer provided license. Each customer logs in with unique credentials, and our registry is able to identify which customer is pulling the chart. This creates a way to pass custom license fields and other customer-specific data into the `values.yaml` as defaults, and consume them in Helm templates.
```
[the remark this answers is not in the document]
paigecalvert: Nice, I prefer this too

docs/vendor/helm-install.md
paigecalvert: Oh nice catch(es)

docs/vendor/helm-install.md
paigecalvert: Nice, I was thinking I'd edit this step a little more too to make it clear that the Deployment file YAML provided is an example

docs/vendor/helm-install.md
paigecalvert: Ahh, that is very helpful to know. I'll revise this to generalize it (similar to the step of referencing the pull secret in the templates directory) to make it more clear that the specific field names/format is up to the vendor and that this is provided as an example

docs/vendor/helm-install.md
paigecalvert: Cool, right, I remember "does this apply for the Replicated private registry" was an open question we had.

To make sure I understand:
- Both external private registries and the Replicated private registry require an image pull secret 
- Vendors need to do these steps to create a pull secret if they are using either type of private registry, and the steps are the same for both.
- The major distinction between using an external private registry and the Replicated private registry that needs to be made is that: The image pull secret grants access to the Replicated private registry. The image pull secret grants _proxy_ access to an external private registry.

ajp-io: That is all correct. The pull secret that is created works for both proxy.replicated.com and registry.replicated.com, because it contains a reference to both domains, as seen above.

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,86 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+
+   ```yaml
+        ...
+        {{ if .Values.FIELD_NAME }}
+        imagePullSecrets:
+          - name: SECRET_NAME
+        {{ end }}
+   ```
+   Replace:
+   * `FIELD_NAME` with the name of the field where you where you added `"repl{{ LicenseDockerCfg }}"`.
```
paigecalvert, line 280: @ajp-io one more question on this one. I'm unsure about what this FIELD_NAME should be, based on what's in this example YAML.

The example field name for this file is `{{ if .Values.images.pullSecrets.replicated }}`, and I can't tell why it's not `{{ if .Values.images.pullSecrets.replicated.dockerconfigjson }}`

Screen Shot 2022-07-28 at 11 06 48 AM

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,86 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+   kind: Secret
+   metadata:
+     name: replicated
+   type: kubernetes.io/dockerconfigjson
+   data:
+     .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
+   {{ end }}
+   ```
+
+1. Add the following to any manifests in the Helm chart `templates` directory that reference private images to inject the pull secret that you created in the previous step:
```
paigecalvert, line 270: "any manifests" means "any and all manifests", right @ajp-io ? (Like, don't just choose one of them to inject the pull secret.)

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,86 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+1. In the `templates` directory of your Helm chart, create a Secret custom resource manifest file (`kind: Secret`). Add the following YAML to the file to evaluate if the `LicenseDockerCfg` template function renders a value for the customer, then write the rendered value into a Secret on the cluster:
+
+   ```yaml
+   {{ if .Values.FIELD_NAME }}
+   apiVersion: v1
+   kind: Secret
+   metadata:
+     name: SECRET_NAME
+   # Kubernetes clusters use the kubernetes.io/dockerconfigjson Secret type
+   # to authenticate with a private image registry.
```
paigecalvert, line 245: I added the comments in lines 244 and 245 after looking up `type: kubernetes.io/dockerconfigjson`. Calling out just to confirm that it's accurate

docs/vendor/helm-install.md
ajp-io: I actually wondered the same thing. I could speculate, but I'm not confident, so @marccampbell would be a better person to explain why he chose that.

docs/vendor/helm-install.md
ajp-io: No, it means the manifests that reference private images. So like you could have a manifest that defines a Kubernetes service, and that wouldn't reference an image. Or you could have a deployment or StatefulSet that references a public image. So you would add the pull secret information to manifests that reference private images and thus need the pull secret (for access to our registry or proxy access to their registry).

docs/vendor/helm-install.md
ajp-io: That's correct

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,86 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
 
-```
+   ```yaml
+   images:
+     pullSecrets:
+       replicated:
+         dockerconfigjson: "repl{{ LicenseDockerCfg }}"
+   ```   
+
+1. In the `templates` directory of your Helm chart, create a Secret custom resource manifest file (`kind: Secret`). Add the following YAML to the file to evaluate if the `LicenseDockerCfg` template function renders a value for the customer, then write the rendered value into a Secret on the cluster:
```
ajp-io, line 236: Secrets aren't custom resources—they're native Kubernetes resources, like a deployment or service.

docs/vendor/helm-install.md
```diff
@@ -203,44 +205,86 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
-  name: replicated
-type: kubernetes.io/dockerconfigjson
-data:
-  .dockerconfigjson: {{ .Values.images.pullSecrets.replicated.dockerconfigjson }}
-{{ end }}
-```
+   ```yaml
+   FIELD_NAME: "repl{{ LicenseDockerCfg }}"
+   ```
+   Replace `FIELD_NAME` with any name for the field.
```
ajp-io, line 225: Maybe a mention here or in the section below on how `FIELD_NAME` can be a flat or nested value ([Helm docs](https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values)).

ajp-io: Reading that is actually interesting because it notes that you have to check for existence at each layer of nesting. We don't need to worry about it now, but I should test out whether this fails if `.Values.images.pullSecrets`, for example, doesn't exist.

docs/vendor/helm-install.md
paigecalvert: Ah, sorry, my question was unclear. I meant to ask:

Does "any manifests in the Helm chart `templates` directory that reference private images" mean "_all_ manifests in the Helm chart `templates` directory that reference private images"?

Or, does it mean "pick one of the many manifests that might reference a private image and add this YAML to it"

paigecalvert: Writing this out, I figure it is the former

ajp-io: Yeah, the former. If a private image is included, it should reference the pull secret.

docs/vendor/helm-install.md
paigecalvert: Sounds good--I'll include the note about it for now

docs/vendor/helm-install.md
marccampbell: good catch here, you are right!

paigecalvert: thanks Marc for confirming!

docs/vendor/helm-install.md
```diff
@@ -203,58 +205,104 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
 
-```
+   ```yaml
+   images:
+     pullSecrets:
+       replicated:
+         dockerconfigjson: "repl{{ LicenseDockerCfg }}"
+   ```   
+
+1. In the `templates` directory of your Helm chart, create a Kubernetes Secret manifest file (`kind: Secret`). Add the following YAML to the file to evaluate if the `LicenseDockerCfg` template function renders a value for the customer, then write the rendered value into a Secret on the cluster:
```
ajp-io, line 236: That will always render a value for a customer, so I'd probably reword slightly.
```suggestion
1. In the `templates` directory of your Helm chart, create a Kubernetes Secret manifest file (`kind: Secret`). Add the following YAML to the file to evaluate if the secret value is set, and then write the rendered value into a Secret on the cluster:
```

docs/vendor/helm-install.md
ajp-io: I think my confusion was that the first sentence says that when this is rendered, it is assumed to be invalid. But then the last sentence says that we render the functions when the chart is served. Those seem to be a bit contradictory, though I understand what it's getting at. The first sentence might be more accurate if it said, "Before Replicated renders the `values.yaml`."

ajp-io: Also I don't think flow control needs a hyphen. https://helm.sh/docs/chart_template_guide/control_structures/

paigecalvert: This was a Marc Paragraph ™️ , so I won't be able to speak to that directly, but that sounds logical to me

ajp-io: Lol definitely trademarked. It took me a couple reads to understand what it was saying, but I get it now. Changing the first word to "Before" would work, because it's not valid _until_ Replicated renders the template functions when the end-user pulls the chart.

jonquil2002 (approved):
lgtm

paigecalvert:
Merging this PR, then creating a new PR for adding documentation on how to manually reference proxy.replicated.com using template functions from the PodSpec per convo here: https://replicated.slack.com/archives/C02BPJQQP28/p1659381671507469

PR #486, opened 2022-07-26
Clarify wording and procedure

ajp-io:


ajp-io:
One thing Mark mentioned during a demo of this today was that there is no information about how this works. And that's because we have a section on packaging (vendor) and a section on installing (enterprise). So the how does it work is largely in the enterprise docs. Should/can we add a note like, "For more information on how to install using the Kubernetes installer associated with a particular release, see the enterprise docs?"

paigecalvert (approved):

paigecalvert:
I must have missed your comment on this yesterday @ajp-io , but yes! That is a great idea. I can make a PR to add in that cross-reference, or I'll approve a PR from you that adds it, if you beat me to it.

PR #489, opened 2022-07-27
Replace "snapshots" with "Backup and Restore"

paigecalvert:
**Story**: https://app.shortcut.com/replicated/story/54106/replace-snapshots-in-titles-and-headings-with-backup-and-restore 

According to this request from the support team, users are struggling to find information about backup and restore in the docs because they don't know what "snapshot" means. We should rename this in the TOC, titles, and headings to make it easier to find, while explaining the snapshots is the name of the feature in the page content.

**Previews**:

Edited some of the titles/headings in this section of the vendor docs: https://deploy-preview-489--replicated-docs.netlify.app/vendor/snapshots-overview

Also edited titles/headings in these enterprise docs (some light rewording in the content of the topics as well. I couldn't resist): https://deploy-preview-489--replicated-docs.netlify.app/enterprise/snapshots-understanding

### DIFF
--- a/docs/enterprise/snapshots-understanding.md
+++ b/docs/enterprise/snapshots-understanding.md
@@ -1,17 +1,19 @@
-# Understanding Snapshots
+# About Backup and Restore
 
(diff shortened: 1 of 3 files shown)


xavpaice (approved):
This makes sense, thanks

PR #494, opened 2022-07-28
Add timeout troubleshooting description

paigecalvert:
https://app.shortcut.com/replicated/story/54091/document-workaround-for-patching-the-velero-deployment-to-edit-restic-timeout

See the Timeout Error section: https://deploy-preview-494--replicated-docs.netlify.app/enterprise/snapshots-troubleshooting-backup-restore#timeout-error-when-creating-a-snapshot

I also did some reformatting to use the Symptom/Cause/Solution format


xavpaice (dismissed):
LGTM, though do we need to point out that the restic timeout can be overwritten by re-running the install?  Not essential for this commit anyway, it's in the upstream docs.

PR #501, opened 2022-08-03
Add procedure for referencing the proxy service

paigecalvert:
https://app.shortcut.com/replicated/story/54786/add-procedure-for-manually-referencing-the-proxy-service-as-part-of-the-using-private-registries-topic

Preview of changes: 
- Updated the overview content to describe how you also need to manually add the ref to the proxy service if you're using an external registry: https://deploy-preview-501--replicated-docs.netlify.app/vendor/helm-install#private-images
- Created a new procedure in a subsection to describe how to reference the proxy service: https://deploy-preview-501--replicated-docs.netlify.app/vendor/helm-install#proxy-service


docs/vendor/helm-install.md
```diff
@@ -299,9 +313,113 @@ To deliver customer-specific image pull secrets for a private registry:
...
+To update the image URL to reference the proxy service:
+
+1. Create a field in the Helm chart `values.yaml` file with the location of the private image on `proxy.replicated.com`:
+
+   ```yaml
+   FIELD_NAME: PROXY_SERVICE_URL
+   ```
+   Replace:
+   * `FIELD_NAME` with any name for the field.
+   * `PROXY_SERVICE_URL` with the URL for the private image on `proxy.replicated.com`.
```
paigecalvert, line 343: Q: Is there any guidance that we can give on how they figure out what this URL would be? Is it always `proxy.replicated.com/proxy/app-name/[external-registry-url-path]`?

docs/vendor/helm-install.md
```diff
@@ -205,13 +205,27 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+
+To use private images with your application for helm CLI installations, use template functions to inject an image pull secret into the Helm chart. If you are using an external private registry, you must also update the image location URL in the Pod spec in your Helm chart to reference the Replicated proxy service.
 
-For installations with the helm CLI, Replicated cannot automatically inject an image pull secret into the Helm chart for your application. To support the use of private images for helm CLI installations, add the Replicated `LicenseDockerCfg` template function to the Helm chart `values.yaml` file. The `LicenseDockerCfg` template function renders a value based on the unique customer license file when the Helm chart is pulled. Write this rendered value to a pull secret, then reference the pull secret in the necessary template files for the Helm chart.
+For more information, see [Deliver Image Pull Secrets for a Private Registry](#pull-secret) and [Update the Image URL to Reference the Proxy Service](#proxy-service) below.
+
+### Deliver Image Pull Secrets for a Private Registry {#pull-secret}
```
paigecalvert, line 226: Note: No changes to this content other than moving it under a subsection

docs/vendor/helm-install.md
ajp-io: It would be best to get an authoritative opinion. But that is my understanding.

docs/vendor/helm-install.md
```diff
@@ -299,9 +313,113 @@ To deliver customer-specific image pull secrets for a private registry:
...
+:::note
+This procedure is not required if you are using the Replicated private registry.
+:::
+
+To update the image URL to reference the proxy service:
+
+1. Create a field in the Helm chart `values.yaml` file with the location of the private image on `proxy.replicated.com`:
+
+   ```yaml
+   FIELD_NAME: PROXY_SERVICE_URL
```
ajp-io, line 339: Could this be confused as proxy.replicated.com, since that is the URL of the proxy service? Should we call this `PROXY_SERVICE_IMAGE_URL` or `PROXY_SERVICE_IMAGE_REPOSITORY`? I might be reading into it too much.

docs/vendor/helm-install.md
```diff
@@ -299,9 +313,113 @@ To deliver customer-specific image pull secrets for a private registry:
...
+   * `FIELD_NAME` with any name for the field.
+   * `PROXY_SERVICE_URL` with the URL for the private image on `proxy.replicated.com`.
+
+   **Example:**
+
+   ```yaml
+   apiImageRepository: proxy.replicated.com/proxy/my-kots-app/quay.io/my-org/api
+   apiImageTag: v1.0.1
+   ```  
+   The example above shows a field for the image URL and a separate field for the image tag.
```
ajp-io, line 351: From my reading, "image name" or "image repository" seems to be the more typical wording, rather than "image URL."

docs/vendor/helm-install.md
```diff
@@ -299,9 +313,113 @@ To deliver customer-specific image pull secrets for a private registry:
...
+
+1. In the Deployment manifest file in your Helm chart, update the `image` field under `spec` to reference the field you created in the `values.yaml` field. Use the following template function format:
```
ajp-io, line 353: This is similar to the image pull secret stuff, where they need to do this in any files that reference a proxied image. Doesn't have to be a Deployment.

docs/vendor/helm-install.md
```diff
@@ -299,9 +313,113 @@ To deliver customer-specific image pull secrets for a private registry:
...
+   ```yaml
+   apiVersion: kots.io/v1beta1
+   kind: HelmChart
+   metadata:
+     name: samplechart
+   spec:
+     ...
+   ```
+
+   The Replicated HelmChart custom resource allows Replicated to process and deploy Helm charts when users install with the kots CLI or with the Kubernetes installer.
```
ajp-io, line 391: This doesn't feel quite true since we're also using it right here to install with the helm CLI.

docs/vendor/helm-install.md
```diff
@@ -299,9 +313,113 @@ To deliver customer-specific image pull secrets for a private registry:
...
+
+   For more information, see [HelmChart](/reference/custom-resource-helmchart) in the _References_ section.
+
+1. Under the HelmChart custom resource `values` field, create a new field and add the location of the private image on the external registry:
```
ajp-io, line 395: I feel like we should clarify why we're doing this. Like explain how the KOTS installations will use this value, and KOTS will rewrite the image names to use the proxy service, but the Helm installations will use the `values.yaml`, so they need to have the proxy service image name instead.

Maybe we should do that for `admin-console.enabled` too, though that would be a separate PR.

ajp-io (commented):
Looking quite good.

docs/vendor/helm-install.md
```diff
@@ -205,13 +205,27 @@ To conditionally include and exclude the `admin-console` Helm chart:
       ```
   1. Save and promote the release to a development environment to test your changes.      
 
-## Deliver Image Pull Secrets for a Private Registry {#private-images}
+## Using Private Registries {#private-images}
+
+This section describes the steps required to connect to an external private registry or the internal Replicated private registry when your users install your application with the helm CLI.
```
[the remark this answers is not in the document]
paigecalvert: Good call, agreed

docs/vendor/helm-install.md
```diff
@@ -205,13 +205,27 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+
+### Overview of Using Private Registries
 
 Using an external private image registry or the Replicated private registry for your application requires an image pull secret for access. The unique license file for each customer can grant access to the Replicated private registry, or grant proxy access to an external registry without exposing registry credentials to the customer.
 
-When users install with the kots CLI or the Kubernetes installer, Replicated automatically uses the customer license to create and inject an image pull secret. For more information about this process, see [Connecting to an External Registry](packaging-private-images).
+In addition to the image pull secret, using an external private registry also requires that Replicated can access the private image through the Replicated proxy service.
```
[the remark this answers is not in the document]
paigecalvert: Yes, I was trying to think of a generic way to word this that would be true for both KOTS and helm CLI installations. We could add an extra sentence here to explain that for KOTS installations, that requires adding the creds in vendor portal to make that more explicit. Whatcha think?

docs/vendor/helm-install.md
paigecalvert: Ahh yeah, I like that better.

docs/vendor/helm-install.md
paigecalvert: Cool, I think we do use image name for frequently in other places in the docs, so I'll go with that

docs/vendor/helm-install.md
ajp-io: Or I guess image name since that's what you decided on down below.

docs/vendor/helm-install.md
ajp-io: We pretty much only use it for mapping the config options to the values.yaml, but it does get used for that at least.

docs/vendor/helm-install.md
ajp-io: This might actually require a section of its own, to explain this process. It probably doesn't need to be jammed into either of these particular sections.

docs/vendor/helm-install.md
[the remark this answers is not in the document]
ajp-io: That's true for all installations. Anytime you're going to use the proxy service, you have to add your registry creds in the vendor portal, so that the proxy service can access them. So it's not unique to one install method or another.

docs/vendor/helm-install.md
paigecalvert: I removed the "with the kots CLI or with the Kubernetes installer" part. I believe my thought here was that KOTS was actually going to be processing this field that they add, rather than Helm processing/using it.

(I have the hardest time conceptualizing how the file is used by KOTS vs Helm--I recall we had a convo about how the different fields in HelmChart apply to legacy helm install vs native helm install vs helm cli install, but it's fuzzy.)

ajp-io: I see better what you were going for now. Let me take a look and see if I can find wording that works in context here.

docs/vendor/helm-install.md
paigecalvert: I included some explanation of this in the content under the "Update the Image Name to Reference the Proxy Service" heading, before jumping into the procedure. I'm adding a commit to copy a sentence from this description down to this step as well; I can see how it's better to have an explanation here when people are working through the steps. (I went back and forth on this on the first pass)

docs/vendor/helm-install.md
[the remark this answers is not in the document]
paigecalvert: Okay, adding a commit to shuffle around the info a little more in the intro.

docs/vendor/helm-install.md
paigecalvert: Oop, just seeing your comment from earlier (my page wasn't refreshed). Take a look and see what you think about having in the step as well as in the intro content before the procedure :)

ajp-io: And because the relationship between the HelmChart `values` and the `values.yaml` is relevant for several sections (`admin-console.enabled`, Config options, proxied images, we should probably address that relationship really specifically somewhere too.

paigecalvert: Yeah, agree. And that relationship is relevant beyond installing with the helm CLI, so I'm thinking it should be higher level than this topic. We currently have some description of this in the `values` section of the HelmChart reference topic (the one that describes what all the fields are used for): https://docs.replicated.com/reference/custom-resource-helmchart#values

ajp-io: We do use it a bit differently for Helm too. Like with Helm there's a relationship between put stuff here that you want for KOTS installs, and put stuff in `values.yaml` that you want in the Helm install. And that relationship is something we should document. And the only `values` we pay attention to for Helm installs are the ones that are templated. Other ones are ignored. I can explain more elsewhere, but yeah, there are some differences.

paigecalvert: Got it, makes sense

ajp-io (approved):
I think some of this is confusing (not your fault, just the complexity of the different approaches between Helm install and KOTS install). Let's talk sometime soon about how we can clarify those things. The whole "do it this way for Helm but put this in the HelmChart custom resource for KOTS" part is inherently confusing.

PR #508, opened 2022-08-05
HOLD FOR RELEASE update support for image tags and digests

cbodonnell:
Removed the section that states that images and digests cannot be used together.  Also added an additional sentence after the online digest support piece to indicate that both can be used together for online installations (except for those configured to use an internal private registry).

### DIFF
--- a/docs/vendor/packaging-private-images.md
+++ b/docs/vendor/packaging-private-images.md
@@ -175,7 +175,7 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
 
 The app manager supports image tags for applications in all use cases.
 
-The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access. Image tags and digests can also be used together in these use-cases. Image digests are not supported for airgap installations or for online installations that are configured, via the **Registry Settings** tab, to push images to a private registry.
+The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access. Image tags and digests can be used together for these installations too. Image digests are not supported for airgap installations or for online installations that are configured to push images to a private registry.
 
 ## Related Topic
 
--- a/docs/vendor/packaging-private-images.md
+++ b/docs/vendor/packaging-private-images.md
@@ -175,9 +175,7 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
 
 The app manager supports image tags for applications in all use cases.
 
-The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access.
-
-However, image tags and digests cannot be used together due to upstream limitations. For more information, see [Why do i see the error 'Docker references with both a tag and digest are currently not supported'?](https://community.replicated.com/t/why-do-i-see-the-error-docker-references-with-both-a-tag-and-digest-are-currently-not-supported/825) on the Replicated Community website.
+The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access.  Image tags and digests can also be used together for online installations.
 
 ## Related Topic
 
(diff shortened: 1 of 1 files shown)


docs/vendor/packaging-private-images.md
```diff
@@ -175,9 +175,7 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
 
 The app manager supports image tags for applications in all use cases.
 
-The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access.
-
-However, image tags and digests cannot be used together due to upstream limitations. For more information, see [Why do i see the error 'Docker references with both a tag and digest are currently not supported'?](https://community.replicated.com/t/why-do-i-see-the-error-docker-references-with-both-a-tag-and-digest-are-currently-not-supported/825) on the Replicated Community website.
+The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access.  Image tags and digests can also be used together for online installations.
```
ajp-io, line 178: @jonquil2002 We're working on making image digests work for air gap too, so this section will be completely different before long. But in the short run, I'm a bit confused by the first sentence here. "The Replicated private registry, a public external registry, or from a private external registry through proxy access"—that is all the possible scenarios that I can think of. Do we need to enumerate those? Can we just say that it works for online (Internet-connected" installations?

jonquil2002: @ajp-io I think some clarity would be good here. If you feel it is too wordy as is, then perhaps change it to "The app manager supports image digests only for inline (Internet-connected) installations where the app manager can pull all images from the registry you specify." Or something along those lines - does that work for you?

jonquil2002 (approved):
lgtm

docs/vendor/packaging-private-images.md
```diff
@@ -175,9 +175,7 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
 
 The app manager supports image tags for applications in all use cases.
 
-The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access.
-
-However, image tags and digests cannot be used together due to upstream limitations. For more information, see [Why do i see the error 'Docker references with both a tag and digest are currently not supported'?](https://community.replicated.com/t/why-do-i-see-the-error-docker-references-with-both-a-tag-and-digest-are-currently-not-supported/825) on the Replicated Community website.
+The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access.  Image tags and digests can also be used together for online installations, excluding those configured to use an internal private registry.
```
ajp-io, line 178: @cbodonnell What is the significance of "internal" in "internal private registry?" Is this only true for a private registry that's in the cluster?

cbodonnell: @ajp-io It is more so that we don't support image tags and digests together (or really digests at all for that matter) when the admin console is configured to push images to a private registry in the "Registry Settings" page.  I've re-worked that paragraph to better indicate the use cases that support image digests.

docs/vendor/packaging-private-images.md
```diff
@@ -175,9 +175,7 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
 
 The app manager supports image tags for applications in all use cases.
 
-The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access.
-
-However, image tags and digests cannot be used together due to upstream limitations. For more information, see [Why do i see the error 'Docker references with both a tag and digest are currently not supported'?](https://community.replicated.com/t/why-do-i-see-the-error-docker-references-with-both-a-tag-and-digest-are-currently-not-supported/825) on the Replicated Community website.
+The app manager supports image digests only for online (Internet-connected) installations where the app manager can pull all images from the Replicated private registry, a public external registry, or from a private external registry through proxy access. Image tags and digests can be used together for these installations too. Image digests are not supported for airgap installations or for online installations that are configured to push images to a private registry.
```
ajp-io, line 178: @jonquil2002 Do we have any docs on setting up a private registry in the admin console? Only reference I found is one sentence [here](https://docs.replicated.com/enterprise/image-registry-embedded-cluster#restoring-deleted-images).

jonquil2002: @ajp-io We do not seem to have this. It would be my pleasure to write this up - I see where it lives in the admin console. I'll ping you to hear what you think about best ways to approach this.

ajp-io (approved):

PR #521, opened 2022-08-11
Update team-management-google-auth.md

JustinN-Replicated:


jonquil2002 (dismissed):
lgtm with my commits

JustinN-Replicated:
@jonquil2002 thanks! Will have the team review as well - I want to make sure everything is technically accurate; one thing that struck me was the 'toggle table' (ie the name of the toggle and description) having a bit of a 'here is what it does' and in some cases some details or description; I broke some of this out for the new toggle because it seemed like it would be too much detail for that line item in the table (maybe we should consider doing similar for the other toggles? Not sure what folks think reads better / is easier to follow (can always link to the 'additional info sections' below too; will hold off adding this though until the above is discussed.

JustinN-Replicated:
@jonquil2002 went with the former (removed additional section and added to the listing in the table). Will review this table in more detail later today too.

jodillion:
[sc54772]

jodillion:
[sc-54772]

paigecalvert:
@JustinN-Replicated with Jonquil out this week, give me a ping if/when you need another review on this story.

JustinN-Replicated:
@paigecalvert should know more after planning today; once the third setting in google auth is no longer behind a feature flag though this will need to go live with those changes. Will keep you posted but this will be by EOW at a maximum.

JustinN-Replicated:
@paigecalvert  and I reviewed on a working session today.

paigecalvert (approved):
@JustinN-Replicated made a few docs/language edits. Take a look and make sure they are OK with you, then feel free to merge if so.

Specifically, I:
- Removed some redundant language (some of the details already appeared in the second step of the invite members procedure)
- Pulled some info out of the Note into plain text
- Re-worded for tense and such

PR #517, opened 2022-08-10
Default to Helm v3

ajp-io:
Reverts replicatedhq/replicated-docs#516

Don't merge until the next KOTS release

PR #522, opened 2022-08-11
change helmVersion default to v3

cbodonnell:
Changing the default value for `helmVersion` to `v3` to document the changes in: https://github.com/replicatedhq/kots/pull/3041.

Additionally, the example manifest has a `helmVersion: v2` along with `useHelmInstall: true`, but native helm installations are only available for `v3` charts, so I changed this so I modified the example to be a valid combination of the two.


ajp-io:
I should have mentioned that I opened #517 already. Since I changed a few more things, I'll close this in favor of mine, which is already approved too.

PR #533, opened 2022-08-16
Update team-management.md

JustinN-Replicated:


JustinN-Replicated:
[sc-55928]

docs/vendor/team-management.md
```diff
@@ -32,6 +32,9 @@ To re-invite a prospective member:
 
 1. Click **Reinvite** in the confirmation dialog.
 
+## Allow Users From a Domain to Join Your Team
+By default, users must be invited to your team. Team administrators may choose to allow users from the same email domain to join their team automatically. Users who join with this method will be automatically assigned the RBAC policy specified when enabling this feature. When enabled, users must validate that they have access to the email address used when signing up on vendor portal by providing an activation code. After validation, the user may either create their own independent team or join your team.
```
alicenstar, line 36: Is this the right docs page for this? It only applies to Google accounts right now, and this doesn't specify that

JustinN-Replicated: Originally yes, intent was to have this separate. After changes from today though this will be changing a bit.

docs/vendor/team-management.md
```diff
@@ -32,6 +32,9 @@ To re-invite a prospective member:
 
 1. Click **Reinvite** in the confirmation dialog.
 
+## Allow Users From a Domain to Join Your Team
+By default, users must be invited to your team. Team administrators may choose to allow users from the same email domain to join their team automatically after they have registered their email address and password or when authenticating with Google authentication. Users who join with this method will be automatically assigned the RBAC policy specified when this feature is enabled. By default, the RBAC policy field for auto-join users is set to the built-in read-only policy. Once enabled, email/password users must validate that they have access to the email address used when signing up on vendor portal by providing an activation code; after validation, the user will be added as a team member to this team.
```
JustinN-Replicated, line 36: Removed from google maybe add here?
> Additionally, you can allow all users from your Google Workspace to join your team automatically.

docs/vendor/team-management.md
```diff
@@ -32,6 +32,12 @@ To re-invite a prospective member:
 
 1. Click **Reinvite** in the confirmation dialog.
 
+## Allow Users to Auto-join Your Team
+By default, users must be invited to your team. Team administrators may choose to allow users from the same email domain to join their team automatically. When enabled, this setting allows a user to automatically join a team once they have signed up in vendor portal and activated their account. Users who join a team via Auto-join will be automatically assigned the RBAC policy configured on this setting. By default, the RBAC policy field for auto-join users is set to the built-in 'Read Only' policy. This feature does not apply to SAML authentication as SAML users will authenticate via their SAML provider's application portal and do not log in through vendor portal directly.
```
markpundsack, line 36: Too many "automatically"s. 

```suggestion
By default, users must be invited to your team. Team administrators may choose to allow users from the same email domain to join their team automatically. Users who join a team via Auto-join will be assigned the RBAC policy configured on this setting. By default, the RBAC policy field for auto-join users is set to the built-in 'Read Only' policy. This feature does not apply to SAML authentication as SAML users will authenticate via their SAML provider's application portal and do not log in through vendor portal directly. 
```

docs/vendor/team-management.md
```diff
@@ -32,6 +32,12 @@ To re-invite a prospective member:
 
 1. Click **Reinvite** in the confirmation dialog.
 
+## Allow Users to Auto-join Your Team
+By default, users must be invited to your team. Team administrators may choose to allow users from the same email domain to join their team automatically. When enabled, this setting allows a user to automatically join a team once they have signed up in vendor portal and activated their account. Users who join a team via Auto-join will be automatically assigned the RBAC policy configured on this setting. By default, the RBAC policy field for auto-join users is set to the built-in 'Read Only' policy. This feature does not apply to SAML authentication as SAML users will authenticate via their SAML provider's application portal and do not log in through vendor portal directly. 
+
+### Auto-join and Google Authentication
```
markpundsack, line 38: I'm not sure why this is a section. We say above it doesn't apply to SAML, so it implies it works for everything else. If we want to be explicit, how about just adding a sentence in the earlier paragraph? For example, the feature itself says "This applies to users registering with an email/password or via Google Authentication if it is enabled for the team. This feature does not apply to SAML authentication."

JustinN-Replicated: I've updated this with your suggestions

jonquil2002 (approved):
lgtm with my commits.

PR #536, opened 2022-08-17
Add helm install info to overview topic

paigecalvert:
This PR:
- Updates the overview topic in the Helm section to better incorporate information about the helm CLI installation method.
- As part of this, it also reorganizes the subnav of the Helm section so that the topics are better grouped into subcategories, reducing noise and making it easier to skim.
- Moves the Helm section closer to the top of the Packaging section, where it's easier to find. 
**Note**: The PR does not yet update xrefs or redirects based on any topic name changes. Needs to be done before merging

Story: https://app.shortcut.com/replicated/story/52755/update-helm-overview-to-better-incorporate-info-about-direct-helm-install

Preview: https://deploy-preview-536--replicated-docs.netlify.app/vendor/helm-overview


docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+### Using the kots CLI
 
-:::note
-Migrating existing installations to the native Helm workflow is not supported. However, new Helm charts within an existing application      can leverage the workflow and the features that come with it.
-:::
+Users can install an application packaged with Helm charts using the kots CLI on an existing cluster or on a cluster provisioned by the Kubernetes installer.
 
-It is also possible for your customers to install your application with Helm rather than the app manager. This is an alpha feature. To learn more, reach out to your contacts at Replicated.
+The kots CLI installs applications packaged with Helm charts using either the "Native Helm" or "Replicated KOTS" deployment method. You specify the deployment method in the Replicated HelmChart custom resource manifest file with the `useHelmInstall` flag. For more information, see [useHelmInstall](/reference/helmchart#usehelminstall) in _HelmChart_.
```
ajp-io, line 17: Is the change in name for the old method to disambiguate? Because Replicated Helm could sound like the Helm CLI install method?

We need some consistency on the naming for our old Helm support. I see "Replicated KOTS" here, "Replicated KOTS Helm" below, and we have historically called it "Replicated Helm". I think it needs Helm in there somewhere, else it isn't clear it has to do with Helm.

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+
+The following describes the Native Helm and Replicated KOTS deployment methods:
+
+* **Native Helm (Recommended)**: The kots CLI uses the Helm binary to install and manage the lifecycle of the chart resources that are part of the application. This is the preferred method because it supports more features of Helm, such as hooks and weights.
+
+   For more information, see see [Native Helm](helm-processing#native-helm) in _How the App Manager Processes Helm Charts_. See also [Enabling and Using Native Helm Charts](helm-installing-native-helm) in _Installing with Native Helm_.
```
ajp-io, line 23:
```suggestion
   For more information, see see [Native Helm](helm-processing#native-helm) in _How the App Manager Processes Helm Charts_. See also [Enabling and Using Native Helm Charts](helm-installing-native-helm#enabling-and-using-native-helm-charts) in _Installing with Native Helm_.
```

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
 
-There are two ways to deploy Helm charts within an application:
+## How Replicated Deploys Helm Charts
 
-* With the default workflow, the app manager renders the Helm templates and deploys them as standard Kubernetes manifests. The app manager directly manages the lifecycle of the resources in this workflow, and it supports rendering the templates with Helm V3.
+When you distribute an application packaged with Helm, your customers can install and manage the application with the Replicated kots CLI or with the helm CLI.
 
-* The native Helm workflow deploys the Helm charts with Helm V3 directly. In this workflow, Helm installs and manages the lifecycle of the chart resources that are part of a given application. For new applications, this workflow is the preferred method because it supports more features of Helm, such as hooks and weights. For more information, see [Installing with Native Helm](helm-installing-native-helm).
+### Using the kots CLI
```
ajp-io, line 13: I don't know that I would limit this to the CLI, since you can deploy from the admin console UI too. It's really just KOTS, or the app manager, I should say.

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+The kots CLI installs applications packaged with Helm charts using either the "Native Helm" or "Replicated KOTS" deployment method. You specify the deployment method in the Replicated HelmChart custom resource manifest file with the `useHelmInstall` flag. For more information, see [useHelmInstall](/reference/helmchart#usehelminstall) in _HelmChart_.
+
+The following describes the Native Helm and Replicated KOTS deployment methods:
+
+* **Native Helm (Recommended)**: The kots CLI uses the Helm binary to install and manage the lifecycle of the chart resources that are part of the application. This is the preferred method because it supports more features of Helm, such as hooks and weights.
+
+   For more information, see see [Native Helm](helm-processing#native-helm) in _How the App Manager Processes Helm Charts_. See also [Enabling and Using Native Helm Charts](helm-installing-native-helm) in _Installing with Native Helm_.
+
+* **Replicated KOTS**: The kots CLI renders the Helm templates and deploys them as standard Kubernetes manifests using `kubectl apply`. The kots CLI manages the lifecycle of the resources.
```
ajp-io, line 25: Same comment as above. More app manager than kots CLI.

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+
+   For more information, see [Replicated KOTS](helm-processing#replicated-kots) in _How the App Manager Processes Helm Charts_.
+
+### Using the helm CLI (Alpha)
+
+Users can install an application packaged with a Helm chart onto an existing cluster using the helm CLI. When users install with the helm CLI directly, Helm, rather than the admin console, manages the lifecycle of the application.
```
ajp-io, line 31: NBD but I think I see "into a cluster" more often than "onto a cluster." Also, I think the difference between admin console and app manager/KOTS is tenuous at best, but in this case I think we should say "app manager" instead of "admin console." Because we did name our Helm chart "admin-console," and I think Marc's reasoning was that it's not the app manager, which manages the app, it's just the console piece. I think it's weird semantics and we should polish our terminology, but anyway, in this case I think we should say "app manager."
```suggestion
Users can install an application packaged with a Helm chart into an existing cluster using the helm CLI. When users install with the helm CLI directly, Helm, rather than the app manager, manages the lifecycle of the application.
```

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+
+Deploying an application with the helm CLI differs from the kots CLI "Native Helm" deployment method described above because, when users install with the helm CLI directly, they have access to all Helm functionality. Some enterprise users also prefer or require using the helm CLI because their existing CI pipeline is already compatible with Helm charts.
```
ajp-io, line 33: It's more the CD part, so let's say CI/CD pipeline. And I don't know if my suggestion is the right wording, but I wanted to convey that their pipeline might be compatible, but they might also have an organizational policy that says deployments should be done via Helm.
```suggestion
Deploying an application with the helm CLI differs from the kots CLI "Native Helm" deployment method described above because, when users install with the helm CLI directly, they have access to all Helm functionality. Some enterprise users also prefer or require using the helm CLI because their existing CI/CD pipeline or policies are already compatible with Helm charts.
```

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+
+Users do not have access to certain Replicated features when they install and manage the application with the helm CLI directly. This is cecause the kots CLI does not manage the lifecycle of the application. For example, users must update the application using the `helm upgrade` command, rather than using the admin console UI or the kots CLI.
```
ajp-io, line 35:
```suggestion
Users do not have access to certain Replicated features when they install and manage the application with the helm CLI directly. This is because the kots CLI does not manage the lifecycle of the application. For example, users must update the application using the `helm upgrade` command, rather than using the admin console UI or the kots CLI.
```

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+
+For more information about how to package an application so that users can install using the helm CLI, see [Using Helm to Install an Application (Alpha)](helm-install).
```
ajp-io, line 37: This title changed to "Supporting helm CLI Installations" in another PR.

ajp-io (changes requested):
Is the "Select Helm Install Method" image used anywhere?

docs/vendor/helm-overview.md
paigecalvert: Yeah, exactly. I was trying to stick with what appears in the UI for now; felt like the safest bet. Though "Replicated KOTS" going forward is likely not what we'll want to go with? I'll defer to you.

docs/vendor/helm-overview.md
paigecalvert: Okay, works for me. I've been struggling with when to use admin console vs app manager vs Replicated vs KOTS vs kots CLI lately. It often feels like "KOTS" is most accurate, but of course we want to pick the best commercial term for these docs. (Probably something that I need to raise with the marketing/content folks)

docs/vendor/helm-overview.md
paigecalvert: I'll go through the replace kots CLI with app manager throughout the topic where appropriate 👍

docs/vendor/helm-overview.md
paigecalvert: Ah got it. I'll commit this, and see if I can take another pass at it

paigecalvert:
> Is the "Select Helm Install Method" image used anywhere?

It's not...I need to remove that. Thanks 👍

docs/vendor/helm-overview.md
paigecalvert: The other thing to consider is that it might not really need a name, when there isn't the Native Helm vs Replicated KOTS options in the vendor portal. We'll just have: either deploy helm charts with the app manager, or deploy them with the helm CLI. And here's the details of how the app manager does it.

docs/vendor/helm-overview.md
ajp-io: I think adding app manger into the mix arguably made it more confusing. I am increasingly thinking we should have a higher level discussion about all these names.

paigecalvert: Yes, I agree. I can kick off that convo with marketing & product. In the short term, do you have a preference for this topic? We could pair on that if you want. LMK.

paigecalvert: Posted an initial feeler in the #product-marketing channel: https://replicated.slack.com/archives/C018N17NATS/p1660843254569209

docs/vendor/helm-overview.md
ajp-io: Yeah, the ideal is to have no name for it. One of our upcoming priorities is to create a migration path from the old way to the new "Native Helm" way, and then we'll deprecate the old way and just have one way to install Helm charts in KOTS. But we need a migration path first, so it'll take a bit of time before we do that.

But in the interim, I guess we can stick with what the UI says, though I definitely find it confusing, since both are really using KOTS. We could try to say "old way" and "new way" (better words, but that kind of gist), but that would be a bigger change too.

IMO we can stick with what's in the UI for the purpose of this docs update, but we can separately consider whether we can come up with better terms for those, or whether we wait until we deprecate the old way.

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
-# Helm Overview
+# About Packaging with Helm
 
-Helm is a popular package manager for Kubernetes applications.
+Helm is a popular package manager for Kubernetes applications. If your application is already packaged using Helm, you can use Replicated to more easily distribute and manage your application. Using Replicated to distribute applications packaged with Helm provides additional functionality not available through Helm, such as preflight checks, support bundles, a user interface for collecting user configuration values, and more.
```
ajp-io, line 3: Let's add a note about support for private images. That's a big advantage over Helm. With Helm, you have to create image pull secrets, give them to your customers, and have them deploy them in their cluster in order to pull your private images. With the Replicated registry and the proxy service, we automate that so you can just reference your private images, and we'll make sure they get pulled.

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
 
-For applications that are already packaged using Helm, Helm support in the Replicated app manager can help get an application packaged faster. The app manager supports delivering Helm charts as an enterprise application, or including Helm charts as components of an application. An application can support more than one Helm chart, and can support more than a single instance of any Helm chart.
+Replicated supports delivering an enterprise application as one or more Helm charts, or including Helm charts as components of an application. An application can use more than one Helm chart, and can use more than a single instance of any Helm chart.
```
ajp-io, line 5: Sorry, this was my edit, but the "one or more" thing twice is bothering me now :)
```suggestion
Replicated supports delivering an enterprise application as Helm charts, or including Helm charts as components of an application. An application can use more than one Helm chart, and can use more than a single instance of any Helm chart.
```

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
 
-It is also possible for your customers to install your application with Helm rather than the app manager. This is an alpha feature. To learn more, reach out to your contacts at Replicated.
+The app manager installs applications packaged with Helm charts using either the "Native Helm" or "Replicated KOTS" deployment method. You specify the deployment method in the Replicated HelmChart custom resource manifest file with the `useHelmInstall` flag. For more information, see [useHelmInstall](/reference/helmchart#usehelminstall) in _HelmChart_.
+
+The following describes the Native Helm and Replicated KOTS deployment methods:
+
+* **Native Helm (Recommended)**: The app manager uses the Helm binary to install and manage the lifecycle of the chart resources that are part of the application. This is the preferred method because it supports more features of Helm, such as hooks and weights.
+
+   For more information, see see [Native Helm](helm-processing#native-helm) in _How the App Manager Processes Helm Charts_. See also [Enabling and Using Native Helm Charts](helm-installing-native-helm#enabling-and-using-native-helm-charts) in _Installing with Native Helm_.
```
ajp-io, line 23:
```suggestion
   For more information, see [Native Helm](helm-processing#native-helm) in _How the App Manager Processes Helm Charts_. See also [Enabling and Using Native Helm Charts](helm-installing-native-helm#enabling-and-using-native-helm-charts) in _Installing with Native Helm_.
```

docs/vendor/helm-overview.md
```diff
@@ -1,19 +1,37 @@
...
+
+* **Replicated KOTS**: The app manager renders the Helm templates and deploys them as standard Kubernetes manifests using `kubectl apply`. The app manager manages the lifecycle of the resources.
```
ajp-io, line 25: In a separate PR we should make a stronger statement here about how people shouldn't use this method unless they're already using it. Something about how Native Helm is the preference moving forward. Maybe even that this old way will be removed in the future. But separate PR, no need here.

ajp-io (changes requested):
A few small comments, but this is about ready to go.

docs/vendor/helm-overview.md
paigecalvert: Sounds good. Yeah, the naming is not great. If deprecating the two options is more in the "months" timeframe, something we could consider is updating both the naming in the UI and in the docs, to make it consistent. (Your "old/new" idea gave me the idea that maybe we could try something like: "Replicated Helm v1" and "Replicated Helm v2", but yeah something we can think about more later on).

docs/vendor/helm-overview.md
paigecalvert: Sounds good. I suspect there will be a few places where we can that out too.

paigecalvert:
@ajp-io -- incorporated the latest round of feedback. Could you give it one more pass and merge if you think it's ready?

ajp-io (approved):

PR #540, opened 2022-08-17
update install instructions for embedded tutorial

erquhart:
Provided updates to clear up some confusing bits in the embedded tutorial, specifically around the getting started steps. These are recommendations, won't be surprised if some or all or rejected! Hope it's helpful.

- Condense instruction about using GCP and note about using other providers to a single instruction, avoids recommending GCP above others and helps clarify that all methods work equally
- Remove extra steps from install instructions. The output from kurl install is translated into steps in the tutorial, but this is a hand holding walkthrough and some steps are out of scope, and are also printed verbatim in the kurl install output for users that want them:
  - instructions to access Grafana
  - instructions to add worker nodes
- Remove prompt from `kubectl get pods` command block - this block has a copy/paste button, but it copies the prompt too, so the pasted command doesn't work
- Update example successful `kubectl get pods` output to be match what a user following this particular tutorial will see (operator and migration pods won't be there, for example)
- Added a note calling out the need to add firewall rules for many cloud providers - I ran into this a while back and we weren't sure why I couldn't successfully follow this tutorial, but it was due to the default GCP firewall and AWS has a similar default setup

### DIFF
--- a/docs/vendor/tutorial-installing-without-existing-cluster.md
+++ b/docs/vendor/tutorial-installing-without-existing-cluster.md
@@ -114,12 +114,16 @@ The app manager can be installed either into an existing Kubernetes cluster or a
 
 To create the test server and install the app manager:
 
-1. Create a server using any cloud provider (GCP, AWS, etc.) or a local virtual machine. Ensure the following criteria:
+1. Create a server using any cloud provider (such as GCP or AWS). Or, use a local virtual machine. The server must meet the following criteria:
 
     * Ubuntu 18.04
     * At least 8 GB of RAM
     * 4 CPU cores
     * At least 50GB of disk space
+
+    :::note
+    If you use a virtual machine that is behind a firewall, make sure that port 8800 (and any other ports you attempt to access through the internet) are allowed to accept traffic. GCP and AWS typically require firewall rule creation to expose ports.
+    :::
 
 1. Use SSH to access the server you just created.
 1. Run the installation script:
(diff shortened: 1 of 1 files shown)


docs/vendor/tutorial-installing-without-existing-cluster.md
```diff
@@ -151,36 +150,22 @@ To create the test server and install the app manager:
...
-1. Run the following script on your other nodes to add worker nodes to this installation:
-
+  bash -l
   ```
-    curl -sSL https://kurl.sh/starter-kots-demo-unstable/join.sh | sudo bash -s kubernetes-master-address=[ip-address]:6443 kubeadm-token=[token] kubeadm-token-ca-hash=sha256:[sha] kubernetes-version=1.16.4 docker-registry-ip=[ip-address]
-
-    ```
 
-1. Reload the shell, following the instructions on the screen, to make `kubectl` now work:
+1. Run a `kubectl` command to test that `kubectl` is working:
```
paigecalvert, line 156: @erquhart I added this phrase to clarify what they are testing. Is this accurate?

docs/vendor/tutorial-installing-without-existing-cluster.md
```diff
@@ -114,16 +114,16 @@ The app manager can be installed either into an existing Kubernetes cluster or a
...
 
     * Ubuntu 18.04
     * At least 8 GB of RAM
     * 4 CPU cores
     * At least 50GB of disk space
 
-  :::note
-  You can also use any cloud provider or local virtual machine.
-  :::
+    :::note
```
paigecalvert, line 124: I copied the note that you added up here as well, since it might be something they'd want to consider before they start using the local virtual machine.

paigecalvert (approved):
Looks great, thank you @erquhart !

I added some minor edits (I added a couple comments calling out the changes). If my changes look OK to you, feel free to merge!

docs/vendor/tutorial-installing-without-existing-cluster.md
erquhart: Yeah that one bugged me too, much better this way

erquhart:
Great edits, thank you @paigecalvert

PR #541, opened 2022-08-17
Fix readonly link and typos

ajp-io:


paigecalvert (approved):
thanks @ajp-io. I noticed a couple more too

PR #543, opened 2022-08-18
Update team-management.md

JMNordeste-dev:
from discussions with @GraysonNull and @markpundsack

### DIFF
--- a/docs/vendor/team-management-saml-auth.md
+++ b/docs/vendor/team-management-saml-auth.md
@@ -22,10 +22,6 @@ If SAML authentication is configured for your team, Two-Factor Authentication (2
 
 Replicated supports Role Based Access Control (RBAC) in the vendor portal. To use RBAC with SAML, you must configure policies and add users to the policies by their username. Usernames are the identity of the user in your identity provide (IDP). Typically, this username is the full email address. For more information about configuring RBAC, see [Creating RBAC Policies](team-management-vendor-rbac).
 
-:::note
-When SAML is enabled and a user has the built-in read-only policy assigned, the user will not be able to see what RBAC policy names are assigned to which team members.
-:::
-
 ## Downloading Certificates from Supported SAML providers
 
 You must retrieve the metadata and .509 public certificate files from your SAML provide before configuring Replicated.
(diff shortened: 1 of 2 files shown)


JustinN-Replicated:
This is what it looks like if you are in a SAML account for team members list (I believe I am logged in with a read-only account)
image
This is what it looks like for read-only in a team without SAML:
image

JustinN-Replicated:
@replicatedhq/replicated-docs this is not urgent, just something that came up so wanted to make sure the PR was logged. Please review at your leisure

docs/vendor/team-management-saml-auth.md
```diff
@@ -22,6 +22,10 @@ If SAML authentication is configured for your team, Two-Factor Authentication (2
 
 Replicated supports Role Based Access Control (RBAC) in the vendor portal. To use RBAC with SAML, you must configure policies and add users to the policies by their username. Usernames are the identity of the user in your identity provide (IDP). Typically, this username is the full email address. For more information about configuring RBAC, see [Creating RBAC Policies](team-management-vendor-rbac).
 
+:::note
+When SAML is enabled and a user has the built-in read-only policy assigned, the user will not be able to see what RBAC policy names are assigned to which team members.
```
paigecalvert, line 26: One question on this is what the vendor would do with this information:
- Is it important that they know this ahead of time? 
- Would it affect their decision on whether or not to use SAML? 
- Is hiding the user roles for SMAL-enabled + read-only accounts an important feature that vendors want to make sure we do? 

I think I'm just lacking that background to be able to best advise on where/how this content should be included

paigecalvert:
@JustinN-Replicated thank you Justin! I added a question on this in a comment above. I also removed the bug label. "Bug" is not generally a concept that applies to documentation, unless maybe something is broken with the website itself, or there's an error in markdown formatting.

paigecalvert:
@jonquil2002 , I made some style updates. Specifically, I removed the note from the SAML topic so that all this information appears under the Viewing Team Members heading. Also did some rewording to shorten it up and make sure consistent terminology is used

jonquil2002 (approved):
lgtm

PR #551, opened 2022-08-23
HOLD FOR RELEASE: Supporting image digests for air gap bundles

paigecalvert:
https://app.shortcut.com/replicated/story/55262/orange-team-image-digest-tagging-sha-in-air-gap

Preview: https://deploy-preview-551--replicated-docs.netlify.app/vendor/packaging-private-images#using-image-tags-and-digests 

Made the following improvements:
- Added details on when image tags and digests are supported.
- Added a procedure for enabling the new air gap bundle format to support image digests, including steps on how to test changes


docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,77 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+  <tr>
+    <td>Online</td>
+    <td>Supported by default</td>
+    <td>Supported by default</td>
+  </tr>
+  <tr>
+    <td>Air Gap</td>
+    <td>Supported by default</td>
+    <td>
+    <p>Supported for applications on the app manager VERSION NUMBER and later when the <b>Enable new air gap bundle</b> toggle is enabled on the channel.</p>
```
ajp-io, line 197:
```suggestion
    <p>Supported for applications on the app manager v1.82.0 and later when the <b>Enable new air gap bundle</b> toggle is enabled on the channel.</p>
```

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,77 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+  </tr>
+</table>
+
+:::note
+You can use image tags and image digests together in any case where both are supported.
+:::
+
+### Using Image Digests in Air Gap Installations {#digests-air-gap}
+
+For applications installed with the app manager VERSION NUMBER or later, you can enable a format for air gap bundles that supports the use of image digests for private images. This air gap bundle format also ensures that identical image layers are not duplicated, resulting in a smaller air gap bundle size.
```
ajp-io, line 209: Put version number in there. And digests work for all images (public too), not just private.
```suggestion
For applications installed with the app manager v1.82.0 or later, you can enable a format for air gap bundles that supports the use of image digests. This air gap bundle format also ensures that identical image layers are not duplicated, resulting in a smaller air gap bundle size.
```

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,77 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+
+You can enable or disable this air gap bundle format using the **Enable new air gap bundle** toggle in the settings for any channel in the vendor portal.
+
+When you enable **Enable new air gap bundle** on a channel, all air gap bundles that you build or rebuild on that channel use the updated air gap bundle format.
+
+If users on a version of the app manager earlier than VERSION NUMBER attempt to install or upgrade an application with an air gap bundle that uses the **Enable new air gap bundle** format, then the admin console displays an error message when they attempt to upload the bundle.
```
ajp-io, line 215:
```suggestion
If users on a version of the app manager earlier than v1.82.0 attempt to install or upgrade an application with an air gap bundle that uses the **Enable new air gap bundle** format, then the admin console displays an error message when they attempt to upload the bundle.
```

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,77 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+
+To enable the new air gap bundle format on a channel:
+
+1. In the Replicated [vendor portal](https://vendor.replicated.com/channels), go to **Channels** and click the **edit** icon in the top right of the channel where you want to use the new air gap bundle format.
+1. Enable the **Enable new air gap bundle** toggle.
+1. (Recommended) To prevent users on a version of the app manager earlier than VERSION NUMBER from attempting to upgrade with an air gap bundle that uses the **Enable new air gap bundle** format, set `minKOTSVersion` to VERSION NUMBER in the Application custom resource manifest file.
```
ajp-io, line 221:
```suggestion
1. (Recommended) To prevent users on a version of the app manager earlier than v1.82.0 from attempting to upgrade with an air gap bundle that uses the **Enable new air gap bundle** format, set `minKOTSVersion` to "1.82.0" in the Application custom resource manifest file.
```

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,77 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+   **Example**:
+
+   ```yaml
+   apiVersion: kots.io/v1beta1
+   kind: Application
+   metadata:
+     name: my-application
+   spec:
+     ...
+     minKotsVersion: "VERSION NUMBER"
```
ajp-io, line 234:
```suggestion
     minKotsVersion: "1.82.0"
```

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,77 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+     ...
+   ```
+
+1. Test your changes:
+   1. Save and promote the release to a development environment.
+   1. On the channel where you enabled **Enable new air gap bundle**, click **Release history**. On the Release History page, click **Build** or **Rebuild** to create an air gap bundle with the new format.
```
ajp-io, line 240: Technically if they just promoted the release, they wouldn't have a "Rebuild" option. Unless they have their channel set up to automatically build air gap bundles, in which case it would have been built using the new format since we enabled the toggle in the steps above. So we can probably remove "Rebuild", though we could consider mentioning the auto-build capability if we want to cover everyone's possible scenarios.

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,77 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+
+      ![Vendor portal release history page](../../static/images/airgap-download-bundle.png)
+
+   1. Install or upgrade the application using the admin console. Upload the new air gap bundle to confirm that the installation or upgrade completes successfully.
```
ajp-io, line 244: Any need to mention upgrading the app manager to 1.82.0 or later first? Depends on whether this is an initial install or upgrade I suppose.

ajp-io (changes requested):

docs/vendor/packaging-private-images.md
paigecalvert: Want to make sure I understand correctly. I was thinking that if they promote the release to a channel where they have already promoted releases in the past, then they would be able to Rebuild an air gap build from a previous release to test the toggle. (Like this use case described above: "When you enable **Enable new air gap bundle** on a channel, all air gap bundles that you build or rebuild on that channel use the updated air gap bundle format.")

 Is this the case? If so, maybe it would be helpful for me to clarify that they can click "Build" on the air gap bundle for the release that they just promoted, or they can click "Rebuild" on an air gap bundle from a previous release promoted to that channel?

paigecalvert: (Or, of course, the auto-build scenario that you described above, in which they wouldn't really need to click anything, except Download)

docs/vendor/packaging-private-images.md
paigecalvert: Oh, yeah, that could be worth mentioning. "Install or upgrade the application with version 1.82.0 or later of the admin console or the kots CLI. Upload..."

docs/vendor/packaging-private-images.md
ajp-io: That is the case, you're correct. I was speaking to the fact that these instructions are related to "test your changes" and "save and promote the release". In that case, this is a new release, and not one that existed already.

So your comments are right. I just felt that this section was about a new release that you're creating, not an old one you're revisiting.

docs/vendor/packaging-private-images.md
ajp-io: Looks good.

docs/vendor/packaging-private-images.md
paigecalvert: Edited. Decided to leave off the auto-build piece because it just got sort of wordy, and I figure it's obvious whether or not the thing was built. Also added a step for "download the bundle", which I think further helps clarify this

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,78 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+
+You can enable or disable this air gap bundle format using the **Enable new air gap bundle** toggle in the settings for any channel in the vendor portal.
+
+When you enable **Enable new air gap bundle** on a channel, all air gap bundles that you build or rebuild on that channel use the updated air gap bundle format.
+
+If users on a version of the app manager earlier than v1.82.0 attempt to install or upgrade an application with an air gap bundle that uses the **Enable new air gap bundle** format, then the admin console displays an error message when they attempt to upload the bundle.
+
+To enable the new air gap bundle format on a channel:
+
+1. In the Replicated [vendor portal](https://vendor.replicated.com/channels), go to **Channels** and click the **edit** icon in the top right of the channel where you want to use the new air gap bundle format.
```
ajp-io, line 219: Does edit need to be bold given that it's not a button that actually says edit? I don't know the policy for that situation, where the button is an icon.

ajp-io (approved):
Approving with one question.

docs/vendor/packaging-private-images.md
paigecalvert: That's a good catch; I had just copied this line from another topic, but you're right that that doesn't really need to be bold

paigecalvert:
@ajp-io thanks for the review! Jonquil is going to give this one pass when she gets a moment, and then it should be good to go.

Just checking: do we still want to hold off on merging this? If it's easier, we can just plan to let you merge it whenever it's ready to be live

ajp-io:
Yeah, I can merge it once we remove the feature flag and GA this. Thanks for getting these ready in advance though.

Is that helpful, or is it better to wait until we're closer to GA before putting together docs like these? I suppose we could have merge conflicts that need to be resolved, but that shouldn't be too common or hard.

paigecalvert:
Yeah, doing it in advance works! Always great not to have to rush. (Especially when we are talking a scale of days/weeks and not months, the merge conflicts shouldn't be a big deal)

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,78 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
 :::
 
 
-## Support for Image Tags and Digests
+## Using Image Tags and Digests
 
-The app manager supports image tags for applications in all use cases.
+This section describes using image tags and digests with your application images. It includes information about when image tags and digests are supported, as well how to enable support for image digests in air gap bundles.
```
jonquil2002, line 176: s/ as well how to / and how to

If you don't change to "and how to", then you need to add "as" to make it "as well as".

docs/vendor/packaging-private-images.md
```diff
@@ -171,11 +171,78 @@ For more information about the `additionalNamespaces` attribute, see [Defining A
...
+
+You can enable or disable this air gap bundle format using the **Enable new air gap bundle** toggle in the settings for any channel in the vendor portal.
+
+When you enable **Enable new air gap bundle** on a channel, all air gap bundles that you build or rebuild on that channel use the updated air gap bundle format.
+
+If users on a version of the app manager earlier than v1.82.0 attempt to install or upgrade an application with an air gap bundle that uses the **Enable new air gap bundle** format, then the admin console displays an error message when they attempt to upload the bundle.
+
+To enable the new air gap bundle format on a channel:
+
+1. In the Replicated [vendor portal](https://vendor.replicated.com/channels), go to **Channels** and click the edit icon in the top right of the channel where you want to use the new air gap bundle format.
```
jonquil2002, line 219: Style question that we haven't decided on. Do we want to get away from using "click" and just use generic terms, like "go to", "select" etc.? I haven't used "go to" much before, but this is what Microsoft UI styles say: https://docs.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui

jonquil2002 (approved):
One grammatical change is needed, but approving anyway.

docs/vendor/packaging-private-images.md
paigecalvert: Oh yeah, we can definitely say click here. I just use "go to" instead of "navigate" usually. But realistically "go to" should be used only if I"m saying "go to the Channels page". Since it's bold, this should say definitely say "click"

docs/vendor/packaging-private-images.md
paigecalvert: done!

docs/vendor/packaging-private-images.md
paigecalvert: Ended up changing this to "go to the Channels page" to avoid redundancy on the word "click" in the same sentence

paigecalvert: (and reviewing the Microsoft guide, looks like that's in line with how they recommend using "go to"? Maybe that's where I picked up this habit 😄 )

jonquil2002 (approved):
lgtm

ajp-io:
@paigecalvert This looks accurate. I made one change to update the toggle name.

One addition is that we're defaulting the toggle to on for newly created teams. Newly created teams are teams created after we ship the change that does this. So I guess we can tie it to the date on which this is released. So for new teams, the three default channels (stable, beta, unstable) will have the toggle enabled by default. If they try to create a new channel, it'll be created with the toggle enabled by default.

So it's a similar situation to semantic versioning here (see date in top section): https://docs.replicated.com/vendor/releases-semantic-versioning

paigecalvert:
@ajp-io sounds good. Thanks for making that update.

Thinking about the default on/off details, maybe we can just specify in the release notes that it defaults to off for any existing teams.

And for the docs themselves, we could keep the assumption that the content applies to the latest version, explain that this setting is the default, and include info about how "to enable/disable this air gap bundle setting, click X". 

Looking at the semver topic again, it's kinda clunky and probably not sustainable since we don't have versioned docs. I'm thinking we can be leaning more on the release notes for migration/upgrade details that are version-specific. What do you think?

ajp-io:
@paigecalvert I think that sounds good. Feel free to make an edit to add the default info, but I'm fine with leaning on other things for migration stuff.

ajp-io (approved):

Issue #518, opened 2022-08-10
Docs feedback on snapshots-backup-hooks


MrSimonEmms:
URL: https://docs.replicated.com/vendor/snapshots-backup-hooks
Feedback details:

Adding the [restore hooks](https://velero.io/docs/v1.9/restore-hooks) to the example would be extremely useful too

paigecalvert:
Thank you @MrSimonEmms! I believe that we do have a story for this in our backlog. I will report back when we plan to pick it up.

paigecalvert:
Okay, found that story that I mentioned. I am adding it to our queue for next week. Thanks for the help getting it prioritized!

paigecalvert:
Hi @MrSimonEmms, I wanted to let you know that we've updated the topic to add a restore hook example & some additional high level info about restore hooks as well. Thank you again for providing this feedback; we appreciate hearing from you!

PR #558, opened 2022-08-24
Adds information about Velero restore hooks

paigecalvert:
Story: https://app.shortcut.com/replicated/story/29970/add-velero-restore-hook-docs

Related to https://github.com/replicatedhq/replicated-docs/issues/518

Preview: https://deploy-preview-558--replicated-docs.netlify.app/vendor/snapshots-hooks

**NOTE**: Need to update the filename accordingly and get a new redirect in place before this can merge.


docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
+# Configuring Backup and Restore Hooks
+
+For many application workloads, additional processing or scripts must be run before or after creating a backup to prepare the system for a backup. Many application workloads also require additional processing or scripts to run during or after the restore process.
+
+Velero supports this through backup hooks and restore hooks.
+
+Some common examples of how to use a hook to create successful backups are:
+- Run `pg_dump` to export a postgres database prior to backup
+- Lock a file before running a backup, and unlock immediately after
+- Delete tmp files that should not be backed up
```
jonquil2002, line 10: s/ tmp / TMP

We decided on uppercase when referring file extensions (it's in the Style Guide).

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+:::
+
+   <table>
+     <tr>
+       <th>Annotation</th>
+       <th>Description</th>
+     </tr>
+     <tr>
+       <td><code>backup.velero.io/backup-volumes</code></td>
+       <td>a comma separated list of volumes from the Pod to include in the backup. The primary data volume is not included in this field.</td>
```
jonquil2002, line 78: Need to start sentence with uppercase.

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+     </tr>
+     <tr>
+       <td><code>pre.hook.backup.velero.io/command</code></td>
+       <td>A stringified JSON array containing the pre backup hook command.
```
jonquil2002, line 82: Looking at the Velero docs, it seems that "pre backup hook command" should be changed to "pre hook backup command" in this description.

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+       This command is a <code>pg_dump</code> from the running database to the backup volume.</td>
+     </tr>
+     <tr>
+       <td><code>pre.hook.backup.velero.io/timeout</code></td>
+       <td>a duration for the maximum time to let this script run for</td>
```
jonquil2002, line 87: Needs to start with uppercase "A" and needs a period at the end.

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+     </tr>
+     <tr>
+       <td><code>post.hook.restore.velero.io/command</code></td>
+       <td>A Velero exec restore hook that runs a script to checks if the database file exists, and restore only if it exists. Then, the script deletes the file when the operation is complete.</td>
```
jonquil2002, line 91: s/ that runs a script to checks if the database file exists, and restore only if it exists. / that runs a script to check if the database file exists, and restores only if it exists.

s/ Then, the script deletes the file when the operation is complete. / The script deletes the file after the operation is complete.

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+     </tr>
+   </table>
+
+* `spec.master.extraVolumes`: A new volume that is injected into the postgres Pod. The new volume is an empty volume, stored in memory. Meaning, it does not require a PVC or storage.
+We mount the volume into the `/scratch` directory of the master Pod, and use it as a destination when running `pg_dump` from the backup hook described above. This is the only volume that is backed up.
```
jonquil2002, line 96: Maybe change it to refer to Replicated (the product) instead of "we'" - and then change the verb tenses accordingly. But it's fine as is too, I just prefer not to say "we".

paigecalvert: Ugh, yeah, missed this one. In this case, the "we" is actually referring to "the person who wrote that example YAML, and the reader". I edited this to say "The `extraVolumeMounts` field mounts the volume into the..." to clarify

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+
+  builder: {}
+
+```
+
+The following describes key fields from the example above:
+
+* `spec.exclude`: A common and recommended pattern for applications. On the Replicated admin console Configuration screen, the customer can choose to bring an external Postgres instance instead of running it in-cluster. The Replicated template function in `spec.exclude` evaluates to true when the user selects the external database option on the Configuration screen. This means that the internal Postgres database is not included in the deployment.
+
+* `spec.values.master.podannotations`: Adds annotations to the postgres master PodSpec. Velero backup and restore hooks are included in the PodSpec annotations. The following table describes the annotations:
```
divolgin, line 66: `podAnnotations`, pretty sure these are case sensitive

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+:::
+
+   <table>
+     <tr>
+       <th>Annotation</th>
+       <th>Description</th>
+     </tr>
+     <tr>
+       <td><code>backup.velero.io/backup-volumes</code></td>
+       <td>A comma-separated list of volumes from the Pod to include in the backup. The primary data volume is not included in this field.</td>
```
divolgin, line 78: Could be good to say why primary data volume is not included in this field: "because data will be exported using the backup hook"?

docs/vendor/snapshots-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+      podAnnotations:
+        backup.velero.io/backup-volumes: backup
+        pre.hook.backup.velero.io/command: '["/bin/bash", "-c", "PGPASSWORD=$POSTGRES_PASSWORD pg_dump -U username -d dbname -h 127.0.0.1 > /scratch/backup.sql"]'
+        pre.hook.backup.velero.io/timeout: 3m
+        post.hook.restore.velero.io/command: '["/bin/bash", "-c", "[ -f \"/scratch/backup.sql\" ] && PGPASSWORD=$POSTGRES_PASSWORD psql -U username -h 127.0.0.1 -d dbname -f /scratch/backup.sql && rm -f /scratch/backup.sql;"]'
+
+      extraVolumes:
+        - name: backup
+          emptyDir:
+            medium: Memory
```
divolgin, line 46: How about removing the medium line.  Storing db dump in memory may not be practical for applications running at production scale.  Someone might copy this example verbatim and end up with an evicted pod.

docs/vendor/snapshots-hooks.md
paigecalvert: Great suggestion! Updated.

divolgin (approved):

jonquil2002 (approved):
lgtm

PR #593, opened 2022-08-29
Adds conceptual info on how to support all install methods

paigecalvert:
https://app.shortcut.com/replicated/story/56654/update-the-helm-install-about-section-with-details-on-using-the-helmchart-cr-to-support-both-helm-and-kots-installs

https://deploy-preview-593--replicated-docs.netlify.app/vendor/helm-install#about-supporting-helm-cli-installations


docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
 For information about how to package an application in the Replicated vendor portal using an existing Helm chart, see [Adding Helm Charts to a Release](helm-release).
 
+## About Supporting helm CLI installations
+
 Using Helm charts to create a release in the vendor portal allows you to package your application one time and support the following application installation methods:
-* Installing on a cluster provisioned by the Replicated Kubernetes installer (kURL)
-* Installing on an existing cluster using the kots CLI
-* Installing on an existing cluster using the helm CLI
+
+* **KOTS Install**: Installing on an existing cluster using the kots CLI or the admin console
```
jonquil2002, line 19: For descriptions, I think it's better to start with a verb, e.g. "Installs" instead of "Installing". How do you feel about that? We don't seem to have anything in our style guide yet about term list definitions, so we ought to officially decide on something. But this is a standard that has always existed wherever I've worked, so I've been using it throughout my doc changes.

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
+* **Embedded Cluster**: Installing on a cluster provisioned by the Replicated Kubernetes installer (kURL)
+* **Helm Install**: Installing on an existing cluster using the helm CLI
+
+To support all installation methods from a single application release, you must be able to conditionally include or exclude values and resources in the release depending on the installation method. This is because Replicated delivers built-in functionality when users install with the KOTS Install or Embedded Cluster methods that cannot be delivered automatically when users install with the helm CLI. So, you must explicitly include the desired functionality for helm CLI installations.
```
jonquil2002, line 23: s/ This is because Replicated delivers built-in functionality when users install with the KOTS Install or Embedded Cluster methods that cannot be delivered automatically when users install with the helm CLI. / This is because Replicated delivers built-in functionality with the KOTS Install or Embedded Cluster methods that cannot be delivered automatically with the helm CLI method.


It's a little hard to follow this sentence, and I think if you extract the wording "when users install" and just focus on the method, I think the point becomes clearer.

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
+
+For example, when users install with the KOTS Install or Embedded Cluster installation methods, Replicated automatically delivers the admin console with the application. To deliver the admin console with your application for helm CLI installations, you must include the `admin-console` Helm chart as a dependency. Then, to prevent Replicated from attempting to deploy this `admin-console` dependency for KOTS Install or Embedded Cluster installations, you must conditionally exclude the `admin-console` Helm chart from these installations.
```
jonquil2002, line 25: s/ For example, when users install with the KOTS Install or Embedded Cluster installation methods... / For example, with the KOTS Install or Embedded Cluster installation methods

or

For example, Replicated automatically delivers...with the KOTS Install or Embedded Cluster installation methods.

jonquil2002 (approved):
lgtm, with suggestions and questions for you.

paigecalvert:
Thank you! All great notes. And yeah, I was struggling with how to refer to all the different types, so I defaulted to what was displayed in the vendor portal for now. I think that should likely change. (Maybe I can clarify at the top that KOTS Install = kots cli or admin console, which both use the app manager. Then just say "app manager or k8s installer installations" later on)

docs/vendor/helm-install.md
paigecalvert: Yes, you're right, the "ing" is a no-go for sure

docs/vendor/helm-install.md
paigecalvert: Nice, good call. More consistent structure that way

docs/vendor/helm-install.md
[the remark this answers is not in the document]
paigecalvert: I originally wrote a version of this in a separate comment below, but going to restate it here as well so as to keep the convo in this thread:

I used the terms KOTS Install, Embedded Cluster, and Helm Install throughout just because that is exactly how they appear in the vendor portal. I found this to be the most fool-proof way of referring to these methods just because I wasn't sure that saying something like "using the app manager to install..." to refer to the non-Helm methods would be clear enough.

Particularly because KOTS Install uses the term "KOTS" and the description of Embedded Install mentions how it uses the Kubernetes installer, it didn't feel that it would be obvious to users that "the app manager" refers to the KOTS Install and Embedded Cluster methods. Also, I'm not sure it's entirely accurate to suggest that the Helm Install method does not involve the app manager at all, since it does process some fields in the HelmChart custom resource. So maybe there's less chance of users getting confused if we use the terms exactly as they appear in the UI?

Another option might be too beef up the descriptions of each method in the bulleted list at the top to more fully explain what's going on so that we can refer to them differently throughout the topic.

Curious what both you and @ajp-io think of this one

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
 Using Helm charts to create a release in the vendor portal allows you to package your application one time and support the following application installation methods:
-* Installing on a cluster provisioned by the Replicated Kubernetes installer (kURL)
-* Installing on an existing cluster using the kots CLI
-* Installing on an existing cluster using the helm CLI
+
+* **KOTS Install**: Install on an existing cluster using the kots CLI or the admin console UI
+* **Embedded Cluster**: Install on a cluster provisioned by the Replicated Kubernetes installer (kURL)
+* **Helm Install**: Install on an existing cluster using the helm CLI
+
+To support all installation methods from a single application release, you must be able to conditionally include or exclude values and resources in the release depending on the installation method. This is because Replicated delivers built-in functionality with the KOTS Install or Embedded Cluster methods that cannot be delivered automatically with the helm CLI method. So, you must explicitly include the desired functionality for helm CLI installations.
```
paigecalvert, line 23: This should be rewritten to better convey that some things are included by default with kots/embedded that are not included by default with helm:
>This is because Replicated delivers built-in functionality with the KOTS Install or Embedded Cluster methods that cannot be delivered automatically with the helm CLI method. So, you must explicitly include the desired functionality for helm CLI installations.

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
+
+For example, Replicated automatically delivers the admin console with the KOTS Install or Embedded Cluster installation methods. To deliver the admin console with your application for helm CLI installations, you must include the `admin-console` Helm chart as a dependency. Then, to prevent Replicated from attempting to deploy this `admin-console` dependency for KOTS Install or Embedded Cluster installations, you must conditionally exclude the `admin-console` Helm chart from these installations.
+
+To control which values and resources are included in your application release for each installation method, you create a Replicated HelmChart custom resource manifest file in the release. The HelmChart custom resource includes several fields that allow Replicated to process and deploy Helm charts. For more information, see [HelmChart](/reference/custom-resource-helmchart) in the _References_ section.
```
paigecalvert, line 27: Maybe "using a HelmChart custom resource" instead of "creating", which might convey that you are creating the resource specifically for this purpose, although you likely already have it and you are just leveraging it

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
+
+The `values` field in the HelmChart custom resource lets you edit the Helm chart `values.yaml` file. A common use case for the HelmChart `values` field is to use Replicated template functions to map user-provided values from the Replicated admin console Configuration screen to the Helm chart `values.yaml` field.
```
paigecalvert, line 29: Could make generic this description to explain that you can use template functions to a map user-specific/dynamic values such as config inputs or license fields.

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
+
+Replicated injects any templated values from the HelmChart custom resource `values` field into the Helm chart `values.yaml` file, but does not inject any static, or _hardcoded_, values into the `values.yaml` file. This means that any static values that you include in the `values` field are used for KOTS Install and Embedded Cluster installations, and ignored for Helm Install installations.
+
+For example, if you include the `admin-console` Helm chart as a conditional dependency in your application `Chart.yaml` file, you can write a static key value pair in the `values` section of the HelmChart custom resource to make the condition evaluate to false only for KOTS Install and Embedded Cluster installations. This would prevent Replicated from attempting to deploy the `admin-console` Helm chart for KOTS Install or Embedded Cluster installations.
```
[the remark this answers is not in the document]
paigecalvert: Or what if we removed the examples? Maybe we can keep it more general, then point down to the procedures.

We need to explain the concept of the values.yaml / values field. Then say, for specific examples, go to the sections below

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,27 @@ To support installations with the helm CLI, you must have an application that co
...
+
+You use this method of including static values in the HelmChart custom resource manifest file in any case where it is required to conditionally include or exclude values or resources depending on the installation method. For more information about how to use this method to deliver Replicated functionality for helm CLI installations, see [Delivering the Admin Console with your Application](#deliver-admin-console) and [Using Private Registries](#private-images) below.
```
paigecalvert, line 35: The first sentence here might be the lead/summary that can be re-written/simplified

docs/vendor/helm-install.md
```diff
@@ -12,10 +12,25 @@ To support installations with the helm CLI, you must have an application that co
...
+
+* **KOTS Install**: Install on an existing cluster using the kots CLI or the admin console UI
+* **Embedded Cluster**: Install on a cluster provisioned by the Replicated Kubernetes installer
+* **Helm Install**: Install on an existing cluster using the helm CLI
+
+To support each of these installations methods from a single release, you use the `values` field of the Replicated HelmChart custom resource. The `values` field supports both templated values that are generated by Replicated template functions, as well as static, or _hardcoded_, values.
+
+Replicated injects any templated values from the HelmChart custom resource `values` field into the Helm chart `values.yaml` file. For example, you can use template functions in the HelmChart custom resource to map user-specific values such as license fields or configuration values to the `values.yaml` file.
+
+Replicated does _not_ inject any static values from the HelmChart custom resource `values` field into the Helm chart `values.yaml` file. Any static values in the `values` field affect only KOTS Install and Embedded Cluster installations, and are ignored for Helm Install installations.
```
ajp-io, line 27: In the second paragraph, we might want to clarify that we don't inject static values for Helm installs. I know the second sentence kind of clarifies that, but the first sentence doesn't feel accurate to me unless it's clear that it's only talking about the Helm install method.

ajp-io (commented):
This is really good. Much more readable than the first iteration.

docs/vendor/helm-install.md
paigecalvert: Makes sense to me--I added the following to clarify it's talking about helm install specifically:

Screen Shot 2022-09-14 at 11 29 06 AM

ajp-io: Nice, I think that top section is great. The only other thing I can think of… And this could be done separately if we want to merge this. Would it make sense to add a sentence in the admin-console Helm chart and private images sections when we leverage this ability?

Like [here](https://deploy-preview-593--replicated-docs.netlify.app/vendor/helm-install#add-the-admin-console-chart), could we say in step 2 that this works because the Helm install will read this `true` condition from the values file? And then in step 5 mention that this part works because the KOTS install will overwrite `true` with the `false` from the HelmChart custom resource? Let me know if you think that's helpful or not, but it brings it full circle a bit. I don't know if we'd want to link up to this new section there or not (maybe that's too circular).

Not saying this is necessary. Let me know what you think. Just looking for it to be clear.

paigecalvert: Totally, I see what you're saying. Let me take a crack at that and will ping you again shortly

paigecalvert: Okay! Sorry for the delay. Here are the additions to step 2 and 5 in the admin-console procedure:
Screen Shot 2022-09-15 at 11 19 17 AM
Screen Shot 2022-09-15 at 11 19 27 AM


Private image additions coming in a subsequent comment...

paigecalvert: Additions to steps 1, 2, 5 in the private images procedure:

Screen Shot 2022-09-15 at 11 19 40 AM

Screen Shot 2022-09-15 at 11 19 57 AM


Screen Shot 2022-09-15 at 11 20 09 AM

ajp-io: That looks great. I'm going to approve based on those screenshots. Didn't read the actual files.

ajp-io (approved):

PR #594, opened 2022-08-29
57161: Added a list of supported `kots velero` commands

jonquil2002:
SC: https://app.shortcut.com/replicated/story/57161/add-list-of-kots-cli-configure-commands-to-the-main-velero-section

Preview: https://deploy-preview-594--replicated-docs.netlify.app/reference/kots-cli-velero-index

### DIFF
--- a/docs/reference/kots-cli-velero-index.md
+++ b/docs/reference/kots-cli-velero-index.md
@@ -10,9 +10,7 @@ kubectl kots velero [command] [global flags]
 
 This command supports all [global flags](kots-cli-global-flags).
 
-#### Available Commands
-
-The following `kots velero` commands are supported. For more information about these commands, see the listing for a specific command:
+The following `kots velero` commands are supported:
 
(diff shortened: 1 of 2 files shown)


paigecalvert:
@jonquil2002 great call on making this addition. It really makes this landing page so much more useful. 

Just a couple suggestions come to mind:
- What do you think about removing the Available Commands heading? I was looking at other CLI topics, and it looks like they don't use a "Available Flags" heading above the table of flags, but just roll right into it. I was thinking we could get away with just jumping into "The following commands are supported..."
- I was also thinking we could probably remove "For more information about these commands, see the listing for a specific command:" just because the way you've formatted it it's very clear that you can click on those links for more info (and they should also be able to see those same commands mirrored in the TOC to the left). Removing this line would also allow the colon to go right after the sentence "The following commands are supported:"

Anyway, pretty minor and up to you. Looks great!

paigecalvert (approved):

jonquil2002:
@paigecalvert I debated about those two items, so glad to change as you suggested. Ready for final review/merge when you have a chance.

paigecalvert (approved):

PR #598, opened 2022-08-31
56494: Valid Air Gap URL

jonquil2002:
SC[:](https://app.shortcut.com/replicated/story/56494/correct-the-url-for-the-air-gap-bundle-in-installing-in-an-air-gapped-environment)

### DIFF
--- a/docs/vendor/tutorial-installing-air-gap.md
+++ b/docs/vendor/tutorial-installing-air-gap.md
@@ -71,13 +71,24 @@ To install the Kubernetes installer bundle:
       curl -LS https://k8s.kurl.sh/bundle/$REPLICATED_APP.tar.gz -o $REPLICATED_APP.tar.gz
       ```
 
-    * To download the Kubernetes installer bundle for all other channels, use the suffix `-<channel_name>`:
+    * To download the Kubernetes installer bundle for all other channels:
 
-      ```shell
-      export REPLICATED_APP=<app_slug>
-      export REPLICATED_CHANNEL=<channel_name>
-      curl -LS https://k8s.kurl.sh/bundle/$REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz -o $REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz
-        ```
+        - Run the following command to get the air gap URL:
+
+          ```shell
+          replicated channel inspect CHANNEL_NAME
+          ```
+          Replace `CHANNEL_NAME` with the exact name of the channel, which can include uppercase letters, such as `Unstable`.
+
+          The output shows valid URLs with the channel name in lowercase. Valid URLs do not support uppercase in the channel name.
+
+        - Run the following commands to download the bundle, replacing the URL syntax below with the valid air gap URL from the previous step:
+
+          ```shell
+          export REPLICATED_APP=<app_slug>
+          export REPLICATED_CHANNEL=<channel_name>
+          curl -LS https://k8s.kurl.sh/bundle/$REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz -o $REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz
+          ```
 
 1. Run `scp` for the `.tar.gz` file from the jumpbox to the air gapped node:
 


docs/vendor/tutorial-installing-air-gap.md
```diff
@@ -71,13 +71,24 @@ To install the Kubernetes installer bundle:
...
-      curl -LS https://k8s.kurl.sh/bundle/$REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz -o $REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz
-        ```
+        - Run the following command to get the air gap URL:
+
+          ```shell
+          replicated channel inspect CHANNEL_NAME
+          ```
+          Replace `CHANNEL_NAME` with the exact name of the channel, which can include uppercase letters, such as `Unstable`.
+
+          The output shows valid URLs with the channel name in lowercase. Valid URLs do not support uppercase in the channel name.
```
paigecalvert, line 83: Recommend specifically calling out that the URL to use is the one that appears under the "AIRGAP:" heading in the CLI output. (There are a few different URLs that are in the output, so good help to be specific)

As part of this, could also provide an example of the output (see my comment in the GitHub issues for the example)

docs/vendor/tutorial-installing-air-gap.md
```diff
@@ -71,13 +71,24 @@ To install the Kubernetes installer bundle:
       curl -LS https://k8s.kurl.sh/bundle/$REPLICATED_APP.tar.gz -o $REPLICATED_APP.tar.gz
       ```
 
-    * To download the Kubernetes installer bundle for all other channels, use the suffix `-<channel_name>`:
+    * To download the Kubernetes installer bundle for all other channels:
```
paigecalvert, line 74: I believe the two steps below this should be numbered, since they are to be in done in sequence

docs/vendor/tutorial-installing-air-gap.md
```diff
@@ -71,13 +71,24 @@ To install the Kubernetes installer bundle:
...
 
-      ```shell
-      export REPLICATED_APP=<app_slug>
-      export REPLICATED_CHANNEL=<channel_name>
-      curl -LS https://k8s.kurl.sh/bundle/$REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz -o $REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz
-        ```
+        - Run the following command to get the air gap URL:
```
paigecalvert, line 76: We might need a first step to download the replicated CLI

docs/vendor/tutorial-installing-air-gap.md
```diff
@@ -71,13 +71,24 @@ To install the Kubernetes installer bundle:
...
+
+          The output shows valid URLs with the channel name in lowercase. Valid URLs do not support uppercase in the channel name.
+
+        - Run the following commands to download the bundle, replacing the URL syntax below with the valid air gap URL from the previous step:
+
+          ```shell
+          export REPLICATED_APP=<app_slug>
+          export REPLICATED_CHANNEL=<channel_name>
+          curl -LS https://k8s.kurl.sh/bundle/$REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz -o $REPLICATED_APP-$REPLICATED_CHANNEL.tar.gz
+          ```
```
paigecalvert, line 91: Since we are going with the method of grabbing the URL from the output, this step of exporting these environment variables is no longer necessary. The user just needs to run the URL that was copied from the previous step.

So, I think this could probably just say "run the URL that you copied" with an example

docs/vendor/tutorial-installing-air-gap.md
paigecalvert: >Valid URLs do not support uppercase in the channel name.

I think this line might be lacking context now that we are telling them to just copy the URL in the output of the channel-inspect command and run it as is. Since we are not asking them to type out the channel name in an environment var anymore in order to construct the URL manually, there is not really a reason that they should have any uppercase letters in the command

If we did want to keep some instructions about how to construct the URL manually if you don't want to download the replicated CLI (though, encouraging people to use the CLI seems like a good move), then I think this would make sense to call out explicitly. Otherwise, can probably remove to avoid people wondering why that's there

paigecalvert (dismissed):
Thank you @jonquil2002 ! Added suggestions

docs/vendor/tutorial-installing-air-gap.md
jonquil2002: My understanding is that this is an either/or step, because they're only going to use the download bundle from the channel that they created the release in.

docs/vendor/tutorial-installing-air-gap.md
jonquil2002: I agree that we should leave out the manual instructions. Thanks for the input, I was on the fence with some of this. It's great to have a second opinion.

docs/vendor/tutorial-installing-air-gap.md
jonquil2002: Oops, just realized you;re talking about the substeps. Yes, you are correct!

docs/vendor/tutorial-installing-air-gap.md
jonquil2002: Good call.

jonquil2002:
@dexhorthy @paigecalvert Please review. Thanks.

docs/vendor/tutorial-installing-air-gap.md
paigecalvert: @jonquil2002 I think we still need to remove the `export` commands and the curl command with the placeholders since they are copying the command now rather than constructing it.

paigecalvert (commented):
@jonquil2002 just noticed that the export commands were still there!

docs/vendor/tutorial-installing-air-gap.md
dexhorthy: agree- running the export and curl commands will not work if the channel contains special characters or capital letters

jonquil2002: Thanks, folks. Is it correct now @dexhorthy ?

paigecalvert (approved):

PR #630, opened 2022-09-14
Updates to admin console icon recommendations

GraysonNull:
We no longer show app icons in the the circle so not all of these recommendations are relevant anymore. I'm not sure if just want to remove them or keep some of the recs and and update all of the screenshots


paigecalvert (approved):
Thanks @GraysonNull -- we can delete for now, and should probably add a new screenshot as you mention since it's a task that would be helpful to have a visual for. I can make a separate story for that.

is the logo change part of the admin console branding? is it something vendors can control? basically just wondering if we need to share any new info about how to define the size and shape of the image, or something
Are the logo changes part of

PR #632, opened 2022-09-14
55332: Add Host Collectors and Analyzer Info to Vendor Docs

jonquil2002:
SC: https://app.shortcut.com/replicated/story/55332/add-host-collectors-for-support-bundles

Previews: 

- https://deploy-preview-632--replicated-docs.netlify.app/vendor/preflight-support-bundle-creating#about-host-collectors-and-analyzers-in-support-bundles
- https://deploy-preview-632--replicated-docs.netlify.app/vendor/support-host-support-bundles
- https://deploy-preview-632--replicated-docs.netlify.app/enterprise/troubleshooting-an-app#run-host-collectors-and-analyzers


docs/vendor/preflight-support-bundle-creating.md
```diff
@@ -268,3 +268,80 @@ To customize a support bundle:
 
     - If you have the Support Bundle Upload Enabled license entitlement, click **Send bundle to vendor**. You can also open the TAR file to review the files. For more information about license entitlements, see [Create a Customer](releases-creating-customer#create-a-customer).
     - Download the `support-bundle.tar.gz` file. Copy the file to the Troubleshoot tab in the [vendor portal](https://vendor.replicated.com) to see the analysis and use the file inspector.
+
+## About Host Collectors and Analyzers
+
+Host collectors and analyzers are designed to collect information from Kubernets installer-created clusters (embedded clusters) that is not available from in-cluster collectors. Host collectors gather information directly from the host they are run on and do not have Kubernetes as a dependency.
```
diamonwiggins, line 274: typo: Kubernetes*

diamonwiggins (dismissed):
looks good. just a single typo

docs/vendor/preflight-support-bundle-creating.md
```diff
@@ -268,3 +268,80 @@ To customize a support bundle:
...
+            outcomes:
+              - fail:
+                  when: "< 4G"
+              - pass:
+                  message: The system has at least 8G of memory
+    ```
+
+## Run Host Collectors and Analyzers
+
+You run host collectors and analyzers to generate a support bundle to help with troubleshooting or to gather information about the environment.
```
paigecalvert, line 325: Is the "you" enterprise customers here?

docs/vendor/preflight-support-bundle-creating.md
```diff
@@ -268,3 +268,80 @@ To customize a support bundle:
...
+
+You can gather information about the environment, such as CPU, memory, available block devices, and the operating system. Host collectors can also be used for testing network connectivity and gathering the output of provided commands.
+
+This information is useful when you need to debug a Kubernetes installer cluster that is offline, troubleshoot a Kubernetes installer that failed before the control plane was initialized, or if you need to collect and analyze information that is not available with in-cluster collectors.
+
+## Define Custom Host Collectors and Analyzers
+
+Create a separate support bundle manifest file to specify host collectors and analyzers because host collectors are intended to run directly on the host using the CLI and not with KOTS. If host collectors run from KOTS, they are unlikely to produce the desired results because they run in the context of the Kotsadm pod.
```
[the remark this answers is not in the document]
paigecalvert: I also got a little tripped up on "because host collectors are intended to run directly on the host using the CLI and not with KOTS. If host collectors run from KOTS,..." I might have been missing the implied information, which I think is:
- host collectors need to run on the host where the `curl -L https://github.com/replicatedhq/troubleshoot/releases/download/VERSION/support-bundle_linux_amd64.tar.gz | tar xzvf -` command is run? (wasn't sure which CLI this was referring to)
- support bundles normally would run on the Kotsadm pod so they can collect info about the cluster and the app on the cluster, so these host bundles are different. (I think it was maybe just that "run from KOTS" and "with KOTS" took me a sec to understand)

paigecalvert (dismissed):
Good stuff, just a few suggestions

docs/vendor/preflight-support-bundle-creating.md
ajp-io, line 282: This is the name of the Pod, and it's lowercase. We're capitalizing Kubernetes resource types now, right? Like Pod?
```suggestion
Create a separate support bundle manifest file to specify host collectors and analyzers because host collectors are intended to run directly on the host using the CLI and not with KOTS. If host collectors run from KOTS, they are unlikely to produce the desired results because they run in the context of the kotsadm Pod.
```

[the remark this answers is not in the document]
ajp-io: It might be clearer to say "If KOTS runs host collectors, they are unlikely…" Rather than "run from KOTS" or "with KOTS".

The support-bundle CLI is the CLI that should be used to run these directly on the host (rather than in a kotsadm Pod running on the host).

docs/vendor/preflight-support-bundle-creating.md
```diff
@@ -268,3 +268,80 @@ To customize a support bundle:
...
+                  when: "< 4G"
+              - pass:
+                  message: The system has at least 8G of memory
+    ```
+
+## Run Host Collectors and Analyzers
+
+You run host collectors and analyzers to generate a support bundle to help with troubleshooting or to gather information about the environment.
+
+Root access is not required to run any of the host collectors. However, depending on what you want to collect, you must run the binary with elevated permissions. For example, if you run the `filesystemPerformance` host collector against `/var/lib/etcd` and the user running the binary does not have permissions on this directory, the collection process fails.
```
ajp-io, line 327: The "must" in the second sentence threw me a bit. The first sentence says you don't need root access, then the second says that (depending on what you're collecting) you _must_ run with elevated permissions. Softer wording like "you might need to run with the binary with elevated permissions" might be more accurate.

docs/vendor/preflight-support-bundle-creating.md
```diff
@@ -268,3 +268,80 @@ To customize a support bundle:
...
+  Replace `VERSION` with `v` and the version number. For example, `v0.40.0`.
+
+  :::note
+  There is no method in Troubleshoot to run host collectors on remote nodes. If you have a multi-node Kubernetes cluster, you must run the support bundle binary on each node and generate a bundle for each.
+  :::
+
+1. Run the following command on the host to generate a support bundle:
+
+  ```
+  ./support-bundle --interactive=false support-bundle.yaml
```
ajp-io, line 346: This command implies that the `support-bundle.yaml` file will be in the home directory (the one that you are in after you SSH onto the node). Is that true? (I assume you won't know, but it might be a good question for someone who has used these. Will the `support-bundle.yaml` be in that top-level directory, or will there be a particular path required for it?)

docs/vendor/preflight-support-bundle-creating.md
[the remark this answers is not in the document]
jonquil2002: I think I resolved this. Thanks!

jonquil2002: Yep, good catch.

docs/vendor/preflight-support-bundle-creating.md
jonquil2002: Good question. 

@diamon - When host collectors are run, is it the enterprise user that is running them? I would think so. Which means that this Run Host Collectors/Analyzers procedure belongs in the Enterprise part of the documentation,  and the "Define Host Collectors" part belongs in the vendor section. Can you confirm who runs this CR on the host please?

docs/vendor/preflight-support-bundle-creating.md
jonquil2002: @diamonwiggins  Can you answer Alex's question here?

docs/vendor/preflight-support-bundle-creating.md
diamonwiggins: For Host Collectors, they are likely to solely be used by someone in the support organization of our Vendors, but not actually defined anywhere in the Vendor Portal, so I think Enterprise makes the most sense here? I'm not sure? docs/enterprise/Troubleshooting makes sense to me.

This is confusing because `kind: HostPreflight` which also leverage Host Collectors and Analyzers would make sense in the Vendor section because those are generally configured by Vendors along with their kURL specs.

docs/vendor/preflight-support-bundle-creating.md
diamonwiggins: That file doesn't exist by default. In the docs here -https://troubleshoot.sh/docs/host-collect-analyze/overview/, we have a code snippet that's meant to be use to create a file, and then you run the command with the file you generate. If we wanted to make things simpler here, we could have the command be:

`./support-bundle --interactive=false https://raw.githubusercontent.com/replicatedhq/troubleshoot-specs/main/host/cluster-down.yaml`

And also caveat that you'd need to download the file prior and copy it to the machine if it's airgap

jonquil2002: @ajp-io @diamon But from the previous question further up, it seems like the vendor is expected to create the file and give it to the customer. I didn't grok that earlier. So I think a few changes are needed:

1. The support-bundle.yaml file creation should likely stay in the vendor section.
2. The "Run Host Collectors/Analyzers" section should go in the enterprise docs.
3. The enterprise customer would use the path for wherever they put the `support-bundle.yaml` file that they get from the vendor and run the command using whatever the path is. So making the command look more like this: `./support-bundle --interactive=false /PATH/TO/support-bundle.yaml`

Yes?

diamonwiggins: That mostly makes sense to me. The only thing I'm thinking with regards to:

> The support-bundle.yaml file creation should likely stay in the vendor section

I think any mention of support bundles which contain host collectors in the Vendor section may confuse users. Because they aren't meant to be run by KOTS, you wouldn't include them in your KOTS app that you build via the Vendor Portal.

cc: @chris-sanders For Vendors who would want to build their own support bundle spec containing host collectors, where do you think is the best place to land those instructions in the docs, but not confuse users that its something that shouldn't be vendored?

chris-sanders: Yup I agree HostCollectors shouldn't be under a subheading of "packaging your app" because they don't belong or even work if delivered in that way.

I think it belongs in a different section that talks about collecting bundles in different ways. Which is better described here https://app.shortcut.com/replicated/story/58743/update-support-request-documentation-to-emphasize-providing-a-support-bundle instead of the above.

If we had a section of the docs that was about collecting bundles rather than a section about packaging your app then it would be appropriate to have a sub-section on HostCollectors. We should reference our own and we can tell vendors how to maintain or write them as well.

jonquil2002:
@ajp-io  Ready for your nits :)  I made changes to the terminology and headings that we talked about, let's see if they really fit. The one that I didn't change is "host collectors YAML file" in the enterprise section, which we are stuck on. Hoping you can come up with a better name. Thanks for all of your help.

docs/vendor/support-host-collectors.md
```diff
@@ -0,0 +1,59 @@
+# Configuring Host Collectors and Analyzers
```
ajp-io, line 1: How do you feel about calling this "Configuring Host Support Bundles" since you introduced that term elsewhere in the docs?

If so, the line below could also be updated to say, "to help customers generate host support bundles"

docs/vendor/support-host-collectors.md
```diff
@@ -0,0 +1,59 @@
...
+
+You can gather information about the environment, such as CPU, memory, available block devices, and the operating system. Host collectors can also be used for testing network connectivity and gathering the output of provided commands.
+
+This information is useful when you need to debug a Kubernetes installer cluster that is offline, troubleshoot a Kubernetes installer that failed before the control plane was initialized, or if you need to collect and analyze information that is not available with in-cluster collectors.
+
+You create the host collectors file separately from your application package and share the file with customers to run on their hosts. This file is separate  from your application package because host collectors are intended to run directly on the host and not with KOTS. If KOTS runs host collectors, the collectors are unlikely to produce the desired results because they run in the context of the kotsadm Pod. For more information about how customers generate a host collectors support bundle, see [Generate a Host Collectors and Analyzers Bundle](troubleshooting-an-app#run-host-collectors-and-analyzers).
```
ajp-io, line 13: You can keep "package" and not use "release". I just put it as an alternative since we talk about releases usually.

I'm trying to consolidate the language around "host support bundle" here.
```suggestion
You create the host support bundle specification separately from your application release and share the file with customers to run on their hosts. This file is separate from your application release because host collectors are intended to run directly on the host and not with KOTS. If KOTS runs host collectors, the collectors are unlikely to produce the desired results because they run in the context of the kotsadm Pod. For more information about how customers generate a host support bundle, see [Generate a Host Bundle](troubleshooting-an-app#generate-a-host-support-bundle.
```

docs/vendor/support-host-collectors.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Configure Host Collectors and Analyzers
```
ajp-io, line 15: This could potentially stay as is, even if we change the overall page title to "Configuring Host Support Bundles".

docs/vendor/support-host-collectors.md
```diff
@@ -0,0 +1,59 @@
...
+
+Configure a SupportBundle custom resource to specify host collectors and analyzers to gather information from a host.
+
+To define host collectors and analyzers:
+
+1. Create a SupportBundle custom resource manifest file (`kind: SupportBundle`).
+
+1. Configure all of your host collectors and analyzers in one manifest file. You can use the following resources to help create your specification:
+
+    - Access sample specifications in the the Replicated troubleshoot-specs repository, which provides aggregate specifications for supporting your customers. See [troubleshoot-specs](https://github.com/replicatedhq/troubleshoot-specs) in GitHub.
```
ajp-io, line 25: I wonder if we should link to [this](https://github.com/replicatedhq/troubleshoot-specs/tree/main/host), since this is the host support bundle subdirectory. Otherwise they have to find this themselves after they click on that link.

Also the language "provides aggregate specifications" is confusing to me. I'm not sure what "aggregate specifications" means here.

docs/vendor/support-host-collectors.md
```diff
@@ -0,0 +1,59 @@
...
+              - fail:
+                  when: "count < 2"
+                  message: At least 2 CPU cores are required, and 4 CPU cores are recommended
+              - pass:
+                  message: This server has at least 4 CPU cores
+        - memory:
+            checkName: "Amount of Memory"
+            outcomes:
+              - fail:
+                  when: "< 4G"
```
ajp-io, line 55: Could probably add a `message` here like the CPU cores failure has.

docs/vendor/support-host-collectors.md
```diff
@@ -0,0 +1,59 @@
...
+              - pass:
+                  message: The system has at least 8G of memory
+    ```
+1. Share the SupportBundle custom resource file with your customers to run on their host.
```
ajp-io, line 59:
```suggestion
1. Share the SupportBundle custom resource file with your customers to run on their host(s).
```

docs/vendor/support-host-collectors.md
```diff
@@ -0,0 +1,59 @@
+# Configuring Host Collectors and Analyzers
+
+This topic describes how to configure host collectors and analyzers to help customers generate support bundles from hosts to troubleshoot Kubernetes installer provisioned clusters (embedded clusters).
+
+## About Host Collectors and Analyzers
+
+Host collectors and analyzers are designed to collect information from hosts that is not available from in-cluster collectors. Host collectors are available only for Kubernetes installer clusters. These host collectors gather information directly from the host they are run on and do not have Kubernetes as a dependency.
```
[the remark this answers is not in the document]
jonquil2002: Paige and I decided that we would say "Kubernetes installer provisioned clusters" at the first mention in a topic, and then say "Kubernetes installer clusters" for the remainder. Though as a Product Team, if we get a majority to just call it "Kubernetes installer cluster" all the time, I'd be ok with that. It was one of those decisions that we had to make quickly with Amber when we were rolling out the new docs site, and none of us was thrilled with it. :)

docs/vendor/support-host-collectors.md
jonquil2002: Yessss!

docs/vendor/support-host-collectors.md
jonquil2002: Not sure what aggregate means here, I copied that from Troubleshoot.sh or the community site. So I removed it now.

Also changed the link to the one you gave me. Thanks.

jonquil2002:
@ajp-io @paigecalvert  Ready for a final review. I addressed all of Alex's suggestions (thank you!)

docs/enterprise/troubleshooting-an-app.md
```diff
@@ -85,3 +93,29 @@ If the application is not installed but the admin console is running, run the fo
   ```
   kubectl support-bundle http://<server-address>:8800/api/v1/troubleshoot
   ```
+
+### Generate a Host Support Bundle
+
+For Kubernetes installer provisioned clusters (embedded clusters), you can generate a host support bundle to help troubleshoot a cluster that is down. Your vendor provides you with a host support bundle YAML file that you run with a command to generate the host support bundle.
+
+Root access is typically not required to run the host support bundle and analyzers. However, depending on what is being collected, you might need to run the support-bundle binary with elevated permissions. For example, if you run the `filesystemPerformance` host collector against `/var/lib/etcd` and the user running the binary does not have permissions on this directory, the collection process fails.
```
ajp-io, line 101: We put `support-bundle` in code style above in the installing CLI section. Should we do that here?

docs/vendor/support-host-support-bundles.md
```diff
@@ -0,0 +1,60 @@
...
+
+You can gather information about the environment, such as CPU, memory, available block devices, and the operating system. Host support bundles can also be used for testing network connectivity and gathering the output of provided commands.
+
+This information is useful when you need to debug a Kubernetes installer cluster that is offline, troubleshoot a Kubernetes installer that failed before the control plane was initialized, or if you need to collect and analyze information that is not available with in-cluster collectors.
+
+You create the host support bundle manifest file separately from your application release and share the file with customers to run on their hosts. This file is separate from your application release because host collectors and analyzers are intended to run directly on the host and not with KOTS. If KOTS runs host collectors, the collectors are unlikely to produce the desired results because they run in the context of the kotsadm Pod. For more information about how customers generate a host support bundle, see [Generate a Host Bundle](troubleshooting-an-app#generate-a-host-support-bundle).
```
ajp-io, line 13: LOL at this question given our naming conversation, but is this a place where we should say app manager instead of KOTS? Not for kotsadm Pod, but for the two mentions of KOTS.

docs/vendor/support-host-support-bundles.md
```diff
@@ -0,0 +1,60 @@
...
+                  when: "count < 2"
+                  message: At least 2 CPU cores are required, and 4 CPU cores are recommended
+              - pass:
+                  message: This server has at least 4 CPU cores
+        - memory:
+            checkName: "Amount of Memory"
+            outcomes:
+              - fail:
+                  when: "< 4G"
+                  message: At least 4G of memory is required, and 8G is recommended.
```
ajp-io, line 56: This is such a nit that I'm making myself laugh, but none of the other messages have periods.
```suggestion
                  message: At least 4G of memory is required, and 8G is recommended
```

ajp-io (approved):
I left a few comments, but this is good to me.

docs/enterprise/troubleshooting-an-app.md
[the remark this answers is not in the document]
jonquil2002: TY, that was a CTL + F error.

docs/vendor/support-host-support-bundles.md
jonquil2002: Done, thanks!

docs/enterprise/troubleshooting-an-app.md
jonquil2002: I believe you were originally correct a couple of weeks ago with not putting it in code because it is the name of the CLI, just like kots CLI. I have fixed this in the installing CLI section.

paigecalvert:
@jonquil2002 LGTM!

paigecalvert (approved):

PR #640, opened 2022-09-16
58430: Corrected Air Gap Installation Info

jonquil2002:
SC: https://app.shortcut.com/replicated/story/58430/air-gap-install-on-embedded-cluster-procedure-incorrect

Previews:

-https://deploy-preview-640--replicated-docs.netlify.app/enterprise/installing-embedded-cluster#air-gap
- https://deploy-preview-640--replicated-docs.netlify.app/enterprise/installing-embedded-cluster#install-with-ha-in-air-gapped-environments

### DIFF
--- a/docs/enterprise/installing-embedded-cluster.md
+++ b/docs/enterprise/installing-embedded-cluster.md
@@ -62,7 +62,7 @@ To install an application in an air gapped environment:
     You can construct the URL for the air gap bundle by prefixing the URL path for online installations with `/bundle` and adding `.tar.gz` to the end. For more information, see [Install in an Online Environment](#online).
     :::
 
-  1. Install the application with one of the following methods:
+  1. Install the application using one of the following methods:
 
       - **Admin console:** Use the `kotsadm` and password from the previous step to log into the admin console. For information about using the admin console, see [Completing Application Setup and Deploying](installing-app-setup).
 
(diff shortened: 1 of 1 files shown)


paigecalvert (approved):
Thank you! Here's what I'm thinking, though approving so that you merge this if you prefer:

With the text about how you can use the CLI or the admin console appearing after the command that installs the app, it doesn't feel like it catches people at the right time. Ie, I might have already run the command by the time I see it. It also leaves the second step feeling like an odd transition from the first step, where we tell them to note the Kotsadm login credentials, then don't bring up the credentials again.

What do you think about breaking step two out into a couple options? Like, "2. Install the app: * With the CLI: ... * With the admin console: ..."

Then, maybe we could clarify for the Kotsadm credential output that you use these if you are installing the app using the admin console.

Something like: "After the installation command finishes, note the Kotsadm and Login with password (will not be shown again) fields in the output of the installation command. You can use these in the next step to log in to the admin console and install the application."

LMK what you think!

jonquil2002:
@paigecalvert  Good suggestions. I made the changes. Please approve or lmk if anything further is needed. Thanks!

paigecalvert (dismissed):
Looks great!

Only thing I noticed is a spacing formatting error with the markdown in the HA procedure (see screenshots)

Screen Shot 2022-09-20 at 11 07 39 AM
Screen Shot 2022-09-20 at 11 07 29 AM

paigecalvert (approved):

PR #649, opened 2022-09-21
Templates for vendors

paigecalvert:
@jonquil2002, in preparation for my Replicon presentation, I created a couple more templates that vendors could use:
* Troubleshooting template (useful for vendors, though also really for anyone at Replicated writing docs)
* App configuration template (as in, a topic that would walk an enterprise user through completing the configuration screen)

Story: https://app.shortcut.com/replicated/story/59314/create-app-config-and-troubleshooting-templates-for-vendors

### DIFF
--- a/docs/templates/app-configuration.md
+++ b/docs/templates/app-configuration.md
@@ -46,7 +46,7 @@ To configure [Section 1]:
 1. (Optional) For **[Field Name]**, enter [information].
   <!-- Put (Optional) at the start to denote optional fields.-->
 
-### Configuration Section 2
+### Configure [Section 2]
 
 To configure [Section 2]:
 
@@ -70,7 +70,7 @@ use the following table format-->
      </tr>
    </table>
 
-### (Optional) Configuration Section 3 <!-- Put (Optional) at the start to denote optional sections.-->
+### (Optional) Configure [Section 3] <!-- Put (Optional) at the start to denote optional sections.-->
 
 To configure [Section 3]:
 
@@ -82,7 +82,7 @@ To configure [Section 3]:
    <!-- Use this bulleted list format if the user needs to
    select between multiple options.-->
 
-<!-- ## Next Steps
+<!-- ## Next Step (or ## Next Steps)
 
 After you configure [App Name], [do this next step]. See [link to more information].
 
--- a/docs/templates/app-configuration.md
+++ b/docs/templates/app-configuration.md
@@ -68,7 +68,7 @@ use the following table format-->
        <td>(Optional) [Field Name]</td>
        <td>Enter [information].</td>
      </tr>
-   <table/>
+   </table>
 
 ### (Optional) Configuration Section 3 <!-- Put (Optional) at the start to denote optional sections.-->
 
--- a/docs/templates/app-configuration.md
+++ b/docs/templates/app-configuration.md
@@ -1,0 +1,101 @@
+# Configuring [App Name]
+
+<!-- Provide a topic sentence, such as the one below -->
+
+This topic includes procedures for configuring the following [components/features/functionality] of [App Name]:
+* [component/feature/functionality]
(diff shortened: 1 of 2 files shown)


docs/templates/app-configuration.md
```diff
@@ -0,0 +1,101 @@
...
+[Configuration Section 1] includes fields for configuring [a feature or service] in [App Name].
+
+To configure [Section 1]:
+
+1. For **[Field Name]**, enter [information].
+
+1. (Optional) For **[Field Name]**, enter [information].
+  <!-- Put (Optional) at the start to denote optional fields.-->
+
+### Configuration Section 2
```
jonquil2002, line 49: This heading doesn't match the one for the previous section. Maybe change this to "Configure [Section 2]".

docs/templates/app-configuration.md
```diff
@@ -0,0 +1,101 @@
...
+       <td>[Field Name]</td>
+       <td>Enter [information].</td>
+     </tr>
+     <tr>
+       <td>(Optional) [Field Name]</td>
+       <td>Enter [information].</td>
+     </tr>
+   <table/>
+
+### (Optional) Configuration Section 3 <!-- Put (Optional) at the start to denote optional sections.-->
```
jonquil2002, line 73: Same comment as for the heading above.

docs/templates/app-configuration.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. For **[Field Name]**, enter [information].
+
+1. For **[Field Name]**, select one of the following options:
+   * **[Option A]**: To [do x], select **[Option A]**.
+   * **[Option B]**: To [do x], select **[Option B]**.
+   <!-- Use this bulleted list format if the user needs to
+   select between multiple options.-->
+
+<!-- ## Next Steps
```
jonquil2002, line 85: "Next Step" (singular) if there is only one next step?

docs/templates/troubleshooting.md
```diff
@@ -0,0 +1,81 @@
+# Troubleshooting
+
+<!-- This template is used for troubleshooting topics. Troubleshooting topics
+provide users with workarounds and extra details about common challenges.
+Troubleshooting topics often include specific error messages.
+
+For an example of a troubleshooting topic, see https://docs.replicated.com/enterprise/snapshots-troubleshooting-backup-restore -->
+
+# Section One <!-- Example: "Service X is Crashing"-->
```
jonquil2002, line 9: Maybe say "Example heading:" so they don't think they need to include the word "example" in there?

docs/templates/troubleshooting.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Cause
+
+<!-- Background info about why the error occurs.-->
+
+### Solution
+
+<!-- Steps to resolve the issue.-->
+
+# Section One
```
jonquil2002, line 67: This says section one but I think it's supposed to be section two. 

Also, side note, I'm finding this template a little difficult to follow with section one, issues 1 ,2, etc. even though I know how we would actually use it. But it's just hard to follow as a template - in part because it's so long. Even though I know you'll be showing them an actual example, they might find this template confusing. Do we need to guide them this much?

paigecalvert: That's a great point, I agree. I think I'll remove the top layer of "section" headings. We can just go with issue headings, for sure. Whatcha think?

docs/templates/troubleshooting.md
paigecalvert: Ended up removing these sections all together; you're right it was overly complex

docs/templates/troubleshooting.md
jonquil2002: Super awesome!

jonquil2002 (approved):
lgtm - nice!

PR #652, opened 2022-09-23
Fix wording and remove mention of test procs

ajp-io:


docs/reference/custom-resource-config.md
```diff
@@ -73,8 +73,8 @@ See the [`ConfigOptionData`](template-functions-config-context#configoptiondata)
 ![File input field on the configuration screen](../../static/images/config-screen-file.png)
 
 ### `heading`
-The `heading` type allows you to display a group heading as a subelement within a group.
-This is useful when you would like to group items all in one config group but would like to separate the items visually, for instance when using a config group test proc.
+The `heading` type allows you to display a group heading as a sub-element within a group.
+This is useful when you would like to use a config group to group config items together, but still separate the config items visually.
```
jonquil2002, line 77: "config group to group config" is a little awkward. 

How about something like: "use a config group to group the config items together, but still separate the items visually"

ajp-io: Yeah, so just adding "the" in there. Alternatively we could say "group items" since it's clear we're talking about config items. Up to you!

jonquil2002 (approved):
lgtm, if you like my changes. (Not sure how to just suggest a commit, hope you don;t mind.)

PR #662, opened 2022-09-27
56498: Replaced Redactors Example

jonquil2002:
SC: https://app.shortcut.com/replicated/story/56498/update-example-in-the-redactor-custom-resource-topic

Previews:

- https://deploy-preview-662--replicated-docs.netlify.app/vendor/preflight-support-bundle-creating#customize-a-support-bundle
- https://deploy-preview-662--replicated-docs.netlify.app/reference/custom-resource-redactor#example-redactors


xavpaice (changes requested):
this is quite an aggressive redactor - `(?P<mask>[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+)` will replace anything that is word.word.word without considering word boundaries and so on.  This would potentially redact info in logs files that is required, e.g. `cluster-resources/resources.json:    "groupVersion": "rbac.authorization.k8s.io/v1"`.

We can test the regex with a bundle that's been un-tarred:

`egrep -r '[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+' *`

jonquil2002:
@xavpaice  Do you mean that I should replace `(?P<mask>[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+)` with this `egrep -r '[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+' *`? Or do you mean that the vendor should generate a support bundle and test the redactor ``(?P<mask>[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+)` using `egrep -r '[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+' *` to see if too much info has been redacted? If it's easier to explain by Zoom, please lmk.

xavpaice:
Unfortunatley, a host name redactor is very unlikely to be a feasible addition in this context.  While an individual could write a regex to achieve that, it would rely on the regex being a lot more specific to collect their naming convention, domain, etc. and that prevents any vendor from shipping a generic form for hostname redactor.

Perhaps what you need to complete this story is some other redactor - I'm struggling to think of some useful redactor right now that might be able to replace the redundant one, since the most generic important ones are already built in.  Perhaps someone who knows what customers are adding to their customisations would be able to chime in?

xavpaice:
can we not just copy the redactor example at the top of https://troubleshoot.sh/docs/redact/redactors/ ?

jonquil2002:
@xavpaice  I originally copied the Troubleshoot.sh example into the Reference section when I wrote this content a few months ago, and provided a simplified version using just password in the Customize a Support Bundle procedure. But someone pointed out that password is already in the default redactors, so I was searching for a generic replacement. Since you're also having trouble finding one, I just deleted the host name example in the procedure and pointed them to the Replicated docs reference section. In the reference section, I removed the portion that redacted password.

Side note - the example on the Troubleshoot.sh site should  be updated as well by removing the password example.

Please review and approve if possible Thanks.

Thank for your help! I'll update the final version shortly.

docs/reference/custom-resource-redactor.md
```diff
@@ -65,7 +65,7 @@ The `removals` object is required and defines the redactions that occur. This ob
   </tr>
 </table>
 
-## Example Redactors
+## Example Redactor
 
 The following example shows `values`, `regex`, and `yamlPath` redaction for a support bundle:
```
xavpaice, line 70: if the example doesn't show `values` we probably should remove that from here

xavpaice (changes requested):
one minor nit on line 70, otherwise yeah that looks great!

docs/reference/custom-resource-redactor.md
jonquil2002: Done, thanks.

jonquil2002:
@xavpaice @paigecalvert  Made Xav's last change. Please review/approve. Thanks.

paigecalvert (approved):
lgtm thank you!

xavpaice (approved):

PR #668, opened 2022-09-28
using-mdx-partials - DON'T MERGE

rianfowler:
example of how to use mdx and partials

### DIFF
--- a/docs/reference/kots-lint-new.mdx
+++ b/docs/reference/kots-lint-new.mdx
@@ -1,0 +1,874 @@
+import PreflightYaml from "./kots-lint-partials/_preflight-yaml.mdx"
+
+# KOTS Lint Rules
+
+This topic describes the rules for the KOTS lint service. The KOTS lint service lints the manifest files for applications packaged with Replicated to ensure that there are no YAML syntax errors, that all required manifest files are present in the release, and more.
+
+The KOTS lint service runs automatically against releases that you create in the Replicated vendor portal, and displays any error or warning messages in the vendor portal UI.
+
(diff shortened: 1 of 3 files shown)


docs/reference/kots-lint-new.mdx
```diff
@@ -0,0 +1,874 @@
...
+Example of **matching** yaml for this rule:
+
+```yaml
+apiVersion: troubleshoot.sh/v1beta2
+kind: Preflight
+```
+
+<PreflightYaml />
+
+
```
rianfowler, line 89: preview: 



Screen Shot 2022-09-28 at 12 56 32

rianfowler (commented):
MDX generates a lot of code when using the code block syntax:
````
```yaml
apiVersion: kots.io/v1beta1
kind: Config
```
````

We can have docusaurus generate this html and put it into a partial for us that can be used in markdown or in html. This is what I did:

1. Rename the page that's going to use the partial with `.mdx`. Docusaurus will pick the page up just like `.md` files. e.g. `kots-lint.md` -> `kots-lint.mdx`
2. Make a partial with the filename beginning with `_`. Files beginning with `_` are not picked up as their own pages. I made `_preflight-yaml.mdx` and put the markdown code in it. You can organize your partials however you want. I put mine in a folder since there will probably be multiple partials for this page. You will need to put each partial in its own file 
3. Import the partial in the page file. In my example it's at the top. There must be a space after the import statements before any markdown and you can call it whatever you want (e.g. PreflightYaml): 

```
import PreflightYaml from "./kots-lint-partials/_preflight-yaml.mdx"

```
4. Use the partial like any HTML element. You can use `<PartialYaml />` in HTML _and_ markdown as I did in my example.

PR #671, opened 2022-09-28
58875: Updated Kubernetes Installer System Requirements

jonquil2002:
SC: https://app.shortcut.com/replicated/story/58875/include-operating-system-requirements-for-kurl-in-documentation

Preview:

https://deploy-preview-671--replicated-docs.netlify.app/enterprise/installing-general-requirements#embedded-cluster-requirements

### DIFF
--- a/docs/enterprise/installing-general-requirements.md
+++ b/docs/enterprise/installing-general-requirements.md
@@ -90,7 +90,7 @@ Because the Kubernetes installer is based on the open source kURL project, which
 
 - **Supported Operating Systems** - For supported operating systems, see [Supported Operating Systems](https://kurl.sh/docs/install-with-kurl/system-requirements#supported-operating-systems) in the kURL documentation.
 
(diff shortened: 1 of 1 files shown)


jonquil2002:
@chris-sanders PTAL and see if this works better for you. It's for SC https://app.shortcut.com/replicated/story/58875/include-operating-system-requirements-for-kurl-in-documentation. 

The reason that the minimal system requirements are enumerated is because there was another story earlier that asked for the system requirements for KOTS be spelled out in the Replicated docs. Customers were wanting to know what the footprint is without having to go to another docs site. The related story is linked in the SC story above.

chris-sanders (dismissed):
This looks good, I like the explicit call out for these kURL requirements. This should be a lot clearer now and easier to find!

paigecalvert (dismissed):
Very well handled, looks great. Only recommendation is to use colons after the lead-in instead of hyphens, to match the formatting elsewhere (eg existing cluster reqs https://deploy-preview-671--replicated-docs.netlify.app/enterprise/installing-general-requirements#existing-cluster-requirements)

jonquil2002:
@paigecalvert  Just changed those to colons. PTAL Thanks!

paigecalvert (approved):

PR #672, opened 2022-09-28
58934: Assigning RBAC Policies

jonquil2002:
SC: https://app.shortcut.com/replicated/story/58934/procedure-for-assigning-a-vendor-portal-rbac-policy-to-a-team-member

Previews:
- https://deploy-preview-672--replicated-docs.netlify.app/vendor/team-management#invite-members
- https://deploy-preview-672--replicated-docs.netlify.app/vendor/team-management-rbac-configuring

### DIFF
--- a/docs/vendor/team-management-rbac-configuring.md
+++ b/docs/vendor/team-management-rbac-configuring.md
@@ -52,3 +52,5 @@ To configure custom RBAC policies:
   :::note
   Click **Cancel** to exit without saving changes.
   :::
+
+1. Assign the policy to team members. This can be done by inviting individual invitations or assigning a policy to a team using the auto-join feature. For more information, see [Managing Team Members](team-management).
--- a/docs/vendor/team-management-rbac-configuring.md
+++ b/docs/vendor/team-management-rbac-configuring.md
@@ -55,8 +55,8 @@ To configure custom RBAC policies:
 
 1. To use RBAC policies, you can:
 
-  - Assign policies to existing team members.
-  - Assign policies to the invitations for new team members.
-  - Set the policy as the default for the auto-join feature.
+    - Assign policies to existing team members.
+    - Assign policies to the invitations for new team members.
+    - Set the policy as the default for the auto-join feature.
 
   For more information about managing team members and assigning policies, see [Managing Team Members](team-management).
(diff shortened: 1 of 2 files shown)


docs/vendor/team-management-rbac-configuring.md
```diff
@@ -52,3 +52,5 @@ To configure custom RBAC policies:
   :::note
   Click **Cancel** to exit without saving changes.
   :::
+
+1. Assign the policy to team members by sending individual invitations or by assigning a policy to a team using the auto-join feature. For more information, see [Managing Team Members](team-management).
```
paigecalvert, line 56: Overall looks great, this was the only part that I got a little tripped up on. 

It looks like it mentions how you can send invites as a way to assign policies, but not assigning the policy to existing users (could maybe just keep it generic and say "you can assign it to individual users or set it as the default for all new members on a team"). I was also thinking it might be easier to parse if the sentence was broken down.  

What do you think about something like this?

"Assign the policy to individual team members. Or, if you have auto-join enabled, set the policy as the default for all new members. See..."

paigecalvert (dismissed):
Just one suggestion, thanks for adding this!

docs/vendor/team-management-rbac-configuring.md
ajp-io: I was going to bring up this same line. I think I have one nit with what Paige said. I see three ways to use these policies:
1. Assign to existing team members
2. Set as the default for auto-join
3. Assign as part of an invite that you send
I think Paige's comments don't hit on 3, and her comment about auto-join might be too expansive since people can still join via invite when auto-join is on. So "all new team members" might be a bit broad—it's really the default for team members who join via auto-join.

docs/vendor/team-management-rbac-configuring.md
```diff
@@ -52,3 +52,11 @@ To configure custom RBAC policies:
   :::note
   Click **Cancel** to exit without saving changes.
   :::
+
+1. To use RBAC policies, you can:
+
+    - Assign policies to existing team members.
+    - Assign policies to the invitations for new team members.
```
ajp-io, line 59: This wording is awkward to me because the policy isn't really assigned to the invitation—it's assigned to the user once they accept the invitation. So something like "Select a policy when inviting new team members." might be better.

paigecalvert (approved):
Nice!

PR #681, opened 2022-10-05
Fix bad support bundle link

ajp-io:


ajp-io:
There is a redirect from `/custom-resource-support-bundle` to `/custom-resource-preflight`, but it doesn't seem to work when you click on a link—only if you go straight to the original URL. So the link on [this page](https://docs.replicated.com/reference/custom-resource-about) doesn't work. This fixes that.

The link is `custom-resource-preflight` but that page now includes preflight and support bundle. Should we change the page name so that it's something that applies to both?

ajp-io:
@jonquil2002 Did you see my comment above? It's not a huge deal, but just a question I had.

ajp-io:
Yeah, I meant the file name, since it's not just preflight anymore. But that's fine, I just wondered. More redirects wouldn't be good either.

PR #685, opened 2022-10-07
58743-New: Updated Support Bundle CLI Procedures

jonquil2002:
SC: https://app.shortcut.com/replicated/story/58743/update-support-request-documentation-to-emphasize-providing-a-support-bundle

Preview: https://deploy-preview-685--replicated-docs.netlify.app/enterprise/troubleshooting-an-app

### DIFF
--- a/docs/enterprise/troubleshooting-an-app.md
+++ b/docs/enterprise/troubleshooting-an-app.md
@@ -30,11 +30,11 @@ You can generate a support bundle using the command line (CLI) instead of the ad
 
 Install or update the support-bundle plugin (a kubectl plugin) before you generate a support bundle using the CLI. The plugin is required to generate a support bundle.
 
+You can install the support-bundle plugin using krew or install it manually from the release archives.
+
 :::note
 If your application was installed using a Kubernetes installer script, then `kubectl` and the `support-bundle` plugin should already be present on all of the control plane nodes.
 :::
-
-You can install the support-bundle plugin using krew or install it manually from the release archives.
 
 #### Install using krew
 
@@ -75,7 +75,7 @@ To install the support-bundle plugin manually:
 
 ### Generate a Bundle on a Running Cluster
 
-If the admin console is running and the application is installed, run the following command to generate a support bundle that includes customizations provided by the software vendor:
+If the admin console is running and the application is installed, run the following command to generate a support bundle that includes customizations provided by your vendor:
 
   ```
   kubectl support-bundle secrets/NAMESPACE/APP_NAMEkotsadm-supportbundle
@@ -83,8 +83,8 @@ If the admin console is running and the application is installed, run the follow
 
 Replace:
 
-- NAMESPACE with the name of the namespace.
-- APP_NAME with the name of the application.
+- NAMESPACE with the name of the namespace
+- APP_NAME with the name of the application
 
 ### Generate a Host Support Bundle
 
@@ -95,7 +95,7 @@ For Kubernetes installer provisioned clusters (embedded clusters), you can gener
 - The admin console is not working
 - To debug a host-specific performance and configuration problems even when the cluster is running
 
-Your vendor typically provides you with a host support bundle YAML file that you run with a command to generate the host support bundle. You can also create a your own host support bundle YAML file. For more information about creating a YAML file, see [All Host Collectors and Analyzers](https://troubleshoot.sh/docs/host-collect-analyze/all/) in the Troubleshoot documentation.
+Your vendor typically provides a host support bundle YAML file that you run with a command to generate the host support bundle. You can also create a your own host support bundle YAML file. For more information about creating a YAML file, see [All Host Collectors and Analyzers](https://troubleshoot.sh/docs/host-collect-analyze/all/) in the Troubleshoot documentation.
 
 If you want to use a default YAML file from the Troubleshoot.sh repository for troubleshooting a degraded cluster, follow the command in the steps below.
 
--- a/docs/enterprise/troubleshooting-an-app.md
+++ b/docs/enterprise/troubleshooting-an-app.md
@@ -22,15 +22,17 @@ Alternatively, you can generate a support bundle using the CLI, which can be hel
 
 1. (Optional) Click **Download bundle** to download the support bundle. You can send the bundle to your vendor for assistance.
 
-## Generating a Bundle Using the CLI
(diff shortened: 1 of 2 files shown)


docs/enterprise/troubleshooting-an-app.md
```diff
@@ -20,15 +22,23 @@ The Replicated admin console includes a Troubleshoot page where you can generate
...
+### Prerequisite: Install the Plugin {#plugin}
+
+Install or update the support-bundle plugin (a kubectl plugin) before you generate a support bundle using the CLI. The plugin is required to generate a support bundle.
 
-You can generate a support bundle using the CLI instead of the admin console. For example, if an installation fails when you are using an embedded KURL cluster to install the Replicated admin console or upload the application, the admin console may not be available.
+You can install the support-bundle plugin using krew or install it manually from the release archives.
 
-### Install the Support Bundle Plugin {#plugin}
+:::note
+If your application was installed using a Kubernetes installer script, then `kubectl` and the `support-bundle` plugin should already be present on all of the control plane nodes.
```
adamancini, line 36: I think it bears mentioning here that this is only an expectation if you're using kURL - I know we're moving away from using the project name as the feature name but I don't want people to think that just any ol' kubernetes installer will set this up for them.

docs/enterprise/troubleshooting-an-app.md
```diff
@@ -20,15 +22,23 @@ The Replicated admin console includes a Troubleshoot page where you can generate
...
+:::
 
-The support-bundle kubectl plugin is required to generate a support bundle from the command line.
+#### Install using krew
 
-To install the plugin, do one of the following actions:
+To install the support-bundle plugin using krew, do one of the following actions:
 
 - If krew is _not_ installed, run the following command:
```
adamancini, line 43: I think this should be:



    - If krew *is not* already installed, run the following command:
    ```
    curl https://krew.sh/support-bundle | bash
    ```
    - If krew *is* already installed, run the following command:
    ```
    kubectl krew install support-bundle
    ```

docs/enterprise/troubleshooting-an-app.md
```diff
@@ -41,81 +51,111 @@ To install the plugin, do one of the following actions:
...
+  sudo mv ./support-bundle /usr/local/bin/kubectl-support-bundle
   ```
+  :::note
+  If you do not have root access, or choose not to add the support-bundle plugin to your path, you can run the binary directly from where you unzipped it by executing `./support-bundle`.  If you choose not to put the plugin into your $PATH, then replace all instances of `kubectl support-bundle` in these instructions with `./support-bundle` or with the absolute path to the binary.
+  :::
 
-  Replace `VERSION` with `v` and the version number. For example, `v0.40.0`. To get the latest release version, see the [Troubleshoot](https://github.com/replicatedhq/troubleshoot/releases) repository in GitHub.
-
-### Create a Support Bundle with the CLI
+1. Run the following command to install the plugin:
```
adamancini, line 70: Once the binary has been `mv`ed into /usr/local/bin or if you run from `./support-bundle` theres no more installation - we could say something like "run this to make sure it's working" or we could remove lines 70-74 I think

docs/enterprise/troubleshooting-an-app.md
```diff
@@ -41,81 +51,111 @@ To install the plugin, do one of the following actions:
...
 
-#### With the Default kots.io Specification
+### Generate a Bundle on a Running Cluster
 
-To use the default kots.io specification, run the following command to create a support bundle:
+If the admin console is running and the application is installed, run the following command to generate a support bundle that includes customizations provided by your vendor:
 
   ```
-  kubectl support-bundle https://kots.io
+  kubectl support-bundle secrets/NAMESPACE/APP_NAMEkotsadm-supportbundle
```
adamancini, line 81: double-checked today and the right syntax is 

`kubectl support-bundle secrets/<namespace>/kotsadm-<appslug>-supportbundle`

docs/enterprise/troubleshooting-an-app.md
```diff
@@ -41,81 +51,111 @@ To install the plugin, do one of the following actions:
...
 
-For Kubernetes installer provisioned clusters (embedded clusters), you can generate a host support bundle to help troubleshoot a cluster that is down. Your vendor provides you with a host support bundle YAML file that you run with a command to generate the host support bundle.
+1. Share the host support bundle with your vendor's support team, if needed.
 
-Root access is typically not required to run the host collectors and analyzers. However, depending on what is being collected, you might need to run the support-bundle binary with elevated permissions. For example, if you run the `filesystemPerformance` host collector against `/var/lib/etcd` and the user running the binary does not have permissions on this directory, the collection process fails.
+1. Repeat these steps for each node because there is no method to generate host support bundles on remote hosts. If you have a multi-node Kubernetes cluster, you must run the support-bundle binary on each node and generate a host support bundle for each node. Note that the support-bundle plugin must be installed on each control plane node to generate a support bundle from that node.
```
adamancini, line 131: I think "Note that the support-bundle plugin must be installed on each control plane node to generate a support bundle from that node." is redundant, we just explained we need to have the support bundle installed on all the hosts in a multi-node cluster.

docs/enterprise/troubleshooting-an-app.md
jonquil2002: Fair point, though we only use "Kubernetes installer" and "Kubernetes installer cluster" (instead of embedded cluster) for kURL installations. We never name a regular installer or I cluster with "Kubernetes", per our branding. But for clarification of your point, I changed it to "Replicated Kubernetes installer" - hopefully PM will be ok with that :)

docs/enterprise/troubleshooting-an-app.md
jonquil2002: @adamancini I changed it to what you have - with the caveat of using `NAMESPACE` instead of `<namespace>` and APP_SLUG` instead of `<appslug> to match our style guidelines. But I just want to make sure that `<>` aren't needed before I mark this resolved. Does it look ok?

jonquil2002:
@adamancini One question above. And if the changes look good, please approve. Thanks!

adamancini (approved):

adamancini:
lgtm

paigecalvert (approved):
@jonquil2002 LGTM

PR #692, opened 2022-10-10
replicated cli reference: registry

marccampbell:
One doc was manually updated, the rest were all generated from the upcoming Replicated CLI release.


paigecalvert (approved):
Approving so that this can merged in today. Style/formatting updates can be done later as part of this follow-up story: https://app.shortcut.com/replicated/story/60634/style-formatting-updates-for-new-cli-commands

PR #699, opened 2022-10-13
Supported Versions table

paigecalvert:
https://app.shortcut.com/replicated/story/59722/add-table-of-support-versions-for-kots-kurl

Preview: https://deploy-preview-699--replicated-docs.netlify.app/vendor/policies-support-lifecycle#supported-versions

### DIFF
--- a/docs/enterprise/installing-general-requirements.md
+++ b/docs/enterprise/installing-general-requirements.md
@@ -33,15 +33,9 @@ The app manager is based on the open source KOTS project. The app manager versio
 
 | KOTS Versions   | Kubernetes Compatibility |
 |-----------------|---------------------------|
-| v1.71 and later | v1.25, 1.24, v1.23, v1.22, and v1.21|
-| v1.66 to v1.70 | v1.23, v1.22, and v1.21   |
-| v1.61 to v1.65 | v1.23, v1.22, v1.21, and v1.20|
-| v1.59.3 to v1.60 | v1.22, v1.21, and v1.20 |
-| v1.48 to v1.59.2 | v1.21, v1.20, and v1.19|
-| v1.36 to v1.47  | v1.20, v1.19, and v1.18 |
-| v1.20 to v1.35  | v1.19, v1.18, and v1.17 |
-| v1.15 to v1.19  | v1.18, v1.17, and v1.16 |
-| v1.11 to v1.14  | v1.17, v1.16, and v1.15 |
+| v1.71 and later | v1.25, 1.24, v1.23 |
+| v1.66 to v1.70 | v1.23 |
+| v1.61 to v1.65 | v1.23 |
 
 ## Minimum System Requirements
 
--- a/docs/enterprise/installing-general-requirements.md
+++ b/docs/enterprise/installing-general-requirements.md
@@ -23,17 +23,17 @@ The following table lists the browser requirements for the latest Replicated adm
 
 Each release of the open source KOTS project maintains compatibility with the current Kubernetes version, and the two most recent versions at the time of its release. This includes support against all patch releases of the corresponding Kubernetes version.
 
-Kubernetes versions 1.21 and earlier are end-of-life (EOL). For more information about Kubernetes versions, see [Release History](https://kubernetes.io/releases/) in the Kubernetes documentation.
+Kubernetes versions 1.22 and earlier are end-of-life (EOL). Kubernetes version 1.23 is nearing EOL. For more information about Kubernetes versions, see [Release History](https://kubernetes.io/releases/) in the Kubernetes documentation.
 
-Replicated recommends upgrading to a KOTS version that is compatible with Kubernetes 1.22 and higher.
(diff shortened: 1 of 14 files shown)


paigecalvert:
@chris-sanders brushing the cobwebs off this one 😄  Before I share this draft of this new "Supported Versions" table with the Product team, would you be able to review the details here to see if it conveys what you were wanting to see?

A few questions in particular:
- Does it over-simplify? I know there's some grey area when it comes to supporting KOTS-specific issues even when the cluster might be running an EOL version of Kubernetes
- I did not include Kubernetes versions that are already EOL in the table. Does that make sense to you? Tricky for KOTS in particular, which still lists 1.21 and 1.22 as supported in the latest releases
- It doesn't look like KOTS supports 1.26 yet, so I put "N/A" in that column. LMK if that makes sense to you
- Would you like to see a link to the official support service description policy?

paigecalvert:
@chris-sanders sounds good! And I was referring to your mention in the story of "While working on support service description the phrasing that has been agreed is the following:". I just assumed that this was something that was published already. But sounds like it's still in approval

paigecalvert:
Hi @ajp-io @kyleprice @AmberAlston , this PR to add a table to clarify the "End of Replicated Support" date for the versions of KOTS and kURL is ready for product review.

There are more details about this in the story requested by Chris: https://app.shortcut.com/replicated/story/59722/add-table-of-support-versions-for-kots-kurl

The high-level summary is that we want to make it easier for vendors to see the dates when Replicated support ends, based on the version of k8s, KOTS, and kURL running in the cluster. (Essentially, we support until the version of k8s is EOL, and there are certain versions of KOTS and kURL that support the different versions of k8s) 

Specific areas for you to confirm:
- Accuracy of the versions listed
- That you agree with the messaging

Thank you!

chris-sanders:
> I just assumed that this was something that was published already. But sounds like it's still in approval

Ok we're on the same page then, and I think it would be great to include it once we have it published.

AmberAlston (changes requested):
Last updated note on page needs to be updated to the intended publish date. Would make a PR but I’m unsure of the target date.

AmberAlston:
Overall, I like the way we’ve constructed this table to not list old versions we consider EOL, rather it’s implied. We’re in a bit of a sticky area right now where we’ve made some commitments to help certain vendors troubleshoot issues even though they are on older K8s/KOTS versions, until they can successful upgrade all the way to in-support versions. My previous concern was that publishing this could imply we’re not honoring those commitments, but as written we can still speak to those concerns as exceptions that just aren’t listed here. 

**Question** - Is the intent to keep this table updated as well, https://docs.replicated.com/enterprise/installing-general-requirements#kubernetes-version-compatibility? It’s actually a pretty useful historical reference while we’ve got folks on outdated versions of kots and kURL, but I could see it becoming redundant over time as we bring folks up-to-date to actually supported versions. May make sense to cross-reference to this section instead.

ajp-io:
I think Amber touched on all of my thoughts or concerns.

In some ways, its nice to have the historical reference of each version and which K8s versions it supports. But maybe that is actually unhelpful to a point. Because if a customer is on app manager v1.48, v1.48 isn't compatible with any in-support versions of Kubernetes. So even if there is an issue, the customer is going to have to upgrade, which means they'll need to upgrade both app manager and Kubernetes in order to stay in support. So maybe the historical stuff is irrelevant for customers. If you're on something no longer supported, it doesn't matter which versions of Kubernetes that was compatible with—you're going to have to upgrade either way.

TL;DR I think [this chart](https://docs.replicated.com/enterprise/installing-general-requirements#kubernetes-version-compatibility) could go away. Supported K8s versions are in release notes anyway.

paigecalvert:
Good catch on that other table.

The one thing is that that compatibility table is part of the Enterprise documentation, so intended to be something that the vendor is using to write their own docs, which are read by their users. (To that end, linking to the support policies might not fit.) I'd also expect good docs from a vendor to list out for their users what the k8s compatibility is for the different versions of KOTS.

LMK what you think of these proposed updates to the compatibility table:
- Remove the versions of KOTS that only support EOL versions of kubernetes (that knocks off 1.60 and earlier)
- For the remaining versions of KOTS, only list versions of Kubernetes that are in-support (as in, even though 1.92 supports 1.21 and 1.22 of Kubernetes, don't list them in this table since they are EOL. My thought process is that there is no need to advertise 1.21 and 1.22 on a table that end users are meant to read _before_ they install for the first time)

What do you think @AmberAlston @ajp-io ? (lmk if I correctly understood the concerns)

AmberAlston:
@paigecalvert I'm good with that proposal. I went ahead and opened an issue with that page because there are a few outdated pieces of content I want to make sure we update. You can likely close out this issue when updating that table - https://github.com/replicatedhq/replicated-docs/issues/882

paigecalvert:
Mistakes were made on resolving merge conflict...closing this PR in favor of https://github.com/replicatedhq/replicated-docs/pull/887 to clean things up.

PR #700, opened 2022-10-13
Fix description of `exclude` logic

markpundsack:
The text for `exclude` logic in preflights and support bundles is backwards.


jonquil2002 (approved):
lgtm

PR #714, opened 2022-10-21
60846: Clarify which fields in Application CR support templating with Go

paigecalvert:


jonquil2002 (approved):
LGTM. The only thing that stands out is the scrolling in the statusInformers table. It's not ideal and I'm sure you've played with it, but if it's possible not to have to scroll on a MacBook laptop size screen, that would be nice. But it's ok as is.

PR #716, opened 2022-10-24
59979: Added renewal clarification

jonquil2002:
SC: https://app.shortcut.com/replicated/story/59979/remove-the-kots-specific-tls-certs-content-from-kurl-sh

This is just part one of this SC card - just adding the ECKO add-on info for the self-signed cert renewal. The removal of the kURL content will happen in another PR in the kURL repo.

Preview: https://deploy-preview-716--replicated-docs.netlify.app/vendor/packaging-using-tls-certs


paigecalvert (approved):
LGTM from docs perspective!

PR #722, opened 2022-10-26
61054: Create a CLI Tutorial Section

paigecalvert:
https://app.shortcut.com/replicated/story/61054/break-out-cli-getting-started-tutorial-into-steps

Preview: https://deploy-preview-722--replicated-docs.netlify.app/vendor/tutorial-cli-install-cli


paigecalvert:
@jonquil2002 this is ready for review. I'm wondering if it might be worth it to pair on merging these PRs to verify that the structure, file naming, organizational choices, redirects, etc are all using the same format. (Or if we don't see any differences, might not be necessary). LMK what you think

PR #726, opened 2022-10-28
61019: Updated Description of Replicated Components

jonquil2002:
SC: https://app.shortcut.com/replicated/story/61019/improve-description-of-replicated-features

Preview: https://deploy-preview-726--replicated-docs.netlify.app/intro-replicated


paigecalvert:
@jonquil2002 this is looking really good! I'm looking forward to seeing what PM thinks on the diagram. I think the only remaining Qs I have for the diagram specifically are:
- I wasn't sure what the multiple "VM" boxes around existing cluster installs & the one box for K8s installer installs represents.
- The term Kubernetes installer doesn't appear, and I think it has to do with the fact that the diagram might not convey that the app and admin console are still being installed _in_ a Kubernetes cluster _on_ a VM (right now it looks like they are just being installed directly on a VM)

And then a couple higher-level thoughts, just for your consideration:
- For the graphical user interfaces (vendor portal and admin console), screenshots could be really helpful for building initial familiarity. I'd lean towards screenshots over tables, if it becomes a decision between the two. (Especially since the info in the tables is not too complex. Could also be beneficial to pull the component names out as headings rather than rows in tables. More on that below.)
- It looks like there is not a section that describes the Kubernetes installer--was this intentional?
- Finally, thinking about the overall organization/heading names, the topic currently uses these headings: Vendor UIs, App Manager, Customer UIs. App Manager is an actual component name, while the other headings are buckets. I found this a little confusing, because it looks like the App Manager is something separate, even though it's what's powering the admin console and kots CLI.

  Maybe it would make sense to put everything under two buckets: vendor stuff versus enterprise customer stuff? Or alternatively, use the exact component names for the headings without buckets?

  Whether or not there are buckets, a benefit of using the exact component names as h2s or h3s would be having the components names appear in the right TOC.

jonquil2002:
@paigecalvert Thanks for the great suggestions. Let's see if this works better.

@ajp-io Would appreciate your input on the diagram, and also the content could use a pass for technical accuracy. Just want to make sure I didn't make mistakes during the rewrites. Please lmk if we should pair, esp. on the diagram. I'm planning to ask the marketing team to render a better diagram once I know this reflects what PM wants to communicate. Thanks!

docs/intro-replicated.md
```diff
@@ -1,60 +1,66 @@
...
+
+The Replicated admin console is a graphical user interface that interacts with the app manager. Enterprise customers can use the admin console to manage, update, configure, monitor, backup and restore, and troubleshoot their application instance.
+
+The admin console is installed with the application by the app manager.
+
+![Admin Console Dashboard](/images/guides/kots/application.png)
+
+### kots CLI
+
+Enterprise customers can install and use the kots command-line interface (CLI) to manage and troubleshoot the application programmatically. The kots CLI interacts with the app manager and provides an alternative to using the admin console. For more information, see [Installing the kots CLI](/reference/kots-cli-getting-started).
```
paigecalvert, line 66: Just wasn't sure about the "troubleshoot" part of this, since downloading the support bundle requires the support-bundle kubectl plugin. Maybe "update" could work here instead?

paigecalvert (commented):
Caught a few things, and added some suggestions. Looks great! I like the idea of just listing everything under it's own heading

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
-users to complete tasks programmatically:
-
-* **Vendor API v3**: The API for the vendor portal. See [Using the Vendor API v3](reference/vendor-api-using).
-* **replicated CLI**: The CLI for the vendor portal. See [Installing the replicated CLI](reference/replicated-cli-installing).
-* **kots CLI**: The CLI for the KOTS open source project. The kots CLI allows users to interact with the app manager on the command line, rather than through the admin console. See [Installing the kots CLI](reference/kots-cli-getting-started).
+Replicated is an enterprise software distribution platform that lets you package and update your application using Kubernetes manifests or Helm charts, and securely distribute your application to Kubernetes clusters in diverse customer environments.
+
+Environments include on-premises (including air gapped) and cloud hosted Kubernetes clusters. For customers who do not have an existing Kubernetes cluster, Replicated provisions a new cluster on a customer's virtual machine (VM), where the application is then installed.
+
+To learn more about Replicated, see [Intro to Replicated: 20 Ways We Help](https://www.youtube.com/watch?v=2eOh7CofY3Q&t=779s) video.
```
paigecalvert, line 7: Just noticed a typo here (missing "the")

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+### App Manager
+
+The Replicated app manager is the underlying technology that installs and manages applications on a Kubernetes cluster. It also installs the Replicated admin console and provides preflight and support bundle functionality from the Troubleshoot open source project. For more information about preflight checks and support bundles, see [Configuring Preflight Checks and Support Bundles](preflight-support-bundle-creating). For more information about the Troubleshoot open source project, see the [Troubleshoot repository](https://github.com/replicatedhq/troubleshoot) in GitHub.
+
+The app manager deploys applications securely to the following Kubernetes cluster environments:
+
+- Existing clusters
+- Replicated Kubernetes installer provisioned clusters on virtual machines (VMs)
+- Air gapped on existing clusters or VM clusters
```
paigecalvert, line 49: Does VM clusters mean Kubernetes installer provisioned clusters? (Recommend changing if so/clarifying what it means if not; I wasn't sure about the difference between "VM clusters" and "Kubernetes installer provisioned clusters" in the bullet above)

paigecalvert (dismissed):
Two small comments (had missed a typo the first pass). Looks great!

docs/intro-replicated.md
ajp-io: Is this link supposed to go to a specific timecode in the video? (The `t=779s` part)

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
-* **kots CLI**: The CLI for the KOTS open source project. The kots CLI allows users to interact with the app manager on the command line, rather than through the admin console. See [Installing the kots CLI](reference/kots-cli-getting-started).
+Replicated is an enterprise software distribution platform that lets you package and update your application using Kubernetes manifests or Helm charts, and securely distribute your application to Kubernetes clusters in diverse customer environments.
+
+Environments include on-premises (including air gapped) and cloud hosted Kubernetes clusters. For customers who do not have an existing Kubernetes cluster, Replicated provisions a new cluster on a customer's virtual machine (VM), where the application is then installed.
+
+To learn more about Replicated, see [Intro to Replicated: 20 Ways We Help](https://www.youtube.com/watch?v=2eOh7CofY3Q&t=779s) video.
+
+## Replicated Components
+
+Replicated includes components and features that make it easier for you to manage and deploy applications, and for enterprise users to install and manage their application instance.
```
ajp-io, line 11: Total suggestion, take it or leave it, but this wording sounds better to me.
```suggestion
Replicated includes components and features that make it easier for you to manage and deploy applications, and for enterprise users to install and manage their instance of your application.
```

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+The following diagram shows the Replicated components as they relate to you as a vendor packaging your application and the deployment to an existing cluster and a VM cluster provisioned by the Replicated Kubernetes installer.
```
ajp-io, line 13: I don't think "VM cluster" is the right wording here. Maybe this?
```suggestion
The following diagram shows the Replicated components as they relate to you as a vendor packaging your application and the deployment to an existing cluster and a cluster provisioned by the Replicated Kubernetes installer on a VM.
```

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+[View larger image](/images/what-is-replicated.png)
+
+### Vendor Portal
+
+The Replicated vendor portal is the web-based user interface that you can use to package and manage applications.
+
+![Create an Application in the vendor portal](/images/guides/kots/create-application.png)
+
+You define Kubernetes manifest files, including application and custom resource manifests, using the built-in YAML editor and KOTS lint service (in the Help pane). These files describe how to distribute the application. Alternatively, you can use Helm charts.
```
ajp-io, line 25: I know this repo or whatever is called kots-lint, but is that a name we should use externally, especially as we move away from KOTS terminology?

ajp-io, line 25: Is this talking about our custom resources or theirs?
>including application and custom resource manifests

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+![YAML editor in the vendor portal](/images/yaml-editor.png)
```
ajp-io, line 27: The file selector sidebar in this image looks funny. The add file and add folder buttons are overlapping existing file names. I wonder if that's an existing bug in the vendor portal.

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+You can also manage other artifacts, such as customer license files, image registries, and release channels.
+
+![Channels](/images/channels.png)
```
ajp-io, line 31: In this picture, "Installer history" overlaps "Kubernetes Installer". Perhaps because the browser window was resized too small for it all to fit. Might want to see if we can adjust that.

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+### Vendor API v3
+
+The Vendor API is the API for the vendor portal. This API can be used to complete tasks programmatically, including all tasks for packaging and managing applications, and managing artifacts such as teams, license files, and so on. For more information, see [Using the Vendor API V3](/reference/vendor-api-using).
+
+### App Manager
+
+The Replicated app manager is the underlying technology that installs and manages applications on a Kubernetes cluster. It also installs the Replicated admin console and provides preflight and support bundle functionality from the Troubleshoot open source project. For more information about preflight checks and support bundles, see [Configuring Preflight Checks and Support Bundles](preflight-support-bundle-creating). For more information about the Troubleshoot open source project, see the [Troubleshoot repository](https://github.com/replicatedhq/troubleshoot) in GitHub.
```
ajp-io, line 43: "also installs" isn't accurate here since the app manager is the only thing installed. This goes back to our naming conversation :) I think we should reword this to say that there is a user interface component, but I don't think we should say that it installs an admin console.

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+The app manager deploys applications securely to the following Kubernetes cluster environments:
+
+- Existing clusters
+- Replicated Kubernetes installer provisioned clusters on virtual machines (VMs)
+- Air gapped on existing clusters or VM clusters
+- High availability clusters
```
ajp-io, line 50: Is this Kubernetes installer clusters that are provisioned with the `ha` flag?

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+### Kubernetes Installer
+
+The Replicated Kubernetes installer provisions a cluster on a customer's virtual machine. This is known as a Kubernetes installer cluster or embedded cluster, and it allows customers to install an application without an existing Kubernetes cluster. For more information about configuring a Kubernetes installer cluster, see [Creating a Kubernetes Installer](/vendor/packaging-embedded-kubernetes). For more information about how enterprise users install with the Kubernetes installer, see [Installing with the Kubernetes Installer](/enterprise/installing-embedded-cluster).
+
+The Kubernetes installer is based on the kURL open source project, which is maintained by Replicated. For more information, see the [kURL repository](https://github.com/replicatedhq/kURL) in GitHub.
+
+### Admin Console
+
+The Replicated admin console is a graphical user interface that interacts with the app manager. Enterprise customers can use the admin console to manage, update, configure, monitor, backup and restore, and troubleshoot their application instance.
```
ajp-io, line 62: I wouldn't say it "interacts with the app manager", because ultimately it is the app manager. Maybe this:
```suggestion
The Replicated admin console is a graphical user interface for the app manager. Enterprise customers can use the admin console to manage, update, configure, monitor, backup and restore, and troubleshoot their application instance.
```

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+The admin console is installed by the app manager.
```
ajp-io, line 64: Not really true, as I said above.

ajp-io (changes requested):
Is Katie going to formalize the diagram?

docs/intro-replicated.md
ajp-io: I see we call it that here. That seems like something we should name differently IMO, though that's a separate issue from this PR. https://docs.replicated.com/reference/kots-lint#resource-requests-cpu

docs/intro-replicated.md
jonquil2002: Yep, good catch. I'm just going to call it a Kubernetes installer cluster on the second mention, since that's what we've been doing elsewhere.

docs/intro-replicated.md
jonquil2002: Yes, it is Replicated CRs. Thanks for the suggestion.

docs/intro-replicated.md
jonquil2002: Thanks, I hesitated over that wording but couldn't find that simpler, better word :)

docs/intro-replicated.md
jonquil2002: @ajp-io  It's high availability mode with the Kubernetes installer: https://docs.replicated.com/enterprise/installing-embedded-cluster#installing-with-high-availability-mode

Is that topic impacted with the recent changes we're making to `ha`?

docs/intro-replicated.md
jonquil2002: @ajp-io It does seem to be a bug in the vendor portal. I'll open an SC. It seems that no matter what size screen I use, that displays like that if there are too many files.

For this PR, I changed the screenshot to show fewer files so that the add file and add folder buttons display well.

jonquil2002:
@ajp-io Please re-review - I incorporated all of your changes. Have a question above on the `ha` discussion.

Also, I'll ask Katie to revise the diagram in a follow-up story. We'll just go with this one for now as long as it's accurate.

docs/intro-replicated.md
ajp-io: It's not impacted right now.

I wonder if we need a dedicated bullet for that. Technically that's just a Kubernetes installer provisioned cluster that's provisioned in a certain way, and it doesn't necessarily follow from the current text ("High availability clusters) that this is related to kURL.

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
-
-Replicated includes a Vendor API and CLIs that allow both vendors and enterprise
-users to complete tasks programmatically:
-
-* **Vendor API v3**: The API for the vendor portal. See [Using the Vendor API v3](reference/vendor-api-using).
-* **replicated CLI**: The CLI for the vendor portal. See [Installing the replicated CLI](reference/replicated-cli-installing).
-* **kots CLI**: The CLI for the KOTS open source project. The kots CLI allows users to interact with the app manager on the command line, rather than through the admin console. See [Installing the kots CLI](reference/kots-cli-getting-started).
+Replicated is an enterprise software distribution platform that lets you package and update your application using Kubernetes manifests or Helm charts, and securely distribute your application to Kubernetes clusters in diverse customer environments.
+
+Environments include on-premises (including air gapped) and cloud hosted Kubernetes clusters. For customers who do not have an existing Kubernetes cluster, Replicated provisions a new cluster on a customer's virtual machine (VM), where the application is then installed.
```
ajp-io, line 5:
```suggestion
Environments include on-premises (including air gapped) and cloud-hosted Kubernetes clusters. For customers who do not have an existing Kubernetes cluster, Replicated provisions a new cluster on a customer's virtual machine (VM), where the application is then installed.
```

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+To learn more about Replicated, see the [Intro to Replicated: 20 Ways We Help](https://www.youtube.com/watch?v=2eOh7CofY3Q) video.
+
+## Replicated Components
+
+Replicated includes components and features that make it easier for you to manage and deploy applications, and for enterprise users to install and manage their instance of your application.
+
+The following diagram shows the Replicated components as they relate to you as a vendor packaging your application and the deployment to an existing cluster and a cluster provisioned by the Replicated Kubernetes installer on a VM.
```
ajp-io, line 13:
```suggestion
The following diagram shows the Replicated components as they relate to you as a vendor packaging your application, and the deployment to an existing cluster and a cluster provisioned by the Replicated Kubernetes installer on a VM.
```

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+The Replicated app manager is based on the open source KOTS project, which is maintained by Replicated.
+
+### Kubernetes Installer
+
+The Replicated Kubernetes installer provisions a cluster on a customer's virtual machine. This is known as a Kubernetes installer cluster or embedded cluster, and it allows customers to install an application without an existing Kubernetes cluster. For more information about configuring a Kubernetes installer cluster, see [Creating a Kubernetes Installer](/vendor/packaging-embedded-kubernetes). For more information about how enterprise users install with the Kubernetes installer, see [Installing with the Kubernetes Installer](/enterprise/installing-embedded-cluster).
+
+The Kubernetes installer is based on the kURL open source project, which is maintained by Replicated. For more information, see the [kURL repository](https://github.com/replicatedhq/kURL) in GitHub.
+
+### Admin Console
```
ajp-io, line 60: Maybe this should come right after App Manager instead of having Kubernetes Installer sandwiched between them.

docs/intro-replicated.md
```diff
@@ -1,60 +1,70 @@
...
+
+The Replicated admin console is a graphical user interface for the app manager. Enterprise customers can use the admin console to manage, update, configure, monitor, backup and restore, and troubleshoot their application instance.
+
+The admin console is the graphical user interface for the app manager.
+
```
ajp-io, line 65: This is in the above paragraph already.
```suggestion
```

ajp-io (commented):
Couple small things. And no problem about the diagram. Just wanted to know if we were going to formalize it at some point.

docs/intro-replicated.md
jonquil2002: I can remove that bullet. Originally, I was wondering whether or not it was needed in this context.

jonquil2002:
@ajp-io @paigecalvert  Made Alex's last changes. Please review/approve if possible. Thanks.

paigecalvert (approved):

ajp-io (approved):

PR #742, opened 2022-11-01
61313: Add CLI commands for kots CLI installation

jonquil2002:
SC: https://app.shortcut.com/replicated/story/61313/add-commands-to-steps-on-installing-the-kots-cli-in-air-gap-cluster


chasehainey (approved):
I think this clears up some of the potential confusion of the CLI install for folks that may be in an Airgap environment.

paigecalvert (approved):

PR #753, opened 2022-11-04
61128: UI Tutorial Updates

jonquil2002:
SC: https://app.shortcut.com/replicated/story/61128/improve-ui-tutorial-content

Preview starts here: https://deploy-preview-753--replicated-docs.netlify.app/vendor/tutorial-ui-setup


docs/vendor/tutorial-ui-create-app.md
```diff
@@ -1,5 +1,7 @@
 # Step 1: Create an Application
```
paigecalvert, line 1: When I logged in, I wasn't taken to the Create an App page, since I'm part of an existing team. The alternative step I think would be going to the Applications drop down and clicking "Create new app..." 

Screen Shot 2022-11-10 at 1 30 47 PM

docs/vendor/tutorial-ui-create-new-version.md
```diff
@@ -1,34 +1,102 @@
...
+     type: text
+     value: ""
+     when: repl{{ ConfigOptionEquals "show_text_inputs" "1" }}
   ```
 
-1. Change the number to `2` or more.
+  The following shows the full YAML for the `kots-config.yaml` file after you add the new field:
+
+  ```yaml
+   ---
```
paigecalvert, line 26: I got a lint error when I included --- in the file. Not sure why, but when I removed it, the error when away. Maybe we should remove the --- from both tutorials

docs/vendor/tutorial-ui-create-new-version.md
```diff
@@ -1,34 +1,102 @@
...
+    This creates a reference to the `more_text` field using a Replicated template function. The ConfigOption template function renders the user input from the configuration item that you specify. For more information, see [Config Context](/reference/template-functions-config-context) in _Reference_.
+
+1. Check the KOTS linter messages in the right pane to ensure there are no errors.
 
 1. Click **Save Release**.
 
-1. Following the same process you before, promote the release to a channel:
+1. Promote the release to a channel:
 
     1. Click **Promote** next to the newly created Sequence 2.
```
paigecalvert, line 94: I didn't see "sequence 2" on the page. here's what I see for the promote button at this point:
Screen Shot 2022-11-11 at 10 46 45 AM

paigecalvert: If I click back to the Release page though I can see that this is labeled as Sequence 2

docs/vendor/tutorial-ui-create-new-version.md
```diff
@@ -1,34 +1,102 @@
...
 
     1. Choose the Unstable channel, and click **Promote**.
 
   Any license installed from the Unstable channel will start with this new release, and any installation already running is prompted to update to the new release.
```
paigecalvert, line 98: Looks like this was here previously, but I don't think all existing installations would prompt to update to the new release. I had to click "check for update" on the dashboard (this might be different is automatic updates were configured)

docs/vendor/tutorial-ui-update-app.md
```diff
@@ -1,20 +1,51 @@
...
+  The application displays the text that you added to the field.
+
+  :::note
+  If you do not see the new text, refresh your browser.
+  :::
 
 ## Next Steps
 
-Now that you are familiar with the basics, we recommend that you run through the [CLI Tutorial](tutorial-cli-setup) to start managing your release YAML in a git repository.
+Congratulations! As part of this tutorial, you:
```
paigecalvert, line 39: You could add that they created a preflight check as well!

paigecalvert (commented):
Looks great @jonquil2002 ! I noted some places where what I saw differed slightly from the steps, and a few other suggestions.

I was thinking (maybe for the next iteration) we could pair on some of the details that appear in both tutorials to make sure it's all super consistent.

docs/vendor/tutorial-ui-install-app-manager.md
```diff
@@ -2,42 +2,61 @@ import VMRequirements from "../partials/getting-started/_vm-requirements.mdx"
...
-Installing the app manager on existing clusters is similar to using a Kubernetes installer provisioned cluster, except instead of bringing a plain virtual machine (VM), you will use a pre-built Kubernetes cluster and deploy your application into a namespace.
+When you install the app manager on existing cluster, you use a pre-built Kubernetes cluster and deploy your application into a namespace.
+
+You can see the installation script options at the bottom of each channel on the Channels page in the vendor portal.
 
 ![Installation Methods](/images/guides/kots/installation-methods-existing.png)
 
 To install the app manager:
 
 1. Access the Kubernetes cluster that you created as part of [Set Up the Environment](tutorial-ui-setup#set-up-the-environment).
```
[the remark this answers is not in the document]
jonquil2002: I can add that link. For this tutorial, they can use the VM, so they do need to SSH into the cluster, so we need to keep this step and maybe just clarify it. I'll be happy to add that link. Thanks!

docs/vendor/tutorial-ui-create-new-version.md
jonquil2002: Yeah, I copy/pasted that from the CLI tutorial but didn't get an error when I did the CLI tutorial. But I'll delete `---` from this tutorial. Thanks for testing it.

docs/vendor/tutorial-ui-create-new-version.md
jonquil2002: That's leftover language from the old tutorial and might be different if automatic updates are configured. I'll adjust the language.

docs/partials/getting-started/_vm-requirements.mdx
```diff
@@ -1,4 +1,4 @@
-The VM must meet the following requirements:
+The VM must meet the following minimum requirements:
```
paigecalvert, line 1: I don't think this should say minimum, since it does have to be Ubuntu 18.04 (not later). Might be better to clarify in the bullet points themselves when the value is minimum

docs/vendor/tutorial-ui-create-app.md
```diff
@@ -1,5 +1,7 @@
...
 
+When you create an application object in the vendor portal, example NGINX manifest files are included that will be used for this tutorial.
+
 :::note
 Ensure that you have completed the required environment setup. See [Set Up the Environment](tutorial-ui-setup#set-up-the-environment) in _Introduction and Setup_.
 :::
```
paigecalvert, line 7: I ended up removing this note from the CLI tutorial based on your note about it breaking up the flow. Linking to the Setup topic from the relevant step in the procedure itself seemed to cover it.

docs/vendor/tutorial-ui-setup.mdx
```diff
@@ -18,19 +18,24 @@ For a tutorial that demonstrates how to use the replicated CLI to create and pro
 
 As part of this tutorial, you install a sample application onto an existing Kubernetes cluster, or onto a cluster provisioned by the Replicated Kubernetes installer. The Kubernetes installer creates the cluster on an existing VM.
 
-Before you begin this tutorial, ensure that you have one of the following:
+Before you begin this tutorial, ensure that you have one of the following environments set up:
 
-* A Kubernetes cluster that you can access with the kubectl command line tool.
+- A Kubernetes cluster that you can access with the kubectl command-line tool. For more information about minimum requirements for installations, see [Installation Requirements](/enterprise/installing-general-requirements).
```
paigecalvert, line 23: There might need to be a stronger segue between the first sentence and the xref. It wasn't clear to me if it is necessary that the cluster meet the requirements listed in that link or not, or if it's provided simply for more info.

(I think the mention of "min requirements _for installations_" in particular made me wonder if it applied to me going through this tutorial or not)

Also, recommend linking specifically to the Existing Cluster section: enterprise/installing-general-requirements#existing-cluster-requirements

docs/vendor/tutorial-ui-setup.mdx
```diff
@@ -18,19 +18,24 @@ For a tutorial that demonstrates how to use the replicated CLI to create and pro
...
-* A VM where the Kubernetes installer can create a cluster.
+    ```
+    gcloud container clusters create NAME --preemptible --no-enable-ip-alias
+    ```
+    Replace `NAME` with any name you want to use for the cluster.
 
-  <VMRequirements/>
+  :::note
+  For information about installing and configuring the kubectl command-line tool, see [Install Tools](https://kubernetes.io/docs/tasks/tools/) and [Command line tool (kubectl)](https://kubernetes.io/docs/reference/kubectl/) in the Kubernetes documentation.
+  :::
```
paigecalvert, line 34: I think this could just be plain text rather than a note

docs/vendor/tutorial-ui-setup.mdx
```diff
@@ -18,19 +18,24 @@ For a tutorial that demonstrates how to use the replicated CLI to create and pro
...
+
+- A VM where the Kubernetes installer can create a cluster. The Kubernetes installer script installs and configures the kubectl command-line tool for you. For more information about minimum requirements for installations, see [Installation Requirements](/enterprise/installing-general-requirements).
+
```
paigecalvert, line 37: Recommend removing the link for more information since the partial here displays all the relevant info inline (unless there is something else specifically in the link that they'd need to be looking for)

docs/vendor/tutorial-ui-install-app-manager.md
```diff
@@ -2,43 +2,48 @@ import VMRequirements from "../partials/getting-started/_vm-requirements.mdx"
...
+
+Choose either installation option:
 * [Install in an Existing Cluster](#existing)
 * [Install in a Kubernetes Installer Cluster](#kubernetes-installer)
 
-## Install in an Existing Cluster {#existing}
+## Prerequisite
 
-Installing the app manager on existing clusters is similar to using a Kubernetes installer provisioned cluster, except instead of bringing a plain virtual machine (VM), you will use a pre-built Kubernetes cluster and deploy your application into a namespace.
+This step requires you to have kubectl access to a Kubernetes cluster. Before you begin, ensure that you have completed the required setup steps to create the cluster and configure kubectl command line access. See [Set Up the Environment](tutorial-ui-setup) in _Introduction and Setup_.
```
paigecalvert, line 17: Looks like this prereq only applies to the Existing Cluster procedure:
- Could nest it under there
- Could update to include info about both existing and embedded
- Could remove (ended up leaving this out of the CLI tutorial)

docs/vendor/tutorial-ui-install-app-manager.md
```diff
@@ -2,43 +2,48 @@ import VMRequirements from "../partials/getting-started/_vm-requirements.mdx"
...
+![Installation Methods](/images/guides/kots/installation-methods-existing.png)
 
-  For example, to set the local kubectl context on a GKE cluster, run the following command:
+To install the app manager:
 
-  ```shell
-  gcloud container clusters get-credentials NAME
-  ```
-  Where `NAME` is the name of the cluster.
+1. Copy the **KOTS Install** command from the Unstable channel in the vendor portal and run it on the cluster. The script installs the latest app manager version as a `kubectl` plugin. For more information about installing an application with the kots CLI, see [install](../reference/kots-cli-install/) in the kots CLI documentation.
```
paigecalvert, line 29: Small part that's inaccurate here is "and run it on the cluster". You don't run commands on a cluster. Rather, to interact with clusters, you use the kubectl command line tool. You run kubectl commands on your command line.

So instead, this would be something like "make sure the kubectl context is set to your existing cluster, then run the following command on your command line:"

docs/vendor/tutorial-ui-install-app-manager.md
```diff
@@ -49,47 +54,56 @@ To install the app manager:
     • Go to http://localhost:8800 to access the Admin Console
   ```
```
paigecalvert, line 55: Suggestion: Could add a sentence at the end of this procedure to tell them to go to the Next Step section at the bottom, since it might not be clear with the embedded cluster procedure immediately following it

docs/vendor/tutorial-ui-create-new-version.md
```diff
@@ -1,34 +1,101 @@
...
 
-1. Following the same process you before, promote the release to a channel:
+1. Promote the release to a channel:
+
+    1. Click **Promote** at the top of the page.
+
+    1. Choose the Unstable channel in the Promote Release dialog, and click **Promote**.
 
-    1. Click **Promote** next to the newly created Sequence 2.
+  Any license installed from the Unstable channel will start with this new release, and any installation already running is prompted to update to the new release when the admin console checks for updates. However, in the next step, you will check for the update manually.
```
paigecalvert, line 97: Suggestion: Might be able to get rid of the "However," sentence. The "is prompted to update to the new release when the admin console checks for updates" does not specify whether that's a manual or automatic update, so probably safe to assume either

If you leave the sentence in, I recommend removing the word "However" because it seemed to suggest that we were not doing things the normal way in this tutorial, without being sure about which parts in particular differ from the norm.

paigecalvert (approved):
Looking very good; pretty small comments!

docs/partials/getting-started/_vm-requirements.mdx
jonquil2002: Ok, good point. And there's no real reason that I know of to add "minimum" anywhere right now, so I'll just delete it.

docs/vendor/tutorial-ui-setup.mdx
jonquil2002: I'll remove it for now. But, for production clusters, the Kubernetes installer minimum requirements are more extensive and I'm not sure whether that's worth bringing to their attention now or not. This is the section that I'm talking about, but we can revisit that decision in the future if needed: https://docs.replicated.com/enterprise/installing-general-requirements#embedded-cluster-requirements

docs/vendor/tutorial-ui-setup.mdx
jonquil2002: I removed it and made it mirror your Setup wording.

docs/vendor/tutorial-ui-install-app-manager.md
jonquil2002: kubectl is used in the embedded cluster too, but the Kubernetes installer installs and configures the context for them. 

I'll leave this out since you left it out of the CLI tutorial.

docs/vendor/tutorial-ui-install-app-manager.md
jonquil2002: You're right - "run it on the cluster" is our internal slang.

jonquil2002: Also, in adding that clarification, the sentence got really long. I broke this into a second paragraph: "The script installs the latest app manager version as a `kubectl` plugin. For more information about installing an application with the kots CLI, see [install](../reference/kots-cli-install/) in the kots CLI documentation."

It could also go after the installation command example as a results step, but the next steps are a continuation of the installation, so it kinda doesn't belong as a result. It's really explaining what will happen - and maybe this sentence isn't even necessary. We could revisit that as a future story.

paigecalvert (approved):
Pushed a minor wording edit to the step about running the install command on an existing cluster and approved!

PR #768, opened 2022-11-11
Update policies-support-lifecycle.md

AmberAlston:
Update to add established timeline for legacy native product


paigecalvert (approved):
thanks @AmberAlston

PR #769, opened 2022-11-11
Clarify namespace description

paigecalvert:
https://app.shortcut.com/replicated/story/60791/clarify-the-description-of-the-namespace-field-in-helmchart-custom-resource

### DIFF
--- a/docs/reference/custom-resource-helmchart.md
+++ b/docs/reference/custom-resource-helmchart.md
@@ -279,9 +279,11 @@ In this case, all keys from `values` and `optionalValues` are included in the me
 
 ## namespace
 
-The `namespace` key allows for a chart to be installed in an alternate namespace.
-If left blank, the namespace will default to the one into which the admin console is installed.
-If an alternate namespace is specified, it is required that the namespace exist or is included in the manifest file for the HelmChart custom resource.
+The `namespace` key specifies an alternative namespace where the app manager installs the Helm chart. By default, if no alternate namespace is provided, then the Helm chart is installed to the same namespace as the admin console.
+
+If you specify an alternate namespace in the `namespace` field, either Helm or your end user must create the namespace in the cluster:
+* **(Recommended) Configure Helm to create the namespace**: To configure Helm to create the namespace, include a Namespace object in the Release object of the Helm chart. For more information, see [Built-in objects](https://helm.sh/docs/chart_template_guide/builtin_objects/) in the Helm documentation.
+* **Instruct the end user to create the namespace**: Alternatively, the end user can run the `kubectl create namespace` command before they begin the installation process to create the namespace. For more information, see [Creating a new namespace](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#creating-a-new-namespace) in the Kubernetes documentation.
 
 ## builder
 


docs/reference/custom-resource-helmchart.md
```diff
@@ -279,9 +279,11 @@ In this case, all keys from `values` and `optionalValues` are included in the me
 
 ## namespace
 
-The `namespace` key allows for a chart to be installed in an alternate namespace.
-If left blank, the namespace will default to the one into which the admin console is installed.
-If an alternate namespace is specified, it is required that the namespace exist or is included in the manifest file for the HelmChart custom resource.
+The `namespace` key specifies an alternative namespace where the app manager installs the Helm chart. By default, if no alternate namespace is provided, then the Helm chart is installed to the same namespace as the admin console.
+
+If you specify an alternate namespace in the `namespace` field, either Helm or your end user must create the namespace in the cluster:
+* **(Recommended) Configure Helm to create the namespace**: To configure Helm to create the namespace, include a Namespace object in the Release object of the Helm chart. For more information, see [Built-in objects](https://helm.sh/docs/chart_template_guide/builtin_objects/) in the Helm documentation.
```
chris-sanders, line 285: The "Release Object" that is being referenced here is available for templating it's not something you can (to my knowledge) include something in.

The path to have Helm create it then is to include a Namespace object in your Helm Chart. For example you would include the following in your Helm Chart with the rest of your manifests:
```
apiVersion: v1
kind: Namespace
metadata:
  name: {{ Release.Namespace }}
```

I'm working from memory here I think Helm 3 will prioritize creating a namespace if you do this. I typically use features of ArgoCD which uses the flag available in Helm to [create the namespace](https://helm.sh/docs/faq/changes_since_helm2/#automatically-creating-namespaces) so someone from the Helm installer team should be consulted to confirm the above works or verify how we recommend you create the namespace with a helm based install.

I'm showing it templated, if the end user isn't suppose to make the choice you can also hardcode your namespace above, again someone from the Helm team should probably say which we're recommending.
Example:
```
apiVersion: v1
kind: Namespace
metadata:
  name: my-awesom-application-namespace
```

chris-sanders (changes requested):
Comment inline

docs/reference/custom-resource-helmchart.md
ajp-io: I agree that we should have someone from our team comment on how this should be done. From some quick reading, I'm not positive that you can include a namespace object in the release and have it get created. I'm seeing more mentions of the `--create-namespace` flag. We do support [`helmUpgradeFlags`](https://docs.replicated.com/reference/custom-resource-helmchart#helmupgradeflags), which someone could use to pass that flag. But again, an engineer should comment on which way is the right way to do this here.

docs/reference/custom-resource-helmchart.md
```diff
@@ -279,9 +279,11 @@ In this case, all keys from `values` and `optionalValues` are included in the me
...
+* **Instruct the end user to create the namespace**: Alternatively, the end user can run the `kubectl create namespace` command before they begin the installation process to create the namespace. For more information, see [Creating a new namespace](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#creating-a-new-namespace) in the Kubernetes documentation.
```
ajp-io, line 286: If there is a good automated solution above, I think we should only document that one. No need to mention a more manual approach with a second command if a single command can do it. But that depends on what is possible for the above approach.

PR #777, opened 2022-11-16
62304: Remove KOTS from KOTS Lint Service

jonquil2002:
SC: https://app.shortcut.com/replicated/story/62304/remove-kots-from-kots-lint-service


jonquil2002:
Throwing this branch away bc something went wonky with the commit. Starting over in a new PR.

PR #779, opened 2022-11-16
Fix inaccuracy in headless install

paigecalvert:
Story: https://app.shortcut.com/replicated/story/46286/confusing-directions-in-headless-installation

Fixing inaccurate statement that says you can access the admin console at port 8800 after running with --no-port-forward. Instead, you can run `kubectl kots admin-console --namespace NAMESPACE` to start the admin console if you want

Preview: https://deploy-preview-779--replicated-docs.netlify.app/enterprise/installing-existing-cluster-automation#disable-admin-console-port-forwarding

### DIFF
--- a/docs/enterprise/installing-existing-cluster-automation.md
+++ b/docs/enterprise/installing-existing-cluster-automation.md
@@ -90,7 +90,7 @@ kubectl kots install app-name \
   --no-port-forward
 ```
 
-After this completes, you can optionally run the following command to start the admin console:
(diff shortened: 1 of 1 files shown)


paigecalvert:
@jonquil2002 quick fix here on the inaccurate statement in the headless install example that's ready for your review.

(After looking into it more, there's lots more to be done to fix this topic and determine the best way to present information about installing with the CLI, but we can do that as part of spin-offs.)

jonquil2002 (approved):
lgtm

PR #780, opened 2022-11-17
Update notes on Redactors

xavpaice:
Change https://github.com/replicatedhq/troubleshoot/pull/734 removes IPv4 address redaction by default.  This change removes that promise from the document, plus adds a little more disclaimer type info on the remaining redactors since the promise in the doc is more broad than the actual redactors can cover.


paigecalvert (approved):
Did some rewording and also found another mention of IP addresses in the docs that I removed. LMK if the edits look okay to you

PR #782, opened 2022-11-17
note vendors must install EKCO in embedded clusters

adamancini:


paigecalvert (approved):
Thank you @adamancini !

@jonquil2002 , do these edits make sense to you? Also, could be beneficial for us to pair on a sustainable/longer-term way to get all the Kubernetes installer requirements and recommendations into the docs, but this feels definitely feels like a good addition regardless

adamancini:
@paigecalvert @jonquil2002 yeah I think we can work on a longer doc for the "how to choose" that should help in the meantime until Replicated's use of kURL becomes more opinionated

paigecalvert:
@adamancini yes that would be great! One of the updates "on deck" in this area is to provide clearer recommendations as to what data storage and data replication add-ons to use based on various use cases

paigecalvert:
Mergin'!

PR #788, opened 2022-11-18
Add info about ConfigValues file

paigecalvert:
Story: https://app.shortcut.com/replicated/story/46383/document-how-to-create-a-configvalues-file

Adds info for vendors about:
- What is the ConfigValues file
- Why do you/your users need it
- How to download it to create a template for your users

Updates info for enterprise users about how to define a ConfigValues file

Previews: 
- (New) Downloading the ConfigValues File https://deploy-preview-788--replicated-docs.netlify.app/vendor/releases-configvalues
- Using Automation to Install (general clean-up related to these changes) https://deploy-preview-788--replicated-docs.netlify.app/enterprise/installing-existing-cluster-automation
- Add recommended step to Distributing workflow https://deploy-preview-788--replicated-docs.netlify.app/vendor/distributing-workflow


paigecalvert:
@jonquil2002 this is ready for your to review when you are back next week. Some additional context for the review:

I ended up needing to make some updates to the Using Automation to Install topic as part of the work to add information about the ConfigValues file. In particular, I wanted to clarify that the user should be getting details from their vendor on how to make a ConfigValues file, since it's app specific.

   I made some fairly large organization updates to this topic because I realized as part of this that the different headings didn't make sense. There is still a lot to be done with this topic/this information in general, so I didn't want to spend too much time as part of this story on it when we are likely going to want to make more foundational changes. (For example, we might decide that we don't want a separate topic for kots CLI installs at all, and that the info should just be incorporated in existing procedures.)
   
  So in terms of reviewing the Using Automation to Install topic in particular, I recommend focusing more on the sentence level stuff than the higher level organizational stuff, though of course welcome any of your thoughts and suggestions!

docs/enterprise/installing-existing-cluster-automation.md
```diff
@@ -1,116 +1,108 @@
...
 
-The app manager runs preflight checks (conformance tests) against the target namespace and cluster to ensure that the environment meets the minimum requirements to support the application.
+## Air Gap Installation
 
-![Preflight Checks](/images/preflight-checks.png)
+Before you can install an application in an air gap environment, you must push admin console images to a private registry using the  `kubectl kots admin-console push-images` command. For more information, see [Install in an Air Gapped Environment](installing-existing-cluster#air-gap) in _Installing on an Existing Cluster_.
 
-#### Resolve strict preflight checks
+To install an application with the kots CLI in an air gap environment, provide the following with the `kots install` command:
+* The license file
```
jonquil2002, line 43: You could, if you wanted, remove "The" and "A" from the start of each bullet in the list, if you feel you can do it and keep a good flow.

jonquil2002 (dismissed):
lgtm, and only on very minor suggestion.

Thanks for the guidance on what to focus on for the review. Looks great!

paigecalvert:
@jonquil2002 made lots of edits to both topics based on our convo this morning. Would love to get another round of feedback from you before I find me an SME to confirm some deats (waiting for responses on the thread I opened in AMA https://replicated.slack.com/archives/C036BTS7JCE/p1669754311916749)

docs/enterprise/installing-existing-cluster-automation.md
```diff
@@ -1,116 +1,85 @@
...
+import ConfigValuesExample from "../partials/configValues/_configValuesExample.mdx"
 
-When these values are provided, they are written as ConfigMaps to the cluster, and the Replicated admin console finds these and processes them to complete the installation.
+# Using Automation to Install on an Existing Cluster
 
-## Online Installation
+This topic describes installing an application with the kots CLI in online and air gap environments.
 
-### Provide a License File
+## Installing in an Online Environment
```
paigecalvert, line 7: Figured I might as well rework the headings to (roughly) align with the style guide while I was here...

docs/enterprise/installing-existing-cluster-automation.md
```diff
@@ -1,116 +1,85 @@
...
-```shell
-kubectl kots admin-console -n <your app namespace>
-```
+As shown in the example above, the ConfigValues file includes the names of the configuration fields for the application, along with a user-supplied value for each field.
 
-## Example
+Your application vendor provides details about the required and optional configuration fields to include in the ConfigValues file. For more information, see [Downloading the ConfigValues File](/vendor/releases-configvalues).
 
-Given the information above, and a config file named `configvalues.yaml`, a license file named `license.yaml`, you could use the following command to automate an application installation:
+## About Disabling the Port Forward
```
paigecalvert, line 74: Moved this down here since it's the same for both sections, like the ConfigValues content is. 

Need to confirm with someone that --no-port-forward isn't required for air gap, or something weird like that

docs/enterprise/installing-existing-cluster-automation.md
```diff
@@ -1,116 +1,85 @@
...
-You can later access the admin console with the following command:
+<ConfigValuesExample/>
 
-```shell
-kubectl kots admin-console -n <your app namespace>
-```
+As shown in the example above, the ConfigValues file includes the names of the configuration fields for the application, along with a user-supplied value for each field.
 
-## Example
+Your application vendor provides details about the required and optional configuration fields to include in the ConfigValues file. For more information, see [Downloading the ConfigValues File](/vendor/releases-configvalues).
```
paigecalvert, line 72: On second thought, it seemed better to just say that the vendor provides this, then link to the other topic for more info. I couldn't really think of a good use case for specifically telling the enterprise user to follow the same steps to download the configvalues file from an installed instance of the app

docs/vendor/releases-configvalues.md
```diff
@@ -0,0 +1,53 @@
...
+
+<ConfigValuesExample/>
+
+As shown in the example above, the ConfigValues file includes the names of configuration fields along with a user-supplied value for each field. Where applicable, the ConfigValues file can also include the default values of the configuration fields.
+
+In the ConfigValues file, users can include any optional or required configuration fields that you defined in the Config custom resource for the release. For any required configuration fields that do not have a default value, users must provide a value in the ConfigValues file to install the application.
+
+For more information about how to create custom configuration fields for your application in the Config custom resource, see [Config](/reference/custom-resource-config) in _Reference_.
+
+## Download the ConfigValues File
```
paigecalvert, line 23: Something else I want to confirm (putting here so I don't forget) is whether or not asking vendors to go download the file is actually the easiest user experience. 

Because alternatively we could just describe exactly how to make the file (go to your Config custom resource. Copy the field names. Paste them in a ConfigValues file. Add a `default` field if there's a default. Add a `values` field for the user to enter a value, or if the value of the field needs to be encrypted, make it a `valuePlaintext` field instead).

paigecalvert: https://replicated.slack.com/archives/CNRGJLDH6/p1669771587831549

docs/enterprise/installing-existing-cluster-automation.md
jonquil2002: Sounds like a good idea.

jonquil2002: The only downside is that it links back to the Vendor docs, but we expect that the vendor will customize our Enterprise docs for their users anyway.

jonquil2002 (dismissed):
lgtm -nice job!

docs/enterprise/installing-existing-cluster-automation.md
paigecalvert: Yeah, I was thinking the same thing. This was basically just if a vendor was reading this enterprise doc and was wondering how to do it

docs/vendor/releases-configvalues.md
```diff
@@ -0,0 +1,85 @@
...
+
+During application installation, the Replicated app manager automatically generates a ConfigValues file and saves the file in a directory called `upstream`. After installing an application, you can view this generated ConfigValues file in the admin console View files page. You can also download the file using the kots CLI.
+
+To allow your users to install and configure your application with the kots CLI, Replicated recommends that you provide your users with an example of an accurate ConfigValues file that they can copy and paste. To create this example, you can use the ConfigValues file generated by the app manager for an instance of your application that you installed in a development environment.
```
ajp-io, line 27: I think some language around tweaking the file could be more accurate than copy and paste. I get where you're going, I just think the copy and paste part is obvious, it's the tweaking part that the enterprise really needs to do.

docs/vendor/releases-configvalues.md
```diff
@@ -0,0 +1,85 @@
...
+          value: "1"
+        readonly_example:
+          value: Xp1l3JHds2
+        hidden_example:
+          value: sIpE9njpy0SWyw9Ht6K+rPzeke1DjG619Lnv4radKJS7lFUfmdzi2+MIJEj33k8PRyYM/eAGtes=    
+    status: {}
+   ```
+   As shown in the example above, the ConfigValue file generated by the app manager:
+   * Lists the `default` value for configuration fields where applicable.
+   * Includes configuration fields that are not editable by the user, such as `readonly` and `hidden` fields. For more information, see [readonly](/reference/custom-resource-config#readonly) and [hidden](/reference/custom-resource-config#hidden) in _Config_.
```
ajp-io, line 77: What should an enterprise do with these fields? There's not indication that these are readonly and hidden fields, except the names you gave them. We should probably specify whether vendors should keep these in their sample and advise their customers to not change them, or what. I don't know the actual behavior here. For example, if it's readonly, does KOTS just ignore it not matter what you put in that field?

docs/vendor/releases-configvalues.md
```diff
@@ -0,0 +1,85 @@
...
+   * Includes `status` and `metadata` fields, which are generated because the file is a Kubernetes custom resource. The app manager does not use the `metadata` or `status` fields.
+   * Decrypts the value for configuration fields with `type: password` in a `valuePlaintext` field when the `--decrypt-password-values` flag is included with the `kots download` command.
+
+1. Copy and paste the contents of the ConfigValues file to a new file to create an example for your users. Consider the following recommendations for the example ConfigValues file that you create:
+   * Indicate which configuration fields are required and optional. For any required configuration fields that do not have a default value, users must provide a value in the ConfigValues file to install the application.
+   * Where applicable, provide the default values for the configuration fields. In the ConfigValues file generated by the app manager, the default values are listed in the `default` fields. User-supplied values in `value` fields overwrite any `default` values.
```
ajp-io, line 83: What happens if a value is not required and the vendor didn't specify a value for it when they set up their test app? Does the ConfigValues file they download include that value or not? I wonder if it doesn't since that value isn't set.

If that's the case, they should make sure that every field is set when they download the file, and then they should instruct their customers to remove that field from the sample if they don't want to set that value. Or something like that.

docs/vendor/distributing-workflow.md
```diff
@@ -144,3 +144,5 @@ To distribute your production application:
 1. Promote the release to the desired channel in the vendor portal. You can use the default channels or edit them, or create a custom channel. See [Creating and Editing Channels](releases-creating-channels).
 
 1. Share with each customer the license file that you created and the installation command available in the vendor portal. See [Sharing the License Files and Application Releases](releases-sharing-license-install-script).
+
+1. (Recommended) To support installations with the kots CLI, provide your users with a template of the ConfigValues file for the release. See [Downloading the ConfigValues File](/vendor/releases-configvalues).
```
ajp-io, line 148: Should this be above the prior point since you would want to have the sample ConfigValues file before you send the install command?

ajp-io (commented):
Only viewed one file last time. My bad! Here is one other comment.

docs/vendor/releases-configvalues.md
paigecalvert: > What happens if a value is not required and the vendor didn't specify a value for it when they set up their test app?

It includes it like this: `my_value: {}`

> and then they should instruct their customers to remove that field from the sample if they don't want to set that value.

It looks like I can leave it as `my_value: {}` to leave it unset (or remove)

That said...

I'm beginning to wonder if the "download the ConfigValues file as a sample" approach is the cleanest/easiest way. Alternatively, we could tell them how to construct the file using their Config fields, because it's pretty simple. Rather than saying "delete this and delete that and make sure there's a fake value in all your fields" etc. It could just be, for every field you want to edit, have customers either: 1) put it in the file with this format or 2) remove it from the file if they don't want to set it/change the default

What do you think?

docs/vendor/releases-configvalues.md
paigecalvert: > For example, if it's readonly, does KOTS just ignore it not matter what you put in that field?

Yeah, KOTS does not change it. (Unless it's a RandomString function, then KOTS generates a random string that is not affected by whatever the user puts in the value: field in the ConfigFile). 

Hidden can be changed (which maybe is a feature not a bug). I think it has the same behavior as outlined here with the different combos of hidden and readonly set to true versus false: https://docs.replicated.com/reference/template-functions-static-context#randomstring

But either way, I did forget to mention what to do with these down in the procedure. (This is related to comment below though about, what if telling vendors to download a sample file is not the easiest approach: https://github.com/replicatedhq/replicated-docs/pull/788#discussion_r1038429318)

ajp-io:
Given our conversation on Zoom earlier, I'm going to leave this to you for now, unless you think there's something else for me to re-review. Looks like your comments are either acknowledgements or related to telling vendors not to do the download approach, which we decided against.

paigecalvert:
Yep sounds good to me @ajp-io . Just pushed the changes we talked about (decided to include a new table that lists out the supported YAML format for each config item type to make it easier to parse). Feel free to take another pass if you'd like!

docs/vendor/releases-configvalues.md
ajp-io: Did you do anything with this? Looks the same to me at a glance. Or maybe you're okay with the wording.

paigecalvert: Oh, weird. I think I was just confused about where the comment was or something. I removed the copy/paste language from somewhere else in the doc, but must have missed this paragraph.

paigecalvert: removed!

ajp-io (approved):

PR #792, opened 2022-11-22
Add info about deleting channels

paigecalvert:
Previews:
- Added a Delete a Channel section: https://deploy-preview-792--replicated-docs.netlify.app/vendor/releases-creating-channels#delete-a-channel
- Updated the screenshots for the creating and editing a channel procedures because I saw they were out of date: https://deploy-preview-792--replicated-docs.netlify.app/vendor/releases-creating-channels#create-a-channel (also updated the procedures accordingly)

Story: https://app.shortcut.com/replicated/story/62558/add-details-about-using-default-channels

### DIFF
--- a/docs/vendor/releases-creating-channels.md
+++ b/docs/vendor/releases-creating-channels.md
@@ -1,6 +1,6 @@
 # Creating and Editing Channels
 
-There are three default channels that can be edited or deleted. You can also add custom channels.
+There are three default channels that can be edited or archived. You can also add custom channels.
 
 The default channels are:
 * **Stable**: The Stable channel is intended for production installations.
@@ -38,7 +38,7 @@ To edit an existing channel:
 
 1. Edit the fields and click **Save**.
 
-## Delete a Channel
+## Archive a Channel
 
 You can archive an existing channel to prevent any new releases from being promoted to the channel.
 
@@ -52,9 +52,9 @@ Assign customers to a different channel and set a different channel as the defau
 
 To archive a channel with the vendor portal or the replicated CLI:
 
-* **Vendor portal**: In the vendor portal, go to the **Channels** page and click the trashcan icon in the top right corner of the card for the channel that you want to delete.
+* **Vendor portal**: In the vendor portal, go to the **Channels** page and click the trash can icon in the top right corner of the card for the channel that you want to archive.
 * **replicated CLI**:
-  1. Run the following command to find the ID for the channel that you want to delete:
+  1. Run the following command to find the ID for the channel that you want to archive:
      ```
      replicated channel ls
      ```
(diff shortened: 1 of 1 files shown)


docs/vendor/releases-creating-channels.md
```diff
@@ -1,6 +1,6 @@
 # Creating and Editing Channels
 
-There are three default channels that can be configured, and you can add custom channels.
+There are three default channels that can be edited or deleted. You can also add custom channels.
```
alicenstar, line 3: I wonder if we should say "archive" instead of delete, since that what the "delete" button says in Vendor Portal.

docs/vendor/releases-creating-channels.md
```diff
@@ -36,13 +36,37 @@ To edit an existing channel:
...
+You cannot archive a channel if:
+* There are customers assigned to the channel.
+* The channel is set as the default channel.
+
+Assign customers to a different channel and set a different channel as the default before archiving.
+:::
+
+To archive a channel with the vendor portal or the replicated CLI:
+
+* **Vendor portal**: In the vendor portal, go to the **Channels** page and click the trashcan icon in the top right corner of the card for the channel that you want to delete.
```
alicenstar, line 55: I think there should be a space in "trashcan".

alicenstar: And again, open to your opinion, but I think instead of delete, we should standardize language to say "archive".

docs/vendor/releases-creating-channels.md
```diff
@@ -36,13 +36,37 @@ To edit an existing channel:
...
+     ```
+     replicated channel ls
+     ```
+     The output of this command includes the ID and name for each channel, as well as information about the latest release version on the channels.
+
+  1. Run the following command to archive the channel:
+     ```
+     replicated channel rm CHANNEL_ID
+     ```
+     Replace `CHANNEL_ID` with the ID for the channel.
```
alicenstar, line 67: Would it be helpful to add something like "You can find the channel ID on the edit channel modal in the Vendor Portal"? Or otherwise some info on how to find the channel ID? Not sure if that's too much to put here.

alicenstar (approved):

docs/vendor/releases-creating-channels.md
paigecalvert: Archive works for me. I'll edit

docs/vendor/releases-creating-channels.md
paigecalvert: I was thinking the same thing. Since this is a CLI procedure, I made the first step be to run `channel ls` so that you can get the channel ID. Do you think that works or do you think adding the vendor portal steps too would be helpful?

paigecalvert: Added some language here to clarify that it's the channel ID that you retrieved in the previous step. LMK what you think

alicenstar (approved):
@paigecalvert Looks great to me!

PR #793, opened 2022-11-24
Update helm-native-helm-install-order.md remove quotes from weight setting in docs

grahamnscp:
Think that the weight number gives an error if it's in quotes


paigecalvert (approved):
Thank you @grahamnscp ! I also see that the example here does not use quotes for the weight field: https://docs.replicated.com/reference/custom-resource-helmchart

PR #807, opened 2022-12-05
clarify the term "Collab Repository" before usage.

dexhorthy:
Not sure if this is the right place for this content, but it seemed like we were assuming a lot of context that some folks may not have. If y'all wanna put this in as a `<note>` or something else I'd be open to that as well.


paigecalvert (approved):
Thanks @dexhorthy --I see what you are getting at.

I updated our description of what the collab repo is to include the link that you provided. I also removed the term "collab repository" from the topic sentence so that users are not unnecessarily introduced to that terminology before they make it to the Overview section.

I wanted to keep the detailed conceptual description in the Overview section and also avoid duplicating that content above. This way, users can keep the lightweight context of what the topic covers as the first thing they read, then can jump into the Overview in the next sentence to learn more.

paigecalvert:
Going to merge this @dexhorthy !

PR #816, opened 2022-12-08
Update policies-support-lifecycle.md

AmberAlston:
Adds date extension asterisk to SWARM to reflect one or more customers has contracted a different timeline beyond the 12-31-22 EOL timeline for all other SWARM users.


paigecalvert (approved):
thanks @AmberAlston !

PR #819, opened 2022-12-09
add known airgap config issue

cbodonnell:


paigecalvert (approved):
Very nice @cbodonnell , thanks for adding this. I figure @ajp-io is still reviewing before we merge, but I added some minor style edits.

I also added another suggestion that they can ensure all required configuration items have a default if they want to upgrade to 1.91.2. (I was thinking that if people _do_ have default values for all required items for their app, then it looks like it's safe for them to upgrade. So this would reinforce that.) Not sure if we want that suggestion in though, so feel free to delete if we'd rather just recommend no one upgrade.

ajp-io:
I think your suggestion is fine. I approved.

ajp-io (approved):

PR #824, opened 2022-12-12
63235: Changes to Creating Releases Topic

jonquil2002:
SC: https://app.shortcut.com/replicated/story/63235/combine-creating-and-promoting-releases-into-one

Previews: 
- https://deploy-preview-824--replicated-docs.netlify.app/vendor/distributing-workflow
- https://deploy-preview-824--replicated-docs.netlify.app/vendor/releases-creating-releases
- https://deploy-preview-824--replicated-docs.netlify.app/reference/replicated-cli-release-promote
- https://deploy-preview-824--replicated-docs.netlify.app/release-notes/rn-app-manager#1680


docs/vendor/distributing-workflow.md
```diff
@@ -65,77 +65,67 @@ Complete the following procedures to import your files, create, and test your fi
...
-      </tr>
-   </table>
-
-  :::note
-  Continue to iterate by updating the release and application, and then testing it in the admin console until you are ready to share the application with your customers.
-  :::
+To iterate on the release of your production application, configure additional manifest functionality and custom resources, and iterate as needed. This is a suggested iteration order, but you can configure these functions in any order. These functions can be used with Kubernetes Operators, except where noted.
+
+Update the release and test it in your development environment after each iteration until you are ready to share the application with your customers.
+
```
paigecalvert, line 77: I like the addition of this description. Only recommendation is to make sure that it's clear the difference between "updating a release" and "iterating over multiple releases".

Here's how I think I understand it (or LMK if this doesn't sound right):

_Iterating_ means that you create and promote several releases.

_Updating a release_ means that the release has not yet been promoted, and you are making edits to the YAML files. For example, if I'm working from the replicated CLI, I might run `replicated release update` if I made an edit to a YAML file that I want my team to be able to see in the vendor portal, but I'm still not done editing yet and not ready to promote.

For the purposes of this workflow, I think we are telling people that they should be _iterating_ by creating, (updating), and promoting several releases, rather than updating a single release and promoting one time when they are done with everything

docs/vendor/releases-creating-releases.md
```diff
@@ -1,54 +1,64 @@
...
+# Creating and Promoting Releases
 
-You can use the Replicated vendor portal to create and release versions of your application to various release channels. The vendor portal hosts a built-in YAML editor and linter to help you write and validate manifest files.
+You can use the Replicated vendor portal to create and promote versions of your application to various release channels.
 
-Alternatively, you can use the replicated CLI and API to automate releases. For more information about using the CLI, see [Installing the replicated CLI](../reference/replicated-cli-installing).
+To create and promote a release in the vendor portal:
 
-Replicated lets you add custom resources to your releases, which are packaged as part of the application but not deployed to the cluster. When included, custom resources are consumed by the app manager, the admin console, or by other kubectl plugins to control the application experience. For more information about the custom resources, see [About Custom Resources](../reference/custom-resource-about).
+1. From the **Applications** dropdown list, select **Create an app** or select an existing release to update.
```
paigecalvert, line 7: I think there is terminology mismatch in this step. This tells us to either create a new _app_, or select an existing _release_.

Should this say "or select an existing app"?

docs/vendor/releases-creating-releases.md
```diff
@@ -1,54 +1,64 @@
...
-To create a release:
+  ![Create Release](/images/release-create-new.png)
 
-1. Log in (or create a new team) on the [vendor portal](https://vendor.replicated.com) and create a new application. After signing up and activating your account, you are prompted to create a new application. Name your application and click **Create Application**.
+  [View a larger image](/images/release-create-new.png)
 
- :::note
- If you are logging in with an existing account, you can update an existing application or select **Create new app** from the application drop-down list.
- :::
+1. In the YAML editor, drag and drop your application files into the file directory. These can be Kubernetes manifest files or Helm charts. Click the plus icon to add a new YAML file, if needed.
```
paigecalvert, line 15: Could consider linking to the "Adding Helm Charts to a Release" topic in this step for more info: https://docs.replicated.com/vendor/helm-release

docs/vendor/releases-creating-releases.md
```diff
@@ -1,54 +1,64 @@
...
+   ![Default YAML](/images/guides/kots/default-yaml.png)
 
-  The Channels page opens and displays a list of your release channels, which are logical stacks for you to stage and promote releases to your customers.
+   [View a larger image](/images/guides/kots/default-yaml.png)
 
-1. Click **Releases** on the left menu, and then click **Create a release**.
+1. Edit the YAML files as needed. For example, you can edit a custom resource or map Helm chart values.
 
-  ![Create Release](/images/guides/kots/create-release.png)
+1. Click **Save release**, and then click **Promote**.
```
paigecalvert, line 23: This could be a place to mention what saving a release does (saves as draft), and that if they need to come back and make any edits after saving but before promoting, they can click "Edit YAML" from the Releases page.

I was thinking just another sentence, but up to you though if you think this would make sense for this iteration

docs/vendor/releases-creating-releases.md
paigecalvert: Another thing to consider that I am just realizing...If someone already has several releases, and they click Create Release, then the vendor portal is going to just copy all the files from the previous release into the new release.

Possible solution for conveying this in this procedure could be mentioning that this step applies only if you are creating a release for the first time, or if you need to bring over new application manifest files for some other reason.

Otherwise, if I'm creating my second or third release for example to add a Config custom resource, I don't need to drag over any files

docs/vendor/releases-creating-releases.md
```diff
@@ -1,54 +1,64 @@
...
+1. In the YAML editor, drag and drop your application files into the file directory. These can be Kubernetes manifest files or Helm charts. Click the plus icon to add a new YAML file, if needed.
 
-  ![Create Application](/images/guides/kots/create-application.png)
+   ![Default YAML](/images/guides/kots/default-yaml.png)
 
-  The Channels page opens and displays a list of your release channels, which are logical stacks for you to stage and promote releases to your customers.
+   [View a larger image](/images/guides/kots/default-yaml.png)
 
-1. Click **Releases** on the left menu, and then click **Create a release**.
+1. Edit the YAML files as needed. For example, you can edit a custom resource or map Helm chart values.
```
paigecalvert, line 21: Since we are not providing detail or links in this step (intentionally), the examples don't feel like they are providing value. ("Map helm chart values" in particular might not have a lot of meaning without some additional context--mapping what values where?)

Could consider removing the examples. And if so, then maybe this step could be combined with the Save step below, since they are both pretty small.

For example:

"Edit the YAML files in the release as needed, and click Save release. This saves a draft of release that you can continue to edit..."

docs/vendor/releases-creating-releases.md
```diff
@@ -1,54 +1,64 @@
...
 
-  ![Create Release](/images/guides/kots/create-release.png)
+1. Click **Save release**, and then click **Promote**.
 
-  [View a larger image](/images/guides/kots/create-release.png)
+1. In the Promote Release dialog, edit the fields:
 
-  A YAML editor displays.
+    ![Promote release dialog](/images/release-promote.png)
```
paigecalvert, line 27: Since the table below lists out the fields and describes them, this screenshot might not be necessary (just thinking as they edit these fields it's easier to keep the table up to date than the screenshot).

If keeping it, recommend reducing the size to avoid having to scroll to see the table beneath it

paigecalvert (dismissed):
Thank you @jonquil2002 ! Huge improvement already. There were a few places where I recommended wording edits (mostly around the difference between updating a release, and creating several new releases in order to iterate)

docs/vendor/distributing-workflow.md
jonquil2002: @paigecalvert  Ah, now I see where our communication is fuzzy. I am defining these differently than you are, so here is my take:

_Updating a release_ means making any kind of change (large or small) and promoting it. I'm taking this meaning from this topic on the docs site: https://docs.replicated.com/vendor/releases-updating. Updating can simply be a small change for an enhancement of a production release or it can mean updating a single release that's part of a larger iteration.

_Editing_ is how I refer to making changes to a YAML file before it's promoted, which I don't think needs to be part of a heading or title. It's just a step in the larger task.

_Iterating_ means continuing to create, promote, and test several releases with the intent of getting to a production-ready version.

Lmk if we should pair on this.

docs/vendor/releases-creating-releases.md
jonquil2002: Good catch!

docs/vendor/releases-creating-releases.md
```diff
@@ -1,54 +1,64 @@
...
 
-  [View a larger image](/images/guides/kots/default-yaml.png)
+  The release appears in an **Active** state on the Releases page.
 
-1. To manage the file directory structure, note the following options:
+## Next Step
 
-    - To delete files, click the Trash icon that displays when you hover over a file.
-    - To create a new file or folder, click the corresponding icons at the bottom of the file directory pane.
+Update and test the release in your development environment using the Replicated admin console. For more information about updating an application release in the admin console, see [Updating an Application](../enterprise/updating-apps).
```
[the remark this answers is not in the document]
jonquil2002: Actually, I don't mean updating as being editing a release only without promoting it. But, this Next Steps section actually links to the Enterprise section for updating an application in the admin console to test it. We don't document the specific steps for testing anywhere except in the Getting Started tutorial if we remove it from here. We were documenting this testing in "Updating and Archiving a Release", but now it's changed to "Archiving" only. I really think this next step should stay. What do you think in light of this explanation?

paigecalvert (approved):
Looks good to me! Let's merge 

Just to close the thread about "updating" (and for future iterations): I also did a poll in the product channel about how people use the word "updating", and got some interesting responses. Dex called out both possible definitions we were seeing, and Alex mentioned that he recommends avoiding the term "updating" all together, since it can be confusing.
https://replicated.slack.com/archives/C68QDSQ64/p1671041686004989

PR #829, opened 2022-12-13
Moves helm CLI private registry info to registries section

paigecalvert:
Story: https://app.shortcut.com/replicated/story/63750/break-up-the-helm-cli-install-topic-move-private-registries-info

Preview of new topic: https://deploy-preview-829--replicated-docs.netlify.app/vendor/helm-image-registry


paigecalvert:
@ajp-io This is ready for your review. But I know you have a few docs things plus Replicon that you are juggling this week, so this can be lower priority.

This is the first PR (of likely several) to begin breaking apart the Supporting helm CLI Installations (Beta) topic.

As part of this effort, I'll also be doing some PRs to break apart the "Packaging with Helm" section in the docs as well.

I'll be in touch with you throughout (I expect this to be my main focus for at least the rest of this month/January)

sidebars.js
```diff
@@ -66,6 +66,17 @@ const sidebars = {
         'vendor/distributing-workflow',
         'vendor/vendor-portal-creating-account',
         'vendor/planning-questionnaire',
+        {
+            type: 'category',
+            label: 'Using Private Registries',
```
jonquil2002, line 71: How do you feel about just naming this section "Private Registries" to match the direction we're going with for "Release" - so, removing the gerund.

I like where you've placed it in the TOC.

docs/vendor/packaging-private-images.md
```diff
@@ -6,10 +6,13 @@ to your customer.
 
 The customer license file can grant revokable image pull access to private images,
 whether the images are stored in the Replicated private registry or a supported
-external registry:
+external registry.
```
jonquil2002, line 9: A general side note for this topic, which can be considered for a follow-up story: We don't tell them that public repositories are supported (though likely not recommended).

I like that you added the other methods here though!

jonquil2002 (approved):
LGTM!

sidebars.js
paigecalvert: That works for me!

paigecalvert:
Talked to Alex on slack. Since this one doesn't include material content changes and just reorgs, going to merge this without SME review.

PR #830, opened 2022-12-14
Update preflight-support-bundle-creating.md

grahamnscp:
Hi 👋 
I noticed on the preflights that the spec entry on a code sample was indented, I carried on in the doc and suggested some more indent changes to all match 2 space steps (vs 3 or more in some examples) Best regards,
Graham


paigecalvert (approved):
Thank you @grahamnscp !

Issue #835, opened 2022-12-15
Docs feedback on custom-resource-lintconfig


dexhorthy:
There's a broken link on this page -- `Linter Rules` links to `#linter-rules` but it looks like that was moved to another page maybe? https://docs.replicated.com/reference/custom-resource-lintconfig#linter-rules

paigecalvert:
Thanks @dexhorthy ! PR to fix above: https://github.com/replicatedhq/replicated-docs/pull/837

paigecalvert:
merged!

dexhorthy:
woohoo!

PR #837, opened 2022-12-15
Fix broken link

paigecalvert:
From https://github.com/replicatedhq/replicated-docs/issues/835


jonquil2002 (approved):
lgtm...but should this be `../reference/linter`?

The link works in this preview though.

paigecalvert:
We should be using links without the `../`. We had to use the `../` previously because docusaurus wasn't properly routing without it, for some reason.

The benefit of _not_ using `../` is that the link will work regardless of where the topic moves. For example, say we had a `(../reference/my-reference.md)` link in a topic here located here: `vendor/some-topic.md`. And then we moved `vendor/some-topic.md` to `vendor/new-folder/some-topic.md`, the `../reference/my-reference.md` link would break. Without the `../` however, docusaurus will always know to just go link to the topic in the `/reference` folder

PR #850, opened 2022-12-21
doc: Clarifies the recommendations to not lead users to make bad design choices by looking the docs examples

camilamacedo86:
## Description 

The example to use `kots.io/exclude` might lead users to think that is recommended they try to install what is installed by KURL when it should not be done and would be considered a bad practice. 

So, the goal of this PR is added the info to let the users be able to carefully think about how they will use this feature.

### DIFF
--- a/docs/vendor/packaging-ingress.md
+++ b/docs/vendor/packaging-ingress.md
@@ -116,3 +116,5 @@ You should think carefully about the options added to be installed via your solu
 
 Assuming that your solution requires Prometheus and you are looking for a design that would be possible to consume it in a pre-existing cluster. Ideally, a cluster should have only one Prometheus installed and configured, which will serve all solutions running on it. Therefore, unless your solution's primary purpose would be installing and managing Prometheus, it does not fit in the application domain. Otherwise, it might become problematic when more than one solution deployed on the cluster tries to install Prometheus in this example. 
 
+For needs like the above example, a recommended approach would be to let the application users know that they can quickly get it running with the embedded cluster(`kURL`) or ensure that your solution's pre-requirements are correctly available on the pre-existent cluster. A typical option is to address this requirement via guidance and docs.
+ 
--- a/docs/vendor/packaging-ingress.md
+++ b/docs/vendor/packaging-ingress.md
@@ -107,3 +107,12 @@ spec:
               serviceName: nginx
               servicePort: 80
 ```
+
+## It is NOT recommended to use this option to address needs outside of the solution domain
+
+You should think carefully about the options added to be installed via your solution. Suppose you are looking to develop a design where your application would be capable of installing what is installed by the embedded cluster ([kURL](https://kurl.sh/)). Than note that it is **not** recommended design. Keep in mind that the kURL will install solutions that are valid for the whole cluster and which does not fit well in the application domain. 
+
+**To better clarifies the above recommendation, let's think in the example following example scenario.**
+
+Assuming that your solution requires Prometheus and you are looking for a design that would be possible to consume it in a pre-existing cluster. Ideally, a cluster should have only one Prometheus installed and configured, which will serve all solutions running on it. Therefore, unless your solution's primary purpose would be installing and managing Prometheus, it does not fit in the application domain. Otherwise, it might become problematic when more than one solution deployed on the cluster tries to install Prometheus in this example. 
+
--- a/docs/vendor/packaging-ingress.md
+++ b/docs/vendor/packaging-ingress.md
@@ -108,9 +108,9 @@ spec:
               servicePort: 80
 ```
 
-## It is NOT recommended to use this option to address needs outside of the solution domain
+## Common Pitfalls and Best Practices
 
-You should think carefully about the options added to be installed via your solution. Suppose you are looking to develop a design where your application would be capable of installing what is installed by the embedded cluster ([kURL](https://kurl.sh/)). Than note that it is **not** recommended design. Keep in mind that the kURL will install solutions that are valid for the whole cluster and which does not fit well in the application domain. 
+It is NOT recommended to use the option described above to address needs outside of the solution domain. You should think carefully about the options added to be installed via your solution. Suppose you are looking to develop a design where your application would be capable of installing what is installed by the embedded cluster ([kURL](https://kurl.sh/)). Than note that it is **not** recommended design. Keep in mind that the kURL will install solutions that are valid for the whole cluster and which does not fit well in the application domain. 
 
 **To better clarifies the above recommendation, let's think in the example following example scenario.**
 


docs/vendor/packaging-ingress.md
```diff
@@ -107,3 +107,12 @@ spec:
...
+
+## It is NOT recommended to use this option to address needs outside of the solution domain
+
+You should think carefully about the options added to be installed via your solution. Suppose you are looking to develop a design where your application would be capable of installing what is installed by the embedded cluster ([kURL](https://kurl.sh/)). Than note that it is **not** recommended design. Keep in mind that the kURL will install solutions that are valid for the whole cluster and which does not fit well in the application domain. 
+
+**To better clarifies the above recommendation, let's think in the example following example scenario.**
+
+Assuming that your solution requires Prometheus and you are looking for a design that would be possible to consume it in a pre-existing cluster. Ideally, a cluster should have only one Prometheus installed and configured, which will serve all solutions running on it. Therefore, unless your solution's primary purpose would be installing and managing Prometheus, it does not fit in the application domain. Otherwise, it might become problematic when more than one solution deployed on the cluster tries to install Prometheus in this example. 
+
```
camilamacedo86, line 118:
```suggestion

For needs like the above example, a recommended approach would be to let the application users know that they can quickly get it running with the embedded cluster(`kURL`) or ensure that your solution's pre-requirements are correctly available on the pre-existent cluster. A typical option is to address this requirement via guidance and docs 
```

docs/vendor/packaging-ingress.md
```diff
@@ -107,3 +107,14 @@ spec:
               serviceName: nginx
               servicePort: 80
 ```
+
```
ricardomaraschini, line 110: Is the purpose of this guidance to discourage users from installing components that are already provided by the cluster, such as OpenEBS, through a Kots application? This documentation section appears to focus on the process of providing two distinct resource files and selecting the appropriate one based on the available environment. If this is the case, it may be more appropriate to include this PR in a "Best Practices" or "Common Pitfalls" section of the documentation. If I have misunderstood the content of the documentation, please let me know.

camilamacedo86: Hi @ricardomaraschini,

Thank you for your time and review here. Following the answers inline:

> Is the purpose of this guidance to discourage users from installing components that are already provided by the cluster, such as OpenEBS, through a Kots application? 

Yes. 

>  This documentation section appears to focus on the process of providing two distinct resource files and selecting the appropriate one based on the available environment.

Note that this doc provide an example over how to install ingress when the solution is not running and using the annotation `kots.io/exclude: '{{repl or (ConfigOptionEquals "enable_ingress" "1" | not) (not IsKurl) }}'` which lead to users think that is totally fine use this option to have the app installing what is installed by kURL.

Because of that I think a note here would be helpful. 

> If this is the case, it may be more appropriate to include this PR in a "Best Practices" or "Common Pitfalls" section of the documentation. If I have misunderstood the content of the documentation, please let me know.

are you suggesting changing the section title from `## It is NOT recommended to use this option to address needs outside of the solution domain` to `Common Pitfalls`? Would that address your feedback or do you think that we should have a new doc for that?

docs/vendor/packaging-ingress.md
```diff
@@ -107,3 +107,14 @@ spec:
...
+## It is NOT recommended to use this option to address needs outside of the solution domain
```
camilamacedo86, line 111:
```suggestion
## Common Pitfalls and Best Practices
```

docs/vendor/packaging-ingress.md
```diff
@@ -107,3 +107,14 @@ spec:
...
+
+You should think carefully about the options added to be installed via your solution. Suppose you are looking to develop a design where your application would be capable of installing what is installed by the embedded cluster ([kURL](https://kurl.sh/)). Than note that it is **not** recommended design. Keep in mind that the kURL will install solutions that are valid for the whole cluster and which does not fit well in the application domain.
```
camilamacedo86, line 113:
```suggestion
It is NOT recommended to use the option described above to address needs outside of the solution domain. You should think carefully about the options added to be installed via your solution. Suppose you are looking to develop a design where your application would be capable of installing what is installed by the embedded cluster ([kURL](https://kurl.sh/)). Than note that it is **not** recommended design. Keep in mind that the kURL will install solutions that are valid for the whole cluster and which does not fit well in the application domain. 
```

docs/vendor/packaging-ingress.md
camilamacedo86: @ricardomaraschini I tried to address your feedback with the changes made in : https://github.com/replicatedhq/replicated-docs/pull/850/commits/3d55f01c2ce51f48b75cbf3ee3e00f98db3d9bca

Let me know if it address it and if we can close this comment within

camilamacedo86: Closing this one as sorted out to make easier win others reviews. If you do not agree please feel free to re-open.

paigecalvert:
Hi @camilamacedo86, just checking in to let you know that we created a story in our backlog for reviewing this PR and plan to pick it up this week: https://app.shortcut.com/replicated/story/65199/edit-review-pr-850

Apologies for the delay--we'll be in touch with questions and updates as we start work on the review. Thanks for writing this up!

paigecalvert:
Hi again, @camilamacedo86 ! Thanks for your patience. I had a couple initial questions for you, when you get a moment:

Is this an accurate summary of your recommendation? (I want to make sure I understand correctly)
> Multiple applications can be installed on the same cluster. For this reason, Replicated recommends that you do not package your application in such a way that it can install additional Kubernetes resources on the cluster beyond the resources that are included in your application. Instead, Replicated recommends that you do one of the following:
> - Instruct your users to install any required services or resources on their existing cluster before they install your application.
> - Configure a Kubernetes installer to provision a cluster that meets all the requirements for your application. Instruct your users to use the Kubernetes installer to provision a cluster, then install the application on that cluster.

I was also curious for more background information about this PR. Was this related to an issue that you recently saw in a customer support case? I ask because it sounds like this recommendation might apply more broadly to the process of packaging an application with Replicated, rather than only to Ingress resources. So, just curious how you were led to this topic :slightly_smiling_face: 

If this recommendation does apply as a higher level best practice, we might want to consider publishing the information in another location as well. (Just for context if you are interested, there is an ongoing effort to document all of our best practices and figure out the best way to get them into the docs: https://docs.google.com/document/d/1Rw2VAKm8Ey2Odm7TsaJd2TloWpxE1mus49GEKQtwEr4/edit)

Thanks again! And please let me know if I can clarify any of these questions for you.

paigecalvert:
@camilamacedo86 just following up to make sure you saw the comment above :)

camilamacedo86:
Hi @paigecalvert, 

Thank you for looking into that. Following the comments inline.

> Is this an accurate summary of your recommendation? (I want to make sure I understand correctly)

It does not pass precisely the same idea. It is not recommended to make the application responsible for installing/managing a solution that is not part of its scope. That is about something other than the quantity or to install of many things within. See the example added in the proposal; Prometheus is a solution valid for any application deployed on the cluster. So, imagine if we have more than one application/solution deployed on the cluster trying to install and manage Prometheus. 

> I was also curious for more background information about this PR. Was this related to an issue that you recently saw in a customer support case?

Yes, the support ticket was about problems faced because their solution was installing and managing something also installed by kURL. Therefore, kURL installs/manages solutions that are valid for all that can be deployed on the cluster., 

In the specific support ticket, they were doing something very similar to the example in this doc to try only installing the "addon" when KURL was not used to do the host install. However, that does not mean that when their solution is deployed will also always work fine because the cluster already has another application and/or the same installed in another way. 
 
Please, let me know if that clarifies the questions and feel free to reach out in slack.

paigecalvert:
Thanks @camilamacedo86 , the Prometheus example makes sense to me. (If the application is not specifically designed to install and manage Prometheus on a cluster, then don't try to make the application install Prometheus on the cluster!)

I think I am still not certain of how the content proposed in this PR relates to the Ingress example provided here specifically. Are you proposing that vendors should not attempt to deploy an Ingress resource as shown in the example, unless the only/main purpose of their application is to install and manage Ingress resources on clusters? (I suspect that is _not_ what you are saying, since that would mean that we should just remove this example entirely.)

Or, is it maybe just that you were concerned that this example was going to give people ideas that they should follow a similar approach with all other types of resources? (Did the person in the support ticket perhaps reference this specific Ingress example when they were explaining how they got the idea?)

Thanks again :)

paigecalvert:
That's where I'm leaning as well, @chris-sanders. It is feeling like a higher-level recommendation ("best practice") that requires more conversation. We are also of course still discussing to what degree it makes sense to incorporate best practices into the documentation. (The Docs team is planning to allocate time in the next quarter around the question of "best practices".)

It might be best @camilamacedo86 to close out this PR for now so that this best practice can be fully discussed in the referenced Google doc, then methodically introduced into the docs along with the other best practices that we want to add. Would that be alright with you?

paigecalvert:
@camilamacedo86 I am going to close this PR for now so that the conversation about best practices can continue in the google doc. (Then it can be revisited when the Docs team begins the process of making sure best practices are captured in the documentation where appropriate.) Thank you again for raising this!

PR #852, opened 2022-12-21
fix label for upstream change troubleshoot/918

adamancini:
https://github.com/replicatedhq/troubleshoot/pull/918


paigecalvert (approved):
Thanks @adamancini

PR #859, opened 2022-12-29
update data transmission policy statements

St0rmz1:
Add bullets for the specific data which is transmitted.
Added language also used for telemetry and reporting.
Added link to the other security info on website.


jonquil2002 (approved):
lgtm @St0rmz1 . I committed a few style edits.

PR #865, opened 2023-01-03
An important note about persisting generated values

divolgin:
Calling out an important distinction

Screen Shot 2023-01-03 at 2 26 01 PM


paigecalvert (approved):
@divolgin Made some minor edits to clarify in a few places additional places that this is talking specifically about when the RandomString function is assigned to a `value` key.

Also edited the note to give a little more info and downgraded it to a "note" rather than an "important", since it's good info to have, but there aren't major consequences that would occur if they do the wrong thing (like a security vulnerability or something)

Feel free to merge if you are ok with these changes or lmk if I need to make any changes. Thanks again!

PR #866, opened 2023-01-03
Move Swarm, Ship, AKA to EOL in support table

AmberAlston:
Now that it's past 12/31/22, these products need to be updated from EOA to EOL


paigecalvert (approved):
Thank you for being on top of this @AmberAlston !

PR #876, opened 2023-01-11
More accurate description for the LocalImageName function

divolgin:


paigecalvert:
Thanks @divolgin for adding these extra details. I committed some edits for formatting and adding a link to more info about the proxy service. LMK if my changes have affected the technical accuracy.

Something else that I was wondering is if it would be valuable to describe/provide an example of "_If no private registry is set, but the image needs to be proxied, the remoteImageName is rewritten to be pulled through the Replicated registry proxy service_".

Questions that came to mind where: Does this mean that `LocalImageName` would rewrite the image name to use the proxy service domain? Also, is there an example use case for needing to proxy an image that's not in a private registry? 

Not sure if these would be obvious to the reader, but wanted to run it by you.

divolgin:
@paigecalvert there is an example and detailed description in [this doc](https://docs.replicated.com/vendor/operator-referencing-images).  Maybe this can link there.

paigecalvert:
@divolgin nice, I hadn't seen that topic. Adding a link to that is a good idea (I'll make a commit).

paigecalvert:
@divolgin added:
- A link to the Referencing Images topic with a description of the Operator use case
- Examples that I pulled from the Referencing Images topic
- Clarified that the "private registry" in this case is referring to the customer's environment (which I think is true based on what I read in References Images)

Should be good to go if you are OK with these changes

paigecalvert (approved):

PR #879, opened 2023-01-12
Filing some suggested additions to this page pls

grahamnscp:
Couple of things to add here please:

- add some labels to the add node command would show that use case.  Node labels can be used for apps as placement constraints, SSD, enterprise-storage, gpu etc  ref: https://kurl.sh/docs/install-with-kurl/advanced-options - labels

- Recent question came up on tainting master nodes for multi-node clusters, with one node starting point workloads can run on the master, when worker nodes are added the master can be tainted to have noexec..  this is performed using the tasks.sh script option dedicated-primary.  I checked the open source doc reference is still valid for in the existing production code base.
ref: https://kurl.sh/docs/install-with-kurl/dedicated-primary

### DIFF
--- a/docs/enterprise/cluster-management-add-nodes.md
+++ b/docs/enterprise/cluster-management-add-nodes.md
@@ -4,9 +4,11 @@ This topic describes how to add primary and secondary nodes to a Kubernetes inst
 
 ## Overview of Adding Nodes
 
-You can generate commands in the admin console to join additional primary and secondary nodes to a Kubernetes installer provisioned cluster. Primary nodes run services that control the cluster. Secondary nodes run services that control the pods that host the application containers. Adding nodes can help manage resources to ensure that your application runs smoothly.
+The node where you first run the Kubernetes installer installation script is the primary node, which runs Kubernetes control-plane components including etcd. After you install, you can generate commands in the admin console to join additional primary and secondary nodes to your cluster.
 
-For high availability clusters, Kubernetes recommends using at least 3 primary nodes, and that you use an odd number of nodes to help with leader selection if machine or zone failure occurs. For more information, see [Creating Highly Available Clusters with kubeadm](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/) in the Kubernetes documentation.
+Primary nodes run services that control the cluster. Secondary nodes run services that control the pods that host the application containers. Adding nodes can help manage resources to ensure that your application runs smoothly.
+
+Adding additional nodes is also important for high availability. To ensure that your cluster has high availability, Kubernetes recommends that you use at least three primary nodes. Kubernetes also recommends that you use an odd number of nodes to help with leader selection if machine or zone failure occurs. For more information, see [Creating Highly Available Clusters with kubeadm](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/) in the Kubernetes documentation.
 
 ## Join Primary and Secondary Nodes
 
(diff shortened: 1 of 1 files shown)


chris-sanders:
I like the idea but I'm concerned it's not fleshed out enough. Single node is not the only reason we don't taint. Consider an end customer with 3 available nodes. The best possible option for them is an HA non-tainted control plane. 

Outside of specific circumstances I would only taint the nodes if I had at least 6 nodes available, which few of our customers are using kURL for today. Until I have dedicated 3 node control plane that's the most important thing. After that you can taint but it makes little sense to do so with 4 or even 5 nodes as your workloads now are single point of failures that can't create quorum. This is how I get to 6 nodes, failover with some minimal risk of resource starvation is better than no failover with less risk of starvation.

paigecalvert:
@grahamnscp let us know if you wanted to make any changes based on Chris' comment above! (Holding off on reviewing until we hear back)

paigecalvert:
@grahamnscp @chris-sanders, I added a few edits:
- Added a bit more context to the new section to make sure that it is more clear that we are not necessarily _recommending_ that people follow the steps to taint primary nodes. Rather, I think the idea is that we want to make sure that they know this is an option and that they have the resources they need to make an informed decision. (Please let me know if I've misunderstood this.)
- Added a link to some related Kubernetes documentation, which seems to provide information about common use cases for tainting/tolerations (@chris-sanders , I considered trying to add more detail about the 6+ nodes use case that you mention, but I think I was lacking the background knowledge to be able to explain it cohesively. Feel free to add a sentence or two if you think it's worthwhile.)
- Added in a link to the Dedicated Primary topic from kurl.sh, rather than copying the content over. This way, the content is not duplicated in different locations, creating a potential maintenance burden. (Just a general best practice.)

Could you both take a look to confirm that you are OK with these changes? Thank you!

chris-sanders:
Yes I do think this is well written now. I would like to have @kyleprice review the change as well. To my knowledge we don't test any form of taints, I don't know how it interacts with add ons, and I don't know if we intend for people to do this and by adding it here we are implying it is normal and supported.

While this is all technically true, if it's not intended, tested, and included in product design it may not remain true and shouldn't be advised to people as something they should be doing. I'm most concerned about add ons like Rook getting scheduled, then tainted, and ending up in an unhealthy state.

grahamnscp:
Nicely done @paigecalvert , thank you for researching the subject.

Small suggestion might be to emphasise "_add secondary nodes_" to "_add sufficient secondary nodes for the application_"

I also share the general product concern around services using persistent volumes being evicted from a single primary to worker nodes in an ha cluster scenario.  I like the idea of minimising this by allowing laying out the ha cluster, tainting the primaries before installing the kotsadm services and application. However we recommend always adding kotsadm as a kurl addon so we should test the effect that the taint primaries has on these components, I've been testing this with various addon combinations and all working well so far.  Some services stay on the primary.  So for me I feel confident to put this in the docs for those knowledgeable kubernetes end customers that will want this, size worker nodes differently to primaries, faster storage, larger volumes, CPUs, networks etc. /G

paigecalvert:
Thank you @chris-sanders and @grahamnscp !

Graham, I made the wording edit that you suggested--good idea.

@kyleprice , I've also added you as a reviewer. We're wondering about if you have any concerns with including this information about tainting primary nodes in the docs. (See Chris' comment above for details https://github.com/replicatedhq/replicated-docs/pull/879#issuecomment-1402526044)

chris-sanders:
> I've been testing this with various addon combinations and all working well so far. Some services stay on the primary. So for me I feel confident to put this in the docs for those knowledgeable kubernetes

Can you confirm:
* Upgrades work fine
* Rook, Longhorn, and OpenEBS are ok with this

Can you provide the tests you're running so the team can consider adding this? Or alternately, you're welcome to propose a PR to the testgrid that will cover it. I just assume that's a big ask, maybe you're already familiar with it.

I'm really happy you've done some testing here. I don't think we should add this to the documentation though until such time as the product test cover it. One off test can't be considered sufficient to expand product use case.

grahamnscp:
@camilamacedo86 please could you help me with @chris-sanders 's question around is there any "testgrid" test for taint primaries in an ha cluster deployment context.  Are there any, where can these be viewed and would I have access as an SE?..
Todate, I've only been running through install test with various addon combinations and config, I haven't tested upgrade at all (probably will when this comes up in a PoV though), thank you /G

paigecalvert:
Hi all, just checking in on this-- @grahamnscp , were you able to get ahold of anyone from kURL for your question?

paigecalvert:
This PR is pretty stale, so I'm going to close it out for now. (Happy to help with improving this info in the docs though depending on the answers to those open questions. Just keep us posted.)

PR #880, opened 2023-01-12
Fixing a couple of syntax items on hostpreflight examples

grahamnscp:
There were a couple of syntax things in the host preflight example here that broken the kurl cluster deployment when I tested it today
Ref troubleshoot doc example: https://troubleshoot.sh/docs/host-collect-analyze/http/


paigecalvert (approved):
thanks @grahamnscp !

PR #886, opened 2023-01-13
IMPORTANT: Add note to recommend those that has ubuntu distro do not use this release

camilamacedo86:
IMPORTANT: Add note to recommend those that has ubuntu distro do not use this release

### DIFF
--- a/docs/release-notes/rn-kubernetes-installer.md
+++ b/docs/release-notes/rn-kubernetes-installer.md
@@ -9,7 +9,7 @@ toc_max_heading_level: 2
 Released on January 3, 2023
 
 :::important
-We recommend use `v2023.01.13-1` instead because a regression issue was identified on this release where if conflicts resolutions are faced to install the host packages then the package manager might remove host one ore more packages installed (valid only for ubuntu distributions). In the best scenarios the packaged removed might not affected the host or it can be sorted out by getting the host re-installed with `sudo apt-get install <package>`. In the worst scenarios it might can remove important host packages such as `openssh` which would no longer allow you be able to connect via ssh in the host.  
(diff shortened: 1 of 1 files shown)


docs/release-notes/rn-kubernetes-installer.md
```diff
@@ -8,6 +8,10 @@ toc_max_heading_level: 2
 
 Released on January 3, 2023
 
+:::important
+We recommend use `v2023.01.13-1` instead because a regression issue was identified on this release where if conflicts resolutions are faced to install the host packages then the package manager might remove host one ore more packages installed (valid only for ubuntu distributions). In the best scenarios the packaged removed might not affected the host or it can be sorted out by getting the host re-installed with `sudo apt-get install <package>`. In the worst scenarios it might can remove important host packages such as `openssh` which would no longer allow you be able to connect via ssh in the host.
```
camilamacedo86, line 12:
```suggestion
We recommend use `v2023.01.13-1` instead because a regression issue was identified on this release where if conflicts resolutions are faced to install the host packages then the package manager might remove host one ore more packages installed (valid only for ubuntu distributions). In the best scenarios the packaged removed might not affected the host or it can be sorted out by getting the package re-installed with `sudo apt-get install <package>`. In the worst scenarios it might can remove important host packages such as `openssh` which would no longer allow you be able to connect via ssh in the host.  
```

chris-sanders:
```suggestion
A regression that can unintentionally remove critical system packages from Ubuntu machines has been found in this versions. Version `v2023.01.13-1` includes a fix for this regression and should be used in all cases instead. 
```

chris-sanders: Suggested reworking

chris-sanders (changes requested):

docs/release-notes/rn-kubernetes-installer.md
laverya: I would say 'in this version.' instead of 'in this versions.', if that makes sense

camilamacedo86: @paigecalvert @laverya @chris-sanders,

I think the most important Is we have the info. Please, feel free to pick the best wording word and move forward with this one. I am OK with all suggestions.

emosbaugh (approved):

paigecalvert (approved):

camilamacedo86:
We got the OK in the chat. Thank you for all help.

PR #892, opened 2023-01-17
66287: Move SemVer Info to About Releases

jonquil2002:
SC: https://app.shortcut.com/replicated/story/66287/move-semver-content-improve-sequence-content

Previews: 

- https://deploy-preview-892--replicated-docs.netlify.app/vendor/releases-about#release-sequence-mechanics

- https://deploy-preview-892--replicated-docs.netlify.app/vendor/releases-about#semantic-versioning
- https://deploy-preview-892--replicated-docs.netlify.app/vendor/releases-creating-channels#create-a-channel
- This change is in a partial that shows up in multiple topics, such as this one: https://deploy-preview-892--replicated-docs.netlify.app/vendor/releases-creating-releases#using-the-vendor-portal


docs/vendor/releases-about.md
```diff
@@ -20,17 +20,45 @@ After you save a release, you can promote it to any of your release channels. Wh
...
+When you enable semantic versioning on a channel, the version label for a release promoted to that channel is verified to ensure that it is a valid semantic version. For more information about valid semantic versions, see [Semantic Versioning 2.0.0](https://semver.org).
+
+If you enable semantic versioning for a channel and then promote releases to it, Replicated recommends that you do not later disable semantic versioning for that channel.
+
 ### Release Properties
 
 You cannot edit the YAML files in a release after the release is promoted to a channel. However, each release has properties, such as the release notes, the version label, and the required status, that you can edit from the channel Release History page in the vendor portal. For more information, see [About the Channels Page](/vendor/releases-about-channels#about-the-channels-page) in _About Release Channels_.
 
-## Release Sequence Mechanics
+### Release Sequence Mechanics
```
paigecalvert, line 39: s/ could delete the word "Mechanics" while keeping the meaning of the heading

docs/vendor/releases-about.md
```diff
@@ -20,17 +20,45 @@ After you save a release, you can promote it to any of your release channels. Wh
...
 
-**Release Sequence**: Each release has a unique, monotonically-increasing sequence number. This number can be used as a fallback to identify a release if the `Version Label` field is not set during promotion, or to identify an unpromoted draft release. For more information, see [Creating and Promoting Releases](releases-creating-releases).
+#### Release Sequence
+
+Each release has a unique, monotonically-increasing sequence number. This number can be used as a fallback to identify a release if the `Version label` field is not set during promotion, or to identify an unpromoted draft release. For more information, see [Creating Releases with Standard Manifest Files](releases-creating-releases#using-the-vendor-portal) or [Creating Releases with Helm Charts](helm-release#ui).
+
+#### Instance Sequence 
+
+When a release is _seen_ by an instance, a release identifier is returned to an app manager instance when checking for an application update. A seen release is assigned a separate instance sequence, starting at 0 and incrementing for each release that is seen.
```
paigecalvert, line 49: The introduction of the term _seen_ might be adding more confusion than it is helpful.

This first sentence, for example, feels like it assumes a lot of knowledge about how instances check for updates, and how the vendor portal and the app manager communicate with one another about new releases becoming available. _Seen_ might add more ambiguity here

docs/vendor/releases-about.md
```diff
@@ -20,17 +20,45 @@ After you save a release, you can promote it to any of your release channels. Wh
...
+If you enable semantic versioning for a channel and then promote releases to it, Replicated recommends that you do not later disable semantic versioning for that channel.
+
 ### Release Properties
 
 You cannot edit the YAML files in a release after the release is promoted to a channel. However, each release has properties, such as the release notes, the version label, and the required status, that you can edit from the channel Release History page in the vendor portal. For more information, see [About the Channels Page](/vendor/releases-about-channels#about-the-channels-page) in _About Release Channels_.
 
-## Release Sequence Mechanics
+### Release Sequence Mechanics
 
 By default, Replicated uses release sequence numbers to organize and order releases, and uses sequence numbers in an instance's internal version history.
```
paigecalvert, line 41: /s One point that could be useful to clarify is the difference of how the _vendor portal_ increments the release sequence, while the _app manager/admin console_ increments the sequence for each installed instance of an app. Sort of highlighting the fact that there are two different sequences going on in two different places, for two different audiences (vendor or enterprise user)

docs/vendor/releases-about.md
```diff
@@ -20,17 +20,45 @@ After you save a release, you can promote it to any of your release channels. Wh
...
 
-**Release Sequence**: Each release has a unique, monotonically-increasing sequence number. This number can be used as a fallback to identify a release if the `Version Label` field is not set during promotion, or to identify an unpromoted draft release. For more information, see [Creating and Promoting Releases](releases-creating-releases).
+#### Release Sequence
+
+Each release has a unique, monotonically-increasing sequence number. This number can be used as a fallback to identify a release if the `Version label` field is not set during promotion, or to identify an unpromoted draft release. For more information, see [Creating Releases with Standard Manifest Files](releases-creating-releases#using-the-vendor-portal) or [Creating Releases with Helm Charts](helm-release#ui).
```
paigecalvert, line 45: The passive voice in the second sentence might be adding some ambiguity about who is using this and why. I think it is saying that the vendor can use the release sequence number to identify specific releases if they themselves did not set a version label?

docs/vendor/releases-about.md
paigecalvert: /s to that end, maybe screenshots showing where the sequence numbers show up in the two different UIs could be valuable? your call on that one though

docs/vendor/releases-about.md
```diff
@@ -20,17 +20,45 @@ After you save a release, you can promote it to any of your release channels. Wh
...
+
+When a release is _seen_ by an instance, a release identifier is returned to an app manager instance when checking for an application update. A seen release is assigned a separate instance sequence, starting at 0 and incrementing for each release that is seen. 
+
+A single release sequence, such as `181`, can have multiple instance sequences in the deployed instances, depending on when those instances came online and how many other releases were seen before release sequence `181`. 
+
+Note that instance sequences are only tracked by app manager instances, and the Replicated SaaS platform has no knowledge of these numbers.
+
+#### Semantic Versioning Sequence
+
+For channels with semantic versioning enabled, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
```
paigecalvert, line 57: Do you know then if this is talking about the _instance sequence_ or the _release sequence_ number?  (Sounds like the instance sequence since it's referring to what the admin console does?)

docs/vendor/releases-creating-channels.md
```diff
@@ -21,7 +21,8 @@ To create a channel:
   ![Create Channel](/images/channels-create.png)
 
 1. Enter a name and description for the channel.
-1. (Optional) To enable semantic versioning on the channel, turn on **Enable semantic versioning**. For more information, see [Enabling Semantic Versioning](releases-semantic-versioning).
+1. (Optional) To enable semantic versioning on the channel, turn on **Enable semantic versioning**. If you enable semantic versioning for a channel and then promote releases to it, Replicated recommends that you do not later disable semantic versioning for that channel. For more information, see [Semantic Versioning](releases-about#semantic-versioning).
```
paigecalvert, line 24: I know that this is unrelated to the story so feel free to ignore, but maybe we want to indicate that enabling semantic versioning is enabled? (That is the direction that we want people to go, right?)

docs/vendor/releases-creating-channels.md
```diff
@@ -30,14 +31,24 @@ To create a channel:
...
-
-  The Edit a channel dialog opens.
+1. Click the gear icon on the top right of the channel that you want to modify.
 
-  ![Edit Channel](/images/channels-edit.png)
+  The Channel settings dialog opens.
 
 1. Edit the fields and click **Save**.
 
+## Enable Semantic Versioning using the CLI
```
paigecalvert, line 40: Nice addition :)

docs/vendor/releases-about.md
jonquil2002: Good ideas.

docs/vendor/releases-creating-channels.md
jonquil2002: Semantic versioning is not enabled on some channels and also depends on what version of app manager is being used. I just added some clarification for that.

docs/vendor/releases-about.md
jonquil2002: @dexhorthy  Can you answer Paige's question? ^

Please see the newly drafted content where you can see the juxtaposition of SemVer with the other sequence types.

docs/vendor/releases-about.md
```diff
@@ -18,19 +18,57 @@ Replicated includes several custom resources that you can add to your releases,
...
+
+Semantic versioning is available with the Replicated app manager v1.58.0 and later. Note the following:
+
+- For applications created in the vendor portal on or after February 23, 2022, semantic versioning is enabled by default on the Stable and Beta channels. Semantic versioning is disabled on the Unstable channel by default.
+
+- For existing applications created before February 23, 2022, semantic versioning is disabled by default on all channels.
+
+When you enable semantic versioning on a channel, the version label for a release promoted to that channel is verified to ensure that it is a valid semantic version. For more information about valid semantic versions, see [Semantic Versioning 2.0.0](https://semver.org).
+
+If you enable semantic versioning for a channel and then promote releases to it, Replicated recommends that you do not later disable semantic versioning for that channel.
```
dexhorthy, line 33: Not just a docs thing, but if we recommend against, we should say why. If the why is really bad (e.g. you'll break everything) then we should adjust the product not to allow it

docs/vendor/releases-about.md
dexhorthy: I'm not sure the answer here. I'd tap @ajp-io or team orange on this one for making sure this is accurate

docs/vendor/releases-about.md
```diff
@@ -18,19 +18,57 @@ Replicated includes several custom resources that you can add to your releases,
...
+For channels with semantic versioning enabled, the Replicated admin console sequences releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
 
-**Release Sequence**: Each release has a unique, monotonically-increasing sequence number. This number can be used as a fallback to identify a release if the `Version Label` field is not set during promotion, or to identify an unpromoted draft release. For more information, see [Creating and Promoting Releases](releases-creating-releases).
+If releases that do not use a valid semantic version are already promoted to a channel, the admin console sorts the releases that do have semantic versions starting with the earliest version and proceeding to the latest. For example, assume that you promote these releases in the following order to a channel: 1.0.0, abc, 0.1.0, xyz, and 2.0.0. Then, you enable semantic versioning on that channel. The admin console sequences the version history for the channel as follows: 0.1.0, 1.0.0, abc, xyz, 2.0.0.
```
dexhorthy, line 69: I think this example would be a lot more useful as a table

dexhorthy (dismissed):

docs/vendor/releases-about.md
jonquil2002: Good idea. I tried with a table but it was hard to make shoter headings that captured the example explanations. So I made bulleted lists instead. It's still easier to parse now.

docs/vendor/releases-about.md
jonquil2002: @ajp-io  Do you know what the repercussions would be if vendors disable SemVer after having it enabled? ^

docs/vendor/releases-about.md
jonquil2002: @ajp-io  Do you know if this section is talking about the instance sequence or the release sequence number?

docs/vendor/releases-about.md
ajp-io: I'm trying to remember the particulars of what happens… I think the basic reason we don't want people disabling it is that the sorting of versions in the version history page gets wonky in some way. @sgalsaleh do you remember the exact situation here?

docs/vendor/releases-about.md
ajp-io: I've never used the term "instance sequence", but that is what it's referring to. The versions pulled into the admin console are either sorted by their sequence in the admin console or by their semver.

jonquil2002: Thanks, Alex. Just btw - "Instance sequence" is a term that Dex came up with. You might want to discuss it as a PM team, but I can roll with this for now since it's accurate. Thanks!

docs/vendor/releases-about.md
```diff
@@ -18,19 +18,71 @@ Replicated includes several custom resources that you can add to your releases,
...
+
+The following graphic shows release sequence numbers in the vendor portal:
+
+![Release sequence numbers](/images/release-sequences.png)
+
+[View a larger version of this image](../../static/images/release-sequences.png)
+
+#### Instance Sequences 
+
+When an app manager instance checks for an application update, a release identifier is returned to the instance and is also assigned an instance sequence number. This instance sequence is separate from a release sequence in the vendor portal. Instance sequences start at 0 and increment for each release identifier that is returned when the app manager checks for an update.
```
paigecalvert, line 55: The wording here tripped me up a bit:
>...a release identifier is returned to the instance and is also assigned an instance sequence number.

I'm wondering if it could be simplified by clarifying which components are taking which actions. For example:

>When an instance checks for updates, the vendor portal returns a release identifier to the instance. The admin console then assigns the release a unique instance sequence number. The instance sequence in the admin console starts at 0 and increments...

What I'm not sure of exactly is which parts the vendor portal handles, and which the admin console handles. This is a best guess

docs/vendor/releases-about.md
```diff
@@ -18,19 +18,71 @@ Replicated includes several custom resources that you can add to your releases,
...
+
+A single release sequence, such as `181`, can have multiple instance sequences in the deployed instances, depending on when those instances came online and how many other releases were seen before release sequence `181`.
+
+Note that instance sequences are only tracked by app manager instances, and the Replicated SaaS platform has no knowledge of these numbers.
```
paigecalvert, line 59: I don't think we have the term "Replicated SaaS platform" elsewhere in the docs, so it might not be intuitive to people what we mean.

I recommend replacing with vendor portal for now, though maybe we need to discuss with Dex if it's necessary to introduce a new term to capture the trifecta of vendor portal + vendor API + replicated CLI (I imagine that's what is meant by Replicated SaaS Platform)

docs/vendor/releases-about.md
```diff
@@ -18,19 +18,71 @@ Replicated includes several custom resources that you can add to your releases,
...
+
+The following graphic shows instance sequence numbers on the Replicated admin console dashboard:
+
+![Instance sequence numbers](/images/instance-sequences.png)
+
+#### Semantic Versioning Sequence
+
+For channels with semantic versioning enabled, the Replicated admin console sequences instance releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
```
paigecalvert, line 67: >...instead of their creation dates.

I don't think there was previous mention of the admin console sequencing by creating dates. Should this creating date piece be included in the section about instance sequences?

paigecalvert: Noting also this is the second instance of "Replicated admin console" (can delete the word Replicated)

docs/vendor/releases-about.md
paigecalvert: Also a question on terminology: the phrasing "an app manager instance" jumped out to me in this paragraph because, in the topic about the Instance Details, there was lots of mentions of instances checking for updates. For example:
>The instance checks for updates. An update check can occur when a user clicks the Check for updates button in the Replicated admin console, or when the admin console runs an automatic update check...

See https://docs.replicated.com/vendor/instance-insights-details

It made me wonder if there might be a difference between an "application instance" and an "app manager instance" (my instinct is that an application instance is an installed instance of a given release of the vendor's app, whereas an app manager instance might be an installed instance of the admin console).

For the instance details page doc, it ended up being mostly about "application instances". 

Anyway, not sure I have a suggestion here, but I wanted to share this to highlight the possible diversion in terminology to see if you might have thoughts

paigecalvert (approved):
Just added some suggestions and questions @jonquil2002 --looks good!

docs/vendor/releases-about.md
jonquil2002: @paigecalvert Apparently app manager instance can be different than application instance. Per Alex, "Because we support multi app in the admin console, you could have one admin console instance with multiple application instances installed through it."

I haven't gotten an answer as to whether the PMs prefer "app manager instance" or "admin console instance" for the terminology. I'll raise that question in the Product channel.

jonquil2002: Looks like we're landing with "app manager instance" bc updates can happen fro the CLI too, and app manager is a function of KOTS rather than the UI.

docs/vendor/releases-about.md
```diff
@@ -18,19 +18,71 @@ Replicated includes several custom resources that you can add to your releases,
...
+
+Note that instance sequences are only tracked by app manager instances, and the vendor portal has no knowledge of these numbers.
+
+The following graphic shows instance sequence numbers on the Replicated admin console dashboard:
+
+![Instance sequence numbers](/images/instance-sequences.png)
+
+#### Semantic Versioning Sequence
+
+For channels with semantic versioning enabled, the admin console sequences instance releases by their semantic versions instead of their creation dates. The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version.
```
jonquil2002, line 67: @dexhorthy We don't mention creation dates under the Instance Sequences section, and I'm not sure how creation dates work. PTAL at the Instances Sequence section and see if adding something about creation dates there would help make this section clearer.

docs/vendor/releases-about.md
```diff
@@ -18,19 +18,71 @@ Replicated includes several custom resources that you can add to your releases,
...
+
+The following graphic shows release sequence numbers in the vendor portal:
+
+![Release sequence numbers](/images/release-sequences.png)
+
+[View a larger version of this image](../../static/images/release-sequences.png)
+
+#### Instance Sequences 
+
+When an app manager instance checks for an application update, the vendor portal returns a release identifier (a sequence number or a version label) to the instance. The admin console then assigns the release a unique instance sequence number. This instance sequence is separate from a release sequence in the vendor portal. The instance sequence in the admin console starts at 0 and increments for each release identifier that is returned when the app manager checks for an update.
```
jonquil2002, line 55: @dexhorthy  I did another rewrite in this section based on feedback from Paige. PTAL and make sure that it's still technically correct. Thanks.

dexhorthy: Sounds right. @sgalsaleh can you confirm that these are 0-indexed and not 1-indexed?

docs/vendor/releases-about.md
dexhorthy: Creation date should probably be changed to "promotion date" - the date the release was promoted to the channel associated with the active license.

What is meant by "does not sort"? Does that mean "sorted to the bottom"? "sorted to the top"? "randomly dropped into the ordered semver set"? (probably not the last one 😂 )

jonquil2002: @dexhorthy I'm not sure what is meant by "Does not sort" - that was written a long time ago! But I think it means that non-semantic versions stay in the order of their promotion date while semantic versions get ordered from earliest to latest version. PTAL the rewrite that I did here to see if it makes sense and is accurate:

I deleted "The admin console does not sort any releases already promoted to the channel that do not use a valid semantic version."

And added this to the next paragraph: "The releases with non-semantic versioning stay in the order of their promotion dates."

docs/vendor/releases-about.md
jonquil2002: Talked with Salah. He confirmed that this is 0-indexed.

PR #903, opened 2023-01-23
31972: Edit kotsadmin-operator info

jonquil2002:
SC: https://app.shortcut.com/replicated/story/31972/docs-kots-rbac-docs-are-wrong

The original SC story is outdated and the current conten is mainly correct, just the note really needed to be changed because kotsadm-operators are no longer used.

Preview: https://deploy-preview-903--replicated-docs.netlify.app/vendor/packaging-rbac#operators-and-multiple-namespaces

### DIFF
--- a/docs/vendor/packaging-rbac.md
+++ b/docs/vendor/packaging-rbac.md
@@ -123,7 +123,7 @@ spec:
 
 #### Reference Objects
 
-The following role is created automatically for namespace-scoped applications:
+The following Role is created automatically for namespace-scoped applications:
 
 ```yaml
 apiVersion: "rbac.authorization.k8s.io/v1"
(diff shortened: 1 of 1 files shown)


docs/vendor/packaging-rbac.md
```diff
@@ -121,19 +121,132 @@ spec:
   requireMinimalRBACPrivileges: true
 ```
 
-#### Reference Objects
+Next, add the reference objects. See [Reference Objects](#objects).
 
-The following Role is created for namespace-scoped applications:
+#### Reference Objects {#objects}
+
+For existing clusters with RBAC, you must add the `kotsadmin-operator-role` and `kotsadmin-operator-rolebinding`.
```
jonquil2002, line 128: @divolgin Is it correct to recommend this only for existing clusters?

divolgin: Yes, but also "with RBAC" is redundant since all clusters use RBAC.

We also no longer ship `operator`.  The rest looks outdated as well.  If you run `kots admin-console generate-manifests --minimal-rbac`, you can see current Role, RoleBinding, and ServiceAccount specs.

jonquil2002:
@divolgin  Please review once again. I made the corrections/changes based on our discussion. Thanks again for your help!

divolgin (approved):

paigecalvert (approved):
thanks! looks good

PR #907, opened 2023-01-25
HOLD FOR RELEASE - 67611: Custom Domains for Download Portal

jonquil2002:
SC: https://app.shortcut.com/replicated/story/67611/feature-custom-domain-download-portal

Preview: https://deploy-preview-907--replicated-docs.netlify.app/vendor/custom-domains#about-custom-domains

### DIFF
--- a/docs/vendor/custom-domains.md
+++ b/docs/vendor/custom-domains.md
@@ -7,7 +7,10 @@ This topic describes using custom domains to alias Replicated endpoints, and how
 Using custom domains have the following limitations:
 
 - A single custom domain cannot be used for both the registry and proxy endpoints. A single domain can map to registry.replicated.com for any number of applications, but cannot map to both registry.replicated.com and  proxy.replicated.com, even if the applications are different.
+
 - Custom domains cannot be used to alias replicated.app (release manifests), api.replicated.com (platform market API), the download portal, or other services.
+
+- Multiple custom domains for the download portal can be active at once, but only one custom domain can be the default. This default custom domain is used in the vendor portal for links to the download portal for all applications in the team. All active custom domains work whether or not they are the default.
 
 ## Custom Registry Domains {#registry}
 
@@ -69,15 +72,9 @@ To assign a custom domain to an application:
 1. Create a new release and add the code snippet to the Application custom resource manifest file to create the new field for either `proxyRegistryDomain` or `replicatedRegistryDomain`. For more information, see [proxyRegistryDomain](../reference/custom-resource-application#proxyRegistryDomain) and [replicatedRegistryDomain](../reference/custom-resource-application#replicatedRegistryDomain) in the _Application_ section.
 
 
-## Custom Domains for the Download Portal {#download}
+## Configure Download Portal Domains {#download}
 
 You can configure the download portal to use a custom domain instead of the default get.replicated.com. The download portal is used to share license and release files with customers. For more information about the download portal, see [Share Files through the Download Portal](releases-sharing-license-install-script#download-portal) in _Share License Files and Releases_.
-
-### Limitations
-
-- Multiple custom domains can be active at once, but only one custom domain can be the default. This default custom domain is used in the vendor portal for links to the download portal for all apps in the team. All active custom domains work whether or not they are the default.
-
-### Configure Custom Domains
 
 To configure a custom domain for the download portal:
 
--- a/docs/vendor/custom-domains.md
+++ b/docs/vendor/custom-domains.md
@@ -1,10 +1,32 @@
 # Using Custom Domains (Beta)
 
-This topic describes using custom domains to alias Replicated endpoints, and how to configure them using the Replicated vendor portal or the vendor API.
+This topic describes using custom domains to alias Replicated endpoints, and how to configure them using the Replicated vendor portal. For information about using the API instead, see [Using the Vendor API v3](/reference/vendor-api-using) in _Reference_.
+
+## About Custom Domains
+
+You can use custom domains as aliases for registry, proxy, and the download portal by creating Canonical Name (CNAME) records for your domains.
+
+Replicated domains are external to your domain and require additional security reviews by your customer. Using custom domains as aliases can bring the domains inside an existing security review and reduce your exposure.
+
+Using the Replicated vendor portal or the vendor API, you configure custom domains at the Team level. Then, you specify those domains within the release in the Application custom resource manifest file. This method lets you:
+
+- Have different domains for your applications, if needed
+- Roll out domain name changes in phases to prevent the application from breaking in production environments
+
+Verification of the domain is required using TXT records that undergo separate verification checks for:
+
+- Domain ownership: This verification is done when you initially add a record.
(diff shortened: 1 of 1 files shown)


docs/vendor/custom-domains.md
```diff
@@ -66,4 +66,38 @@ To assign a custom domain to an application:
...
+
+
+## Custom Domains for the Download Portal {#download}
+
+You can configure the download portal to use a custom domain instead of the default get.replicated.com. The download portal is used to share license and release files with customers. For more information about the download portal, see [Share Files through the Download Portal](releases-sharing-license-install-script#download-portal) in _Share License Files and Releases_.
+
+### Limitations
+
+- You can change from one custom domain to another, but only one domain per team can be active at any time.
+- If you have multiple teams, each team must use the same active domain.
```
ajp-io, line 79: Might have been a small misunderstanding here. These are configured at the team level, and the vendor can set up any number, and they'll all work. But only one can be the default custom domain, and that default custom domain is the one we link out to from the vendor portal.
```suggestion
- Multiple custom domains can be active at once, but only one custom domain can be the default. This default custom domain is used in the vendor portal for links to the download portal for all apps in the team. All active custom domains work whether or not they are the default.
```

docs/vendor/custom-domains.md
```diff
@@ -66,4 +66,38 @@ To assign a custom domain to an application:
...
+
+1. For **Verify ownership**, copy the text string and use it to create a TXT record in your DNS account. Click **Validate & continue**.
+
+  Your changes can take up to 24 hours to propagate.
+
+1. For **TLS cert creation verification**, copy the text string and use it to create a TXT record in your DNS account. Click **Validate & continue**.
+
+    Your changes can take up to 24 hours to propagate.
+
+1. For **Use Domain**, click **Yes, set as default** to set the domain as the active domain, or click **Not now** to save the domain without making it the default.
```
ajp-io, line 103: I added a note about what the default domain means, in case someone missed it above in the limitations. I also removed the part about "save it without making it the default" because it's already saved—they're just dismissing the modal when they click that.
```suggestion
1. For **Use Domain**, click **Yes, set as default** to set the domain as the default, so that the vendor portal will use it for links to the download portal, or click **Not now**.
```

jonquil2002:
@ajp-io I made some organization changes, and some light wording changes. Please review again and see if these make sense. The biggest changes was moving the limitations for the download portal to the main Limitations section. On relfection, it seemed to make sense to keep these in one place, but please lmk if you have a better idea. Thanks.

docs/vendor/custom-domains.md
```diff
@@ -7,8 +7,11 @@ This topic describes using custom domains to alias Replicated endpoints, and how
 Using custom domains have the following limitations:
 
 - A single custom domain cannot be used for both the registry and proxy endpoints. A single domain can map to registry.replicated.com for any number of applications, but cannot map to both registry.replicated.com and  proxy.replicated.com, even if the applications are different.
```
ajp-io, line 9:
```suggestion
- A single custom domain cannot be used for multiple endpoints. For example, a single domain can map to registry.replicated.com for any number of applications, but cannot map to both registry.replicated.com and proxy.replicated.com, even if the applications are different.
```

docs/vendor/custom-domains.md
```diff
@@ -7,8 +7,11 @@ This topic describes using custom domains to alias Replicated endpoints, and how
...
+
 - Custom domains cannot be used to alias replicated.app (release manifests), api.replicated.com (platform market API), the download portal, or other services.
```
ajp-io, line 11:
```suggestion
- Custom domains cannot be used to alias replicated.app (release manifests), api.replicated.com (platform market API), or other services.
```

docs/vendor/custom-domains.md
```diff
@@ -7,8 +7,11 @@ This topic describes using custom domains to alias Replicated endpoints, and how
...
 
+- Multiple custom domains for the download portal can be active at once, but only one custom domain can be the default. This default custom domain is used in the vendor portal for links to the download portal for all applications in the team. All active custom domains work whether or not they are the default.
+
 ## Custom Registry Domains {#registry}
 
 You can use custom domains as aliases for the replicated.registry.com and proxy.replicated.com endpoints, by creating Canonical Name (CNAME) records for your domains.
```
ajp-io, line 17: These whole sections (from "You can use custom domains as aliases" to "The TXT records can be removed once verification is complete") is applicable to download portal to. I think this stuff should be generalized and put outside of the registry domains section. Then we just have sections beneath it for configuring registry domains and configuring download portal domains.

jonquil2002:
@ajp-io  I moved the content around and generalized it, as you suggested. SWYT

docs/vendor/custom-domains.md
```diff
@@ -1,39 +1,45 @@
...
 
-Whether you use the Replicated private registry or the proxy service for your own private registry, these Replicated domains are external to your domain and require additional security reviews by your customer. Using custom domains as aliases for Replicated domains can bring the Replicated registry and proxy service inside an existing security review and reduce your exposure.
-
-Using the Replicated vendor portal or the vendor API, you configure custom domains for the Replicated private registry and proxy service at the Team level. Then, you specify those domains within the release in the Application custom resource manifest file. This method lets you:
+Using the Replicated vendor portal or the vendor API, you configure custom domains at the Team level. Then, you specify those domains within the release in the Application custom resource manifest file. This method lets you:
 
 - Have different domains for your applications, if needed
 - Roll out domain name changes in phases to prevent the application from breaking in production environments
```
ajp-io, line 14: Darn, I'm so sorry I misled you, but this part about adding it to the release and the benefits is only applicable to the registry domains. Download portal domains aren't added to a release. But everything else in this section is valid for both.

docs/vendor/custom-domains.md
```diff
@@ -57,13 +63,40 @@ To configure a custom domain in the vendor portal:
 
 1. Create a new release and create the new field for either `proxyRegistryDomain` or `replicatedRegistryDomain` in the Application custom resource manifest file. You can use the code snippet template from the previous step or manually code the field. For more information, see [proxyRegistryDomain](../reference/custom-resource-application#proxyRegistryDomain) and [replicatedRegistryDomain](../reference/custom-resource-application#replicatedRegistryDomain) in the _Application_ section.
 
-### Assign a Custom Domain to an Application
+## Assign a Custom Domain to an Application
```
ajp-io, line 66: Do we need a comparable section for setting the download portal domain as default? We already document that you can do that when setting it up, but a section comparable to this would show that you can do it after the fact too.

jonquil2002: Good idea. I added one, let's see wyt

docs/vendor/custom-domains.md
jonquil2002: Thanks for catching that. I did a little reorg, let's see if you think it makes sense.

ajp-io (approved):
I'll wait to merge this. There's a bug preventing the feature from working quite yet.

paigecalvert (approved):
looks great!

PR #909, opened 2023-01-26
Update latest supported Velero version

ajp-io:


ajp-io:
Feel free to change the wording (like how I linked to the supported Velero version table, or how I wrote "1.5.1 through 1.9.x"). But this should get you started :)

ajp-io:
Actually, I think the mention of Linux AMD64 is too prescriptive. The CLI runs on the user's machine, which could be a different architecture. So if they're on a new Mac, they'll need the Darwin binary instead. So I think we should remove the mention of Linux AMD64 and tell them to download the appropriate asset instead.

jonquil2002 (approved):
lgtm

PR #917, opened 2023-01-30
Update KOTS install reference to include new env variable options

ousbots:
Updates the KOTS reference install section to include instructions on how and when to use the new env variables the install script reads `REPL_INSTALL_PATH` and `REPL_USE_SUDO`

### DIFF
--- a/docs/reference/kots-cli-getting-started.md
+++ b/docs/reference/kots-cli-getting-started.md
@@ -13,22 +13,22 @@ If you are using an [embedded Kubernetes installer-created cluster](../enterpris
 
 ## Install
 
-To install the latest version of the kots CLI, run:
+To install the latest version of the kots CLI to `/usr/local/bin`, run:
 
 ```bash
 curl https://kots.io/install | bash
+```
+
+To install to a directory other than `/usr/local/bin`, run:
+
+```bash
+curl https://kots.io/install | REPL_INSTALL_PATH=/path/to/cli bash
 ```
 
 To install a specific version of the kots CLI, run:
 
 ```bash
 curl https://kots.io/install/<version> | bash
-```
-
-To install to a directory different than the default, `/usr/local/bin`, use the `REPL_INSTALL_PATH` environment variable.
-
-```bash
-curl https://kots.io/install | REPL_INSTALL_PATH=/usr/sbin bash
 ```
 
 To verify your installation, run:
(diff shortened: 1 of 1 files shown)


ajp-io (approved):

ajp-io:
@replicatedhq/replicated-docs This is ready for your review. We will merge it when it's released, but you can approve when you're ready.

paigecalvert:
Thank you @ousbots @ajp-io ! I'll review this afternoon

paigecalvert (approved):
I made some edits to reorganize/rename headings: https://deploy-preview-917--replicated-docs.netlify.app/reference/kots-cli-getting-started#install-without-root-access

Goal was to make it easier to see the different options for installing without root access at a glance.

I also added some extra context around the new examples, so lmk if anything I wrote is inaccurate. Otherwise, looks good to go!

ajp-io:
@paigecalvert I had to fix a typo. Feel free to approve again.

paigecalvert (approved):

docs/reference/kots-cli-getting-started.md
```diff
@@ -13,12 +13,18 @@ If you are using an [embedded Kubernetes installer-created cluster](../enterpris
 
 ## Install
 
-To install the latest version of the kots CLI, run:
+To install the latest version of the kots CLI to `/usr/local/bin`, run:
```
jonquil2002, line 16: It might be good to add an intro sentence here to clarify the access that works with these commands. Maybe something like, "If you have root access, you can run one of the following commands to install..."

docs/reference/kots-cli-getting-started.md
```diff
@@ -33,9 +39,57 @@ kubectl kots --help
 
 ## Install without Root Access
 
-You can install the kots CLI on computers without root access or computers that cannot write to the `/usr/local/bin` directory. This procedure applies to online clusters and air gapped environments.
+You can install the kots CLI on computers without root access or computers that cannot write to the `/usr/local/bin` directory.
+
+To install the kots CLI without root access, you can do any of the following:
```
jonquil2002, line 44: This is such a helpful section, nice addition!

docs/reference/kots-cli-getting-started.md
```diff
@@ -33,9 +39,57 @@ kubectl kots --help
...
+
+You can set the `REPL_INSTALL_PATH` environment variable to install the kots CLI to a directory other than `/usr/local/bin` that does not require elevated permissions.
+
+**Example:**
+
+```bash
+curl -L https://kots.io/install | REPL_INSTALL_PATH=~/bin bash
+```
+
+In the example above, the installation script installs the kots CLI to `~/bin` in the local directory.
```
jonquil2002, line 60: This is probably a good style guide discussion item. I've always been required to write an intro sentence for the example, so that a user has context before they see the example. And then, if needed, follow the example with more detail. So, something like: "In the following example, the installation script installs the kots CLI to `~/bin` in the local directory:" (You could potentially get rid of the note by adding something about the `~` in the intro too, or just leave it after the example code.)

Same comment for the rest of the examples. It really just depends on what we want our style to be.

jonquil2002 (dismissed):
lgtm

These sections look great. Just a few thoughts to consider.

docs/reference/kots-cli-getting-started.md
ajp-io: I think the caveat is that you don't necessarily need root access. It depends on the computer and whether writing to /usr/local/bin requires root access. So this is the standard install, but if it turns out that that needs root access, you can either change the installation directory or use sudo.

docs/reference/kots-cli-getting-started.md
ajp-io: I'll leave that to you and Paige to decide, but that could help make better sense of examples.

paigecalvert: Yes great idea, will make that edit

paigecalvert (approved):

PR #922, opened 2023-01-31
67320: Updating Customer License Types in Vendor Portal

jonquil2002:
SC: https://app.shortcut.com/replicated/story/67320/add-licensing-use-case

Preview: https://deploy-preview-922--replicated-docs.netlify.app/vendor/licenses-about-types#updating-a-customer-license-type


docs/vendor/licenses-about-types.md
```diff
@@ -23,6 +23,14 @@ additional information can be provided.
 application. For more details about this type, see [Community Licenses](#community-licenses)
 below.
 
+## Updating a Customer License Type
+
+You can update a customer license type in the Replicated vendor portal, for example, from a trial license to a paid license. An update impacts all customers who are using the same license. Then, your customers use the Replicated admin console to synchronize their licenses. If they are changing from a trial license to a paid license, they are also prompted to redeploy the existing version of the application.
```
sgalsaleh, line 28: They won't be prompted to redeploy the existing version of the application. It's a bit more complicated, which makes me wonder if it's okay to remove that last sentence.
```suggestion
You can update a customer license type in the Replicated vendor portal, for example, from a trial license to a paid license. An update impacts all customers who are using the same license. Then, your customers use the Replicated admin console to synchronize their licenses. If they are changing from a trial license to a paid license, they are also prompted to redeploy the existing version of the application.
```

jonquil2002: @sgalsaleh  We can remove the last sentence since it's inaccurate. But it sounds like the more complex situation might need documentation, unless it's an edge case that might be better off as a community post? 

@dexhorthy Do you have thoughts about whether or where to document end what end customers need to do when going form a trial license to a paid license?

dexhorthy: 1. let's not suggest/encourage folks to have multiple customers using the same license.
2. Let's yes point out that a license could be used for multiple instances
3. Let's clarify the mechanism that causes app updates to be staged, but I'm also kind of on board w/ @sgalsaleh  's instinct that this is more complicated than we want to sign up for explaining here, and maybe we just drop the last thought from the paragraph.

context: (I am pretty sure) we don't have template functions for `trial` vs `paid`, so that change in itself wont cause any application changes. But license fields do matter. But a vendor should know that, that context may not really belong here. It **is** however, worth covering the fact that a License change **may very well** stage a new application release. 

Also worth considering - I think even if the paid-trial change is a no-op (no app changes), do we still stage a release and prompt the user to deploy it in Version History? @sgalsaleh do you know?

```suggestion
You can update a customer license type in the Replicated vendor portal, for example, from a trial license to a paid license. An update impacts all instances that are deployed using that license. Your customer can use the Replicated admin console to synchronize the licenses on their instances. If they are changing from a trial license to a paid license, it is unlikely that the license change would prompt any changes to the application deployment itself. However, if you also modify other [License Fields](/vendor/license-fields) as part of the trial to paid conversion, for example, this may trigger actual changes to the deployed application.
```

docs/vendor/licenses-about-types.md
```diff
@@ -23,6 +23,14 @@ additional information can be provided.
...
+
+With the exception of community licenses, it is not possible for an end customer to replace one license file with another license file without completely reinstalling the Replicated app manager and your application. Using the vendor portal to change the license type prevents the end customer from having to reinstall.
```
dexhorthy, line 30: What is the exception? This could be more clear that it's specifically a one-way conversion from community to something else

docs/vendor/licenses-about-types.md
jonquil2002: @dexhorthy  PTAL my changes here. Also, your link to license fields produces a 404 error. Should this link to built-in license fields or custom license fields?

docs/vendor/licenses-about-types.md
```diff
@@ -23,6 +23,22 @@ additional information can be provided.
...
+
+You can update a customer license type in the Replicated vendor portal, for example, from a trial license to a paid license. An update impacts all instances that are deployed using that license.
+
+![Customer Page in Vendor Portal](/images/customer-license-type.png)
+
+[View a larger image](/images/customer-license-type.png)
+
+It is unlikely that the license change will prompt any changes to the application deployment itself. However, if you also modify other license fields as part of the license conversion, this can trigger changes to the deployed application. For more information about license fields, see [License Fields](license-fields).
+
+Your customer uses the Replicated admin console to synchronize the licenses on their instances. Unless a customer is upgrading from a community license, it is not possible to replace one license file with another license file without completely reinstalling the application. Using the vendor portal to change the license type prevents end customers from having to reinstall.
```
dexhorthy, line 36: Adding some detail about when and how - might also be worth grabbing the snippet from here, or a reference to it https://docs.replicated.com/vendor/instance-insights-details#about-reporting

```suggestion
Your customer uses the Replicated admin console to synchronize the licenses on their instances. For online instances, License updates will be pulled from the Replicated SaaS during the following events:

- An automatic or manual update check is performed (also happens on [App Status Change](https://docs.replicated.com/vendor/instance-insights-details#about-reporting) events) 
- The user clicks the "sync license" in the admin console

For Airgap (need more here but jumping on a call)

Unless a customer is upgrading from a community license, it is not possible to replace one license file with another license file without completely reinstalling the application. Using the vendor portal to change the license type prevents end customers from having to reinstall. 
```

jonquil2002: @dexhorthy  I decided to merge this PR as is because it's been sitting for a long time and it's accurate. It's better if we can iterate to capture your additional changes. Here is the new SC card I started for the iteration: https://app.shortcut.com/replicated/story/69153/add-details-to-customer-license-types

PR #942, opened 2023-02-07
Added some clarification for RBAC policies

sj-porter-knime:
Added some clarification on how to prevent promotions to the wrong channel. It was unclear to me that Channel ID is not the same as Channel Name.

### DIFF
--- a/docs/vendor/team-management-rbac-about.md
+++ b/docs/vendor/team-management-rbac-about.md
@@ -46,18 +46,20 @@ The most specific rule definition is always applied, when compared with less spe
 
 ### Rule Order Examples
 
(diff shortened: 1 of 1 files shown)


paigecalvert:
Thanks very much @sj-porter-knime ! Great clarification. I made a couple minor edits.

I also noticed that we don't have good instructions in the docs on where the find the channel ID. (It's listed in the Settings modal for the given channel. See the screenshot here: https://docs.replicated.com/vendor/releases-creating-channels#edit-a-channel.) Going to update the description of the channel page here to clarify this: https://docs.replicated.com/vendor/releases-about-channels

paigecalvert (approved):

PR #950, opened 2023-02-13
Be consistent with v before app manager version num

ajp-io:


ajp-io:
Looks like we use the v before the version number in most places, so I just standardized that all around.

Might want to consider doing it on the release notes page too, since technically it looks like the GitHub release number has the v in it (see [GitHub releases](https://github.com/replicatedhq/kots/releases)). Maybe should ask an engineer about that.

paigecalvert (approved):
thanks @ajp-io !

paigecalvert:
@ajp-io I'll also ask about the v in release notes

PR #956, opened 2023-02-14
[HOLD FOR RELEASE] add limitations for the `admin-console generate-manifests` command

cbodonnell:
These changes support https://github.com/replicatedhq/kots/pull/3624, which added support for OpenShift and GKE Autopilot when running the `admin-console generate-manifests` command with a Kubernetes cluster context.  It also adds a limitation regarding upgrades using the generated manifests from older KOTS versions.


docs/reference/kots-cli-admin-console-generate-manifests.md
```diff
@@ -4,7 +4,9 @@ Running this command will create a directory on the workstation containing the R
 
 ### Limitations
 
-`generate-manifests` does not support generating manifests for Red Hat OpenShift clusters.
+`generate-manifests` does not support generating manifests for Red Hat OpenShift clusters or GKE Autopilot clusters if executed without a Kubernetes cluster context.
+
+The generated manifests cannot be used to upgrade an app manager instance that was ever on version 1.72.0 or earlier if run without a Kubernetes cluster context.
```
jonquil2002, line 9: @cbodonnell  I think this could be made a little clearer, but I'm not sure about my technical accuracy. This is what I suggest instead:

"To upgrade an app manager instance that has ever been on version 1.72.0 or earlier, you must run `generated manifests` with a Kubernetes cluster context."

cbodonnell:
Thanks for the review @jonquil2002.  FYI, I may have to circle back around to this one at a later time since there's now another item left to investigate regarding the implementation in https://github.com/replicatedhq/kots/pull/3624, which may impact the docs PR.

paigecalvert (approved):

cbodonnell:
Merging since we've released the change.  Tagging @ajp-io in case there are any edits we might want to make after the fact.

PR #959, opened 2023-02-14
Rewrite Creating Backups topic

paigecalvert:
https://app.shortcut.com/replicated/story/69033/enhancements-for-creating-and-scheduling-backups

Preview: https://deploy-preview-959--replicated-docs.netlify.app/enterprise/snapshots-creating

(Also updated a handful of topics to remove xrefs to the scheduling topic)


docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,91 @@
...
+* **Admin console restore**: Restores admin console data only.
 
-    ![Instance Backup UI](/images/snapshot-instance-backup.png)
+For more information about restoring from full backups, see [Restoring from Backups](snapshots-restoring-full).
 
-After a full snapshot has been created, you can get a command to do a restore by clicking the restore icon (check screenshot below).
-There are two available options for doing a restore. You can either do a full restore which will restore the admin console and the application with all of its data, or you can choose to do a partial restore of just your application and its data.
+You can create a full backup with the kots CLI or in the admin console UI:
+* [Create a Backup with the CLI](#cli-backup)
+* [Admin Console](#admin-console-backup)
```
jonquil2002, line 23: I think the bullets here should mirror each other, so maybe change "[Admin Console]" -> "[Create a Backup in the Admin Console]"

and you could change the intro sentence to something like "You can create a full backup using one of the following methods:"

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,91 @@
...
+### Create a Backup with the CLI {#cli-backup}
 
-If you have multiple applications within the admin console, each application should have a backup resource in order to be included in the full snapshot backup. For more information, see [Backup](../reference/custom-resource-backup) in the _Custom resources_ section.
+To create a full backup with the kots CLI, run the following command:
 
-## Partial Snapshots {#partial}
+   ```
+   kubectl kots backup --namespace NAMESPACE
+   ```
+   Replace `NAMESPACE` with the namespace where the admin console is installed. For more information, see [backup](/reference/kots-cli-backup-index) in _kots CLI_.
```
jonquil2002, line 32: Sometimes we've used "_Reference_" when we could be more specific, and here it's "_kots CLI_". Do we have a preference?

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,91 @@
...
-Partial snapshots only back up application volumes and application manifests; they do not back up the admin console or the metadata about an application.
-They are useful for capturing information before deploying a new release, in case you need to roll back, but they are not suitable for full disaster recovery.
-For backups that give you the ability to do full disaster recovery, use full snapshots. For more information about full snapshots, see [Full Snapshots (recommended)](#full).
+### Create a Backup in the Admin Console {#admin-console-backup}
 
-Partial snapshots can only be created via the admin console (check screenshot below).
+To create a full backup in the admin console:
 
-![Application Backup UI](/images/snapshot-application-backup.png)
+1. Log in to the admin console and go to **Snapshots > Full Snapshots (Instance)**.
```
jonquil2002, line 38: You could delete "Log in to the admin console" since you;re orienting them in the stem sentence.

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,91 @@
...
-They are useful for capturing information before deploying a new release, in case you need to roll back, but they are not suitable for full disaster recovery.
-For backups that give you the ability to do full disaster recovery, use full snapshots. For more information about full snapshots, see [Full Snapshots (recommended)](#full).
+### Create a Backup in the Admin Console {#admin-console-backup}
 
-Partial snapshots can only be created via the admin console (check screenshot below).
+To create a full backup in the admin console:
 
-![Application Backup UI](/images/snapshot-application-backup.png)
+1. Log in to the admin console and go to **Snapshots > Full Snapshots (Instance)**. 
+1. Click **Start a snapshot**.
```
jonquil2002, line 39: It kind of feels like we're leaving them hanging sometimes when there's no result sentence at the end. I think it doesn't matter when the result is obvious. With something like this, if you feel it's less obvious, you can mention that all of the completed snapshots get listed on this page.

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,91 @@
...
 
+   ![Full Snapshots page](/images/snapshot-instance-backup.png)
+
+## Create a Partial Backup {#partial}
+
+Partial backups, or _application snapshots_, back up only application volumes and application manifests. They do not include admin console data or application metadata.
```
jonquil2002, line 45: I posted a question in the Orange channel about what actually gets backed up a for partials, because this conflicts with what's in the admin console text. But this text is what we were told to clarify awhile back, and we had to dig hard to get this info. I want to be sure though.

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,91 @@
...
 
-## Full Snapshots (Recommended) {#full}
+## Create a Full Backup (Recommended) {#full}
 
-Full snapshots back up the admin console and all application data.
-They can be used for full Disaster Recovery by restoring over the same instance or into a new cluster. For limitations, see [Limitations and Considerations](snapshots-config-workflow#limitations-and-considerations) in _How to Set Up Backup Storage_.
+Full backups, or _instance snapshots_, back up the admin console, application volumes, application manifests, and application metadata. If you manage multiple applications with the admin console, then data from all applications that support backups is included in a full backup. You can use a full backup for disaster recovery scenarios by restoring over the same instance or into a new cluster.
 
-There are two ways to create a full snapshot. First, make sure that your license has the snapshots feature enabled, then:
+From a full backup, you can do any of the following types of restores:
```
jonquil2002, line 13: This list looks good. It's repeated info, so I'm wondering if we might want to make the bulleted list a partial or potentially move it to the "About" topic later. But you could merge this and then we can pair on it/iterate.

jonquil2002 (dismissed):
lgtm

docs/enterprise/snapshots-creating.md
paigecalvert: Great fixes

docs/enterprise/snapshots-creating.md
paigecalvert: Good question...I think it makes sense to be more specific if possible. Maybe the guideline could be to use the lowest level section name possible

docs/enterprise/snapshots-creating.md
paigecalvert: Removed this from the other procedures as well and made sure that "in the admin console" was in the stem sentence

docs/enterprise/snapshots-creating.md
paigecalvert: Ah good call, I like this idea

docs/enterprise/snapshots-creating.md
paigecalvert: Agreed, I decide to just replace this with a sentence that says you can do different types of restore, then pointed to your topic for more info

docs/enterprise/snapshots-creating.md
paigecalvert: Update: we learned that Salah does not know what was meant by "application metadata", so we removed this terminology from both the admin console and the docs

docs/enterprise/snapshots-storage-destinations.md
```diff
@@ -170,7 +170,7 @@ If no Velero installation is detected, instructions are displayed to install Vel
 ## Next Steps
 
 * (Existing clusters only) Configure Velero namespace access if you are using RBAC and optionally increase the default memory limits. See [Configuring Namespace Access and Memory Limit](snapshots-velero-installing-config).
```
ajp-io, line 172: We use different wording in the NFS and host path docs for this next step. We don't mention existing clusters only and RBAC in those cases. To be consistent, we could say something like this in all cases.
```suggestion
* (Existing clusters only) Configure Velero namespace access if you are using minimal RBAC, and optionally increase the default memory limits. See [Configuring Namespace Access and Memory Limit](snapshots-velero-installing-config).
```

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,93 @@
...
 
-## Full Snapshots (Recommended) {#full}
+## Create a Full Backup (Recommended) {#full}
 
-Full snapshots back up the admin console and all application data.
-They can be used for full Disaster Recovery by restoring over the same instance or into a new cluster. For limitations, see [Limitations and Considerations](snapshots-config-workflow#limitations-and-considerations) in _How to Set Up Backup Storage_.
+Full backups, or _instance snapshots_, back up the admin console and all application data, including volumes and manifest files. You can use a full backup for disaster recovery scenarios by restoring over the same instance or into a new cluster.
 
-There are two ways to create a full snapshot. First, make sure that your license has the snapshots feature enabled, then:
+If you manage multiple applications with the admin console, data from all applications that support backups is included in a full backup. Reach out to your software vendor for information about support for backup and restore.
```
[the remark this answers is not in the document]
ajp-io: I think that's true if each application includes a backup custom resource. @sgalsaleh is that right?

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,93 @@
...
+   ![Snapshot Settings and Schedule page](/images/snapshot-schedule.png)
+
+1. (Partial Backups Only) If you manage multiple applications in the admin console, use the dropdown to select the application that you want to back up.
+
+1. Select **Enable automatic scheduled snapshots**. 
+
+1. Configure the automatic backup schedule:
+
+   * For **Schedule**, select Hourly, Daily, Weekly, or Custom.
+   * For **Cron Expression**, enter a cron expression to customize the automatic backup schedule that you selected. For information about supported cron expressions, see [Cron Expressions](/reference/cron-expressions) in _Reference_.
```
ajp-io, line 84: You really only need to change the cron expression if you chose Custom. If you chose anything else, the cron expression is already correct, and if you change it, it'll switch you to Custom automatically.

docs/enterprise/snapshots-creating.md
```diff
@@ -1,40 +1,93 @@
...
+
+1. (Optional) For **Retention Policy**, edit the amount of time that data from the automatic backups is saved.
```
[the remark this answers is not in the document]
ajp-io: I asked that recently and was told all backups. My understanding is that all backups of that type (regardless of automatic or manual) are removed by the retention policy. But it does look like the retention policy is set per type (full or partial), so it would only remove the backups of that type.

@sgalsaleh again for confirmation.

A question I also have is whether you can not have a retention policy. Based on the UI, it doesn't look like you can disable a retention policy.

ajp-io (commented):
Looks good. One important thing that didn't stand out to me is that full backups can do what partial backups do and more. I think you could read this and say, if I want to do disaster recovery, I use full snapshots, and if I want to backup the app for a rollback situation, I use partial backups. But really you could do both with full snapshots and never need partial snapshots (which is why full are recommended). I'm not saying that wasn't present at all, but I feel like it could be a clearer takeaway so people are more likely to use full all the time.

docs/enterprise/snapshots-storage-destinations.md
paigecalvert: Good catch--thanks

docs/enterprise/snapshots-creating.md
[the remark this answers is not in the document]
sgalsaleh: Yes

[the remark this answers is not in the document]
ajp-io: So we should specify that too, @paigecalvert, that each app needs a backup resource.

docs/enterprise/snapshots-creating.md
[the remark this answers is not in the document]
sgalsaleh: Yes, the Retention Policy is set per type like Alex mentioned.

> A question I also have is whether you can not have a retention policy. Based on the UI, it doesn't look like you can disable a retention policy.

Velero docs do not mention a way to set backups to never expire. The default value is 30 days (720 hours). It can be configured to be in years, but I don't know/think there's a way to disable it completely.

docs/enterprise/snapshots-creating.md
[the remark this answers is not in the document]
paigecalvert: I have this story to add the details about Backup resources for multi-app to the vendor docs, since the vendor would be the one to create those resources: https://app.shortcut.com/replicated/story/69448/add-info-about-multi-app-backup-resources-to-vendor-docs

That said, maybe it's useful to tell the enterprise user that they could go check the files for each app to see if a Backup resources is there in order to confirm which apps are included in the backup?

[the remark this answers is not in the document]
paigecalvert: (I feel that like second sentence was confusing. LMK if I should rephrase)

jonquil2002 (approved):
lgtm!

docs/enterprise/snapshots-creating.md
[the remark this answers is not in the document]
ajp-io: Yeah, we could do that. It's a stinky UX to have to go check the files. You could also check the partial snapshots dropdown to see which apps you can do partial snapshots on. Also not ideal.

Unless the Snapshots tab only shows if you have an app selected that supports snapshots.

I'm realizing this is a very weird UX. Everything with multi-app seems a bit off :)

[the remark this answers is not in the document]
paigecalvert: Multi-app is indeed a bit goofy haha

PR #962, opened 2023-02-16
Add daemonset status informer info

ousbots:
Add information about the new daemonset status informer to the vendor and enterprise "Displaying Application Status" pages.

This is needed due to [this KOTS PR](https://github.com/replicatedhq/kots/pull/3636)

Things I'm not sure about:
* Are the column widths in the status table distributed in a visually appealing way? Maybe the leftmost column should shrink so everything else can be wider.
* Calling the pods in the daemonset daemons. They're really pods that run on each node, but they're also not replicas in the deployment or statefulset sense.

### DIFF
--- a/docs/enterprise/status-viewing-details.md
+++ b/docs/enterprise/status-viewing-details.md
@@ -27,6 +27,7 @@ The following resource types are supported for displaying application status:
 * Service
 * Ingress
 * PersistentVolumeClaims (PVC)
+* DaemonSet
 
 Applications can specify one or more of the supported Kubernetes workloads listed above. The app manager watches all specified workloads for state changes.
 
--- a/docs/partials/status-informers/_statusesTable.mdx
+++ b/docs/partials/status-informers/_statusesTable.mdx
@@ -1,11 +1,12 @@
 <table>
  <tr>
     <th width="10%"></th>
-    <th width="18%">Deployment</th>
-    <th width="18%">StatefulSet</th>
-    <th width="18%">Service</th>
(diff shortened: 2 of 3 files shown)


paigecalvert:
Thank you @ousbots ! This looks great.

- I see what you're saying about the table...I don't think it's a blocker to merging this PR, but I can look at some other ways to format this info (maybe we want to split it up into several smaller tables or lists, for example)
- For the question about what to call them, it seems like the Kubernetes docs are also undecided on this, haha. In the [topic about DaemonSets](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#communicating-with-daemon-pods), they use all these terms:
  - Daemon Pods
  - Pods
  - Daemons
It looks like probably any term is probably fine, but would you want to take a look at that topic and see if the any ways they use those terms in context resonates with you for our context?

paigecalvert (dismissed):
(Approving in case you decide you don't want to change the "daemon" term and would like to merge as is)

paigecalvert:
Hi @ousbots , just checking in on this PR. Is this ready to merge or are we waiting on a release?

ousbots:
Hi @paigecalvert sorry for the delay, we ran into a bug in reliably determining the updating state in the related kots PR, and I was waiting to resolve that before returning back here. I went ahead and updated the table to use the term "daemon pods" since it's a bit clearer. Sorry for the delay!

paigecalvert (approved):

paigecalvert:
No worries @ousbots ! Just re-approved

PR #974, opened 2023-02-21
Update Velero Install Procedures

paigecalvert:
https://app.shortcut.com/replicated/story/69790/add-step-for-running-velero-ensure-permissions-when-installing-velero-in-minimal-rbac-cluster and https://app.shortcut.com/replicated/story/57056/improve-velero-openshift-cluster-installation

### DIFF
--- a/docs/partials/snapshots/_installVelero.mdx
+++ b/docs/partials/snapshots/_installVelero.mdx
@@ -1,0 +1,27 @@
+1. Run the following command to install Velero:
+
+    ```bash
+    velero install \
+      --no-default-backup-location \
+      --no-secret \
+      --use-restic \
+      --use-volume-snapshots=false \
+      --plugins velero/velero-plugin-for-aws:v1.5.3
+    ```
+
+1. If your cluster uses one of the following Kubernetes distributions, then you must configure the Restic DaemonSet specification:
+     * RancherOS 
+     * OpenShift 
+     * Microsoft Azure 
+     * VMware Tanzu Kubernetes Grid Integrated Edition
+     
+    For information about how to configure the Restic DaemonSet, see [Configure Restic DaemonSet spec](https://velero.io/docs/v1.9/restic/#configure-restic-daemonset-spec) in the Velero documentation.
+
+1. (Namespace-Scoped RBAC Only) If you installed the app manager with the `--use-minimal-rbac` flag to enable namespace-spaced scoped RBAC permissions, run the following command to ensure that the app manager can access the velero namespace:
+
+   ```
+   kubectl kots velero ensure-permissions --namespace NAME --velero-namespace VELERO-NAME
+   ```
+   Replace:
+   * `NAME` with the namespace where the admin console is running
+   * `VELERO-NAME` with the namespace where Velero is installed
--- a/docs/enterprise/snapshots-configuring-hostpath.md
+++ b/docs/enterprise/snapshots-configuring-hostpath.md
@@ -1,3 +1,5 @@
+import InstallVelero from "../partials/snapshots/_installVelero.mdx"
+
 # Configuring a Host Path Storage Destination
 
 This topic describes how to install Velero and configure a host path as your storage destination for backups.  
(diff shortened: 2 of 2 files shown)


docs/partials/snapshots/_installVelero.mdx
```diff
@@ -0,0 +1,40 @@
...
+    * **Velero versions earlier than 1.10**:
+
+    ```bash
+    velero install \
+      --no-default-backup-location \
+      --no-secret \
+      --use-restic \
+      --use-volume-snapshots=false \
+      --plugins velero/velero-plugin-for-aws:v1.5.3
+    ```
```
paigecalvert, line 23: Updated this step ^ to include a command for velero 10.0 and later / earlier than 10.0. (Tested out both of these and it looks like they work, but could use confirmation)

docs/partials/snapshots/_installVelero.mdx
```diff
@@ -0,0 +1,40 @@
...
+    For information about how to configure the Restic DaemonSet, see [Configure Restic DaemonSet spec](https://velero.io/docs/v1.9/restic/#configure-restic-daemonset-spec) in the Velero documentation.
+
+1. (Namespace-Scoped RBAC Only) If you installed the app manager with the `--use-minimal-rbac` flag to enable namespace-spaced scoped RBAC permissions, run the following command to ensure that the app manager can access the Velero namespace:
+
+   ```
+   kubectl kots velero ensure-permissions --namespace NAME --velero-namespace VELERO-NAME
+   ```
+   Replace:
+   * `NAME` with the namespace where the admin console is running
+   * `VELERO-NAME` with the namespace where Velero is installed
```
paigecalvert, line 40: New step to call out running `ensure-permissions` if using minimal RBAC. 

Oddly, I tested this out with a minimal RBAC KOTS installation and it seemed like I was able to install velero/configure a host path storage destination even _without_ running `ensure-permissions`. So, not sure what was going on there...maybe I did something else weird.

But either way, this could use confirmation as well. 😄

docs/partials/snapshots/_installVelero.mdx
paigecalvert: Also just a note that the `velero install` instructions in the CLI output show the `--use-restic` flag, which I think won't work if they have 1.10+ installed?

docs/partials/snapshots/_installVelero.mdx
paigecalvert: Removed this step; Salah reminded me that this is in fact _not_ necessary 

>there shouldn’t be a need to run ensure-permissions if you’re configuring hostpath or nfs from the cli

docs/enterprise/snapshots-configuring-nfs.md
```diff
@@ -18,52 +21,56 @@ Complete the following items before you perform this task:
...
-Run the following command to install Velero and configure an NFS storage destination in an online environment. For more information about required storage destination flags, see [`velero`](/reference/kots-cli-velero-index) in _Reference_.
+To install Velero and configure NFS storage in an online environment:
 
-```bash
-kubectl kots velero configure-nfs --namespace NAME --nfs-path PATH --nfs-server HOST
-```
+<InstallVelero/>
 
-Replace:
+3. Run the following command to install Velero and configure an NFS storage destination in an online environment:
```
jonquil2002, line 30: It looks like Velero is installed in the first step and this step should just be about configuring the storage destination now.

docs/enterprise/snapshots-configuring-nfs.md
```diff
@@ -18,52 +21,56 @@ Complete the following items before you perform this task:
...
-- HOST with the hostname or IP address of the NFS server
-- PATH with the path that is exported by the NFS server
-- REGISTRY_HOSTNAME with the registry endpoint where the images are hosted
-- REGISTRY_NAMESPACE with the registry namespace where the images are hosted
-- REGISTRY_USERNAME with the username to use to authenticate with the registry
-- REGISTRY_PASSWORD with the password to use to authenticate with the registry
+<InstallVelero/>
 
-If no Velero installation is detected, instructions are displayed to install Velero and configure the storage destination.
+3. Run the following command to install Velero and configure an NFS storage destination in an air gapped environment.
```
jonquil2002, line 51: Same comment here: It looks like Velero is installed in the first step and this step should just be about configuring the storage destination now.

docs/partials/snapshots/_installVelero.mdx
```diff
@@ -0,0 +1,35 @@
+1. Run one of the following commands to install Velero, depending on the version of the Velero CLI that you installed as part of [Installing the Velero CLI](snapshots-velero-cli-installing):
```
jonquil2002, line 1: Should this be "velero CLI" instead of "Velero CLI"? We're not capitalizing the names of other CLIs.

docs/partials/snapshots/_registryCredentialsNote.mdx
```diff
@@ -0,0 +1,3 @@
+:::note
```
jonquil2002, line 1: It looks a little odd to not have an intro paragraph and just start a procedure with a note. I wonder if this note might fit better at the end of the step for installing Velero or keep it where it is but just make it a regular paragraph.

docs/partials/snapshots/_registryCredentialsNote.mdx
```diff
@@ -0,0 +1,3 @@
...
+It is typical for the velero and node-agent (restic) Pods to be in the `ErrImagePull` or `ImagePullBackOff` state after you run the `velero install` command because Velero does not support passing registry credentials during installation. In the app manager v1.94.0 and later, this situation resolves itself after you complete the instructions to configure the storage destination.
```
jonquil2002, line 2: What happens if they're on an earlier version of app manager? What might they need to do?

jonquil2002 (approved):
lgtm - a few questions/suggestions

docs/enterprise/snapshots-configuring-nfs.md
paigecalvert: Ahh yes indeed, thank you

docs/partials/snapshots/_installVelero.mdx
paigecalvert: yes, I will edit!

docs/partials/snapshots/_registryCredentialsNote.mdx
paigecalvert: Oh, I'm not sure. That's a good question. This note was here previously, I just added it to a partial.

docs/partials/snapshots/_registryCredentialsNote.mdx
paigecalvert: It definitely would be better to have it after that step...only reason I didn't was because I was using a partial and this note only applies to the air gap instructions.

I think I'll try splitting up the partial into two smaller partials to make this more flexible

docs/partials/snapshots/_registryCredentialsNote.mdx
paigecalvert: I created this follow up story for researching and updating this note: https://app.shortcut.com/replicated/story/70342/update-note-about-registry-credentials-for-air-gap-velero-install

PR #979, opened 2023-02-23
Updating upgrade info for embedded clusters

paigecalvert:
https://app.shortcut.com/replicated/story/69761/instructions-on-upgrading-kurl-by-more-than-2-minor-versions

Going to split this story into a couple PRs. This PR focuses on updating the info in the enterprise docs to explain what they'd need to request from the application vendor

Preview: https://deploy-preview-979--replicated-docs.netlify.app/enterprise/updating-embedded-cluster. Specifically, added this new section called [Multi-version Kubernetes Updates](https://deploy-preview-979--replicated-docs.netlify.app/enterprise/updating-embedded-cluster#kubernetes-multi) to clarify the process.


docs/enterprise/updating-embedded-cluster.md
```diff
@@ -5,27 +5,32 @@ import UpgradePrompt from "../partials/updating/_upgradePrompt.mdx"
...
+
+If the script detects that the version of Kubernetes in your cluster is more than two minor versions earlier than the target version, it prints an error message similar to the following: `The currently installed kubernetes version is 1.23.16. The requested version to upgrade to is 1.26.0. Kurl can only be upgraded two minor versions at time. Please install 1.25.x. first.`
+
+To continue with the upgrade, you must first update Kubernetes in your cluster to the version specified in the error message.
+
+To update Kubernetes, contact your application vendor for a different Kubernetes installer installation script that specifies a version of Kubernetes no more than two minor versions later than the version currently installed. For example, to upgrade from Kubernetes 1.23.x to 1.26.x, first run an installation script that specifies Kubernetes 1.25.x. Then, run the target installation script that specifies 1.26.x.
```
jonquil2002, line 33: This paragraph was a little challenging to follow. It might be clearer to say something like "To update Kubernetes when your currently installed version is more than two minor versions behind the target version, contact your vendor for the additional installation script needed to perform the prerequisite updates. For example, ..."

jonquil2002 (approved):
lgtm - so much better! one area of confusion noted.

docs/enterprise/updating-embedded-cluster.md
paigecalvert: Great call, this was pretty clunky. I also noticed in revisiting it that line 31 is probably unnecessary. 

What do you think about something like this?

>If the script detects that the version of Kubernetes in your cluster is more than two minor versions earlier than the target version, it prints an error message similar to the following: `The currently installed kubernetes version is 1.23.16. The requested version to upgrade to is 1.26.0. Kurl can only be upgraded two minor versions at time. Please install 1.25.x. first.`
>
>To update Kubernetes when your currently installed version is more than two minor versions behind the target version, contact your application vendor for an additional Kubernetes installer installation script that specifies the prerequisite Kubernetes version indicated in the error message.
>
>After you update Kubernetes in your cluster to the prerequisite version, you can continue with the upgrade by running the target installation script. For example, to upgrade from Kubernetes 1.23.x to 1.26.x, first run an installation script that specifies Kubernetes 1.25.x. Then, run the target installation script that specifies 1.26.x.

banjoh (approved):
LGTM

My only comment here is to clarify whether there is a different path that we haven't thought of say when an upgrade is detected by KOTS Admin Console. But this change covers the issue fully

PR #982, opened 2023-02-23
Add steps for removing an app and its resources

paigecalvert:


docs/enterprise/delete-admin-console.md
```diff
@@ -2,8 +2,9 @@
 
 This topic describes how to delete the Replicated admin console and remove installed applications, including information about:
 
-* Removing references for installed applications from the admin console.
-* Deleting Kubernetes objects and resources for the admin console from a cluster.
+* Removing references for installed applications from the admin console. See [Remove an Application Reference](#remove-an-application-reference).
+* Removing an application and all its resources from a cluster. See [Remove an Application](#remove-an-application).
+* Deleting Kubernetes objects and resources for the admin console from a cluster. See [Delete the Admin Console](#delete-the-admin-console)
```
ajp-io, line 7: IMO resource and object are used fairly interchangeably (I think there is technically slight nuance, but I don't fully grasp it in this moment), and I don't think we need to mention both here. I think just resources would suffice, and we only say resources in the above line about deleting the app.

docs/enterprise/delete-admin-console.md
```diff
@@ -13,6 +14,8 @@ When you use `kots remove`, the admin console no longer manages the application
 
 It can be useful to remove the reference to an application from the admin console if you want to reinstall the application, but you do not want to recreate the namespace or other Kubernetes resources. For example, you could use `kots remove` if you installed an application using an incorrect license file and need to reinstall with the correct license.
 
+For information about how to completely remove an application and its resources from a cluster, see [Remove an Application](#remove-an-application) below.
```
ajp-io, line 17: I can't comment on the line because it is unchanged, but above it says that `kots remove` doesn't delete any of the resources for the app. We probably need to clean that up (by default? or something), because it now can remove those resources if you pass the `--undeploy` flag.

docs/enterprise/delete-admin-console.md
```diff
@@ -31,6 +34,31 @@ To remove an application reference from the admin console:
...
+   kubectl kots get apps -n NAMESPACE
+   ```
+   Replace `NAMESPACE` with the name of the namespace where the admin console is installed.
+
+   In the output of this command, note the slug for the application that you want to remove.
+
+1. Run the following command to remove the application and all of its resources:
+
+   ```
+   kubectl kots remove APP_SLUG -n NAMESPACE --undeploy
```
sgalsaleh, line 55:
```suggestion
   kubectl kots remove APP_SLUG -n NAMESPACE --force --undeploy
```

docs/enterprise/delete-admin-console.md
```diff
@@ -31,6 +34,31 @@ To remove an application reference from the admin console:
    * `NAMESPACE` with the name of the namespace where the admin console is installed.
    * `APP_SLUG` with the slug for the application that you want to remove.
 
+
+## Remove an Application
+
+You can completely remove an installed application and all its resources from a cluster using the `kots remove` command with the `--undeploy` flag.
```
ajp-io, line 40: I worry that how this is written in two sections gives the impression that the command above removes the reference and this one removes the resources. But this removes the resources and the reference. It's like this is an extension of the above section. It's still removing the reference, but if you add `--undeploy` it'll also remove the resources.

Note that if you run this after first removing the reference alone, this will fail because the reference will no longer be present. So you can't run `kots remove` and then `kots remove --undeploy` on the same app. You have to choose which you want to do (keep resources or delete them).

docs/enterprise/delete-admin-console.md
```diff
@@ -31,6 +34,31 @@ To remove an application reference from the admin console:
...
+ 
+To remove an application and its resource from a cluster:
```
ajp-io, line 42:
```suggestion
To remove an application and its resources from a cluster:
```

docs/enterprise/delete-admin-console.md
```diff
@@ -13,6 +14,8 @@ When you use `kots remove`, the admin console no longer manages the application
 
 It can be useful to remove the reference to an application from the admin console if you want to reinstall the application, but you do not want to recreate the namespace or other Kubernetes resources. For example, you could use `kots remove` if you installed an application using an incorrect license file and need to reinstall with the correct license.
```
sgalsaleh, line 15: I would recommend having this be the first sentence/statement in this section since it clarifies why you'd wanna only remove the reference

docs/enterprise/delete-admin-console.md
ajp-io: Or we could say add `--force` optionally if the app was already deployed.

We should also mention this in the section above about just removing the reference (this flag is relevant there too).

docs/enterprise/delete-admin-console.md
```diff
@@ -1,43 +1,54 @@
...
+     ```
+     Replace:
+     * `APP_SLUG` is the slug for the application that you want to remove.
+     * `NAMESPACE` is the name of the namespace where the admin console is installed.
 
-When you install an application with the admin console, Replicated also creates the Kubernetes objects for the admin console itself on the cluster. The admin console includes Deployment and Service objects, Secret objects, and other objects such as Services, StatefulSets, and PersistentVolumeClaims.
+   * **Remove the reference to the application from the admin console and remove its resources from the cluster**:
+
+      ```
+      kubectl kots remove APP_SLUG -n NAMESPACE --undeploy
```
sgalsaleh, line 39: This almost always needs the `--force` flag since it's assumed that the application is deployed.
```suggestion
      kubectl kots remove APP_SLUG -n NAMESPACE --undeploy
```
Actually, @ajp-io should we automatically assume that the `--force` flag is passed if the user only passes the `--undeploy` flag?

ajp-io: I thought this same thing when I read it. I think that's a good idea: `--undeploy` passes `--force` automatically. Which means we don't need this change in the docs, but we can amend the note below to say that `--force` is implied if `--undeploy` is passed.

We'll need to update the docs for the CLI command too to mention that.

docs/enterprise/delete-admin-console.md
```diff
@@ -1,43 +1,54 @@
...
+      ```
+      
+   :::note
+   Optionally, use the `--force` flag to remove the application reference from the admin console when the application has already been deployed. For more information, see [remove](/reference/kots-cli-remove) in _kots CLI_.
```
ajp-io, line 43: Just a suggestion. Feel free to change the wording.
```suggestion
   Optionally, use the `--force` flag to remove the application reference from the admin console when the application has already been deployed. The `--force` flag is implied when `--undeploy` is used. For more information, see [remove](/reference/kots-cli-remove) in _kots CLI_.
```

docs/enterprise/delete-admin-console.md
```diff
@@ -52,7 +63,7 @@ These procedures do not uninstall the kots CLI. To uninstall the kots CLI, see [
 
 If you installed on an existing cluster and the admin console is not installed in the `default` namespace, you delete the admin console by using the kubectl Kubernetes command line tool to delete the namespace where the admin console is installed.
 
-After you delete the namespace, if `requireMinimalRBACPriviledges` is _not_ set to `true` in the Application custom resource manifest, delete the admin console ClusterRole and ClusterRoleBindings on the cluster.
+After you delete the namespace, if `requireMinimalRBACPrivileges` is _not_ set to `true` in the Application custom resource manifest, delete the admin console ClusterRole and ClusterRoleBindings on the cluster.
```
ajp-io, line 66: Just fixed this typo because I saw it. Can be a separate PR: we talk about if `requireMinimalRBACPrivileges` is `true`, but we also have `supportMinimalRBACPrivileges`, which means that minimal RBAC is determined at install time. Might need to mention that scenario here too. https://docs.replicated.com/reference/custom-resource-application#supportminimalrbacprivileges

paigecalvert: Ahh, yeah. This needs to be updated. I noticed a few mentions of this on my last pass but must have missed some others. We can probably get away with just saying "when namespace-scoped access is enabled" or "when minimal RBAC is enabled" here, the link off to that topic for more info

docs/enterprise/delete-admin-console.md
```diff
@@ -50,25 +61,29 @@ These procedures do not uninstall the kots CLI. To uninstall the kots CLI, see [
...
 
-After you delete the namespace, if `requireMinimalRBACPriviledges` is _not_ set to `true` in the Application custom resource manifest, delete the admin console ClusterRole and ClusterRoleBindings on the cluster.
+If you installed the admin console with namespace-scoped access, then the admin console Role and RoleBinding RBAC resources are also deleted when you delete the namespace. Alternatively, if you installed with the default cluster-scoped access, then you manually delete the admin console ClusterRole and ClusterRoleBindings resources from the cluster.
+
+The application vendor can require, support, or not support namespace-scoped installations. For more information, see [supportMinimalRBACPrivileges](/reference/custom-resource-application#supportminimalrbacprivileges) and [requireMinimalRBACPrivileges](/reference/custom-resource-application#requireminimalrbacprivileges) in _Application_.
+
+For more information about installing with cluster- or namespace-scoped access, see [RBAC Requirements](/installing-general-requirements#rbac-requirements) in _Requirements for Installation_.
```
paigecalvert, line 70: @ajp-io re-wrote this section to better explain minimal RBAC/namespace-scoped access^

feel free to suggest/make edits as needed

docs/enterprise/delete-admin-console.md
```diff
@@ -1,43 +1,54 @@
...
-   ```
-   kubectl kots remove APP_SLUG -n NAMESPACE
-   ```
-   Replace:
-   * `NAMESPACE` with the name of the namespace where the admin console is installed.
-   * `APP_SLUG` with the slug for the application that you want to remove.
+1. Run _one_ of the following commands:
 
-## Delete the Admin Console
+   * **Remove the reference to the application from the admin console only**:
```
ajp-io, line 27: With "only" at the end, could this sound like we're only removing it from the admin console and not from somewhere else?
```suggestion
   * **Remove only the reference to the application from the admin console**: 
```

docs/enterprise/delete-admin-console.md
```diff
@@ -50,25 +61,29 @@ These procedures do not uninstall the kots CLI. To uninstall the kots CLI, see [
...
 1. Run the following command to delete the namespace where the admin console is installed:
 
    :::note
    * You cannot delete the `default` namespace.
-   * This command deletes everything inside the specified namespace.
+   * This command deletes everything inside the specified namespace, including the admin console Role and RoleBinding resources if you installed with namespace-scoped access.
    :::
 
    ```
    kubectl delete ns NAMESPACE
```
ajp-io, line 82: What do you do if the admin console is in the default namespace? I don't see that accounted for here, since we acknowledge that you can't delete the default ns. (Maybe people don't often install in `default`.)

docs/enterprise/delete-admin-console.md
ajp-io: Looks good except for the link.
```suggestion
For more information about installing with cluster- or namespace-scoped access, see [RBAC Requirements](/enterprise/installing-general-requirements#rbac-requirements) in _Requirements for Installation_.
```

docs/enterprise/delete-admin-console.md
paigecalvert: Went back and forth on this; I like your rewrite better

docs/enterprise/delete-admin-console.md
paigecalvert: Yeah that's a good question, I'm not sure that we address that use case

ajp-io (approved):

ajp-io:
@paigecalvert This is good to merge, but it needs a docs review too.

paigecalvert:
I can merge as admin if you want, if the release is ready to go? @ajp-io

ajp-io:
It has been released. We're all set to merge.

docs/enterprise/delete-admin-console.md
```diff
@@ -1,43 +1,54 @@
 # Deleting the Admin Console and Removing Applications
 
-This topic describes how to delete the Replicated admin console and remove installed applications, including information about:
+This topic describes how to remove installed applications and delete the Replicated admin console from a cluster. See the following sections:
+* [Remove an Application](#remove-an-application)
```
jonquil2002, line 4: Style question for future consideration - do we need this list since it appears in the right hand nav? We haven't employed this consistently. Might we only want to employ this kind of a list when we're trying to list the multiple procedures in a process?

docs/enterprise/delete-admin-console.md
```diff
@@ -1,43 +1,54 @@
...
+1. Run _one_ of the following commands:
 
-## Delete the Admin Console
+   * **Remove only the reference to the application from the admin console**: 
+
+     ```
+     kubectl kots remove APP_SLUG -n NAMESPACE
+     ```
+     Replace:
+     * `APP_SLUG` is the slug for the application that you want to remove.
```
jonquil2002, line 33: I think this should be "with the slug..." instead of "is". Same comment for the following bullet.

jonquil2002 (dismissed):
lgtm

docs/enterprise/delete-admin-console.md
ajp-io: Yep, fixed

ajp-io:
My fix dismissed the review, but this should be good to merge now, except for the style comment from Jonquil.

jonquil2002:
@paigecalvert I'm going to merge this since you're out. I wasn't sure if this was tied to a release, so I didn't merge it earlier. But the style comment can be a future consideration.

PR #995, opened 2023-03-10
71228: Pod Volume Backups

jonquil2002:
SC: https://app.shortcut.com/replicated/story/71228/clarify-how-volumes-are-backed-up

and SC: https://app.shortcut.com/replicated/story/38953/kots-snapshots-documentation-suggests-mentions-using-opt-in-for-volume-backups-which-only-works-with-restic

Using the same PR for both of these SC because they are closely related.

Also, note that one of the SC stories addresses an issue that a vendor opened here: https://github.com/replicatedhq/kots.io/issues/598

Previews:

- https://deploy-preview-995--replicated-docs.netlify.app/vendor/snapshots-overview#pod-volume-data
- https://deploy-preview-995--replicated-docs.netlify.app/vendor/snapshots-configuring-backups

### DIFF
--- a/docs/vendor/snapshots-configuring-backups.md
+++ b/docs/vendor/snapshots-configuring-backups.md
@@ -2,7 +2,9 @@
 
 The snapshots feature is a backup and restore option that lets you define a manifest for creating snapshots and restoring previous snapshots. The backups include all of the annotated volumes in the archive. For more information, see [About Backup and Restore](snapshots-overview/).
 
-**Note:** If you are using multiple applications, repeat this procedure for each application. Every application must have its own Backup resource to be included in a full backup.
+:::note
+If you are using multiple applications, repeat this procedure for each application. Every application must have its own Backup resource to be included in a full backup.
+:::
 
 To configure backups:
 
--- a/docs/vendor/snapshots-configuring-backups.md
+++ b/docs/vendor/snapshots-configuring-backups.md
@@ -1,12 +1,16 @@
 # Configuring Backups
 
-The snapshots feature is a backup and restore option that lets you define a manifest for executing snapshots and restoring previous snapshots.
+The snapshots feature is a backup and restore option that lets you define a manifest for creating snapshots and restoring previous snapshots. The backups include all of the annotated volumes in the archive. For more information, see [About Backup and Restore](snapshots-overview/).
 
-1. To enable snapshots:
+**Note:** If you are using multiple applications, repeat this procedure for each application. Every application must have its own Backup resource to be included in a full backup.
 
-    1. Add a backup resource to the application using the Velero manifest. The following minimal YAML example enables snapshots in the application. When a snapshot is executed in the admin console or by a schedule, snapshots will include all annotated volumes in the archive (see the additional information on annotating volumes later in this task).
+To configure backups:
 
-        Example:
+1. Enable snapshots:
+
+    1. Add a Backup resource (`kind: Backup`) using `apiVersion: velero.io/v1` to the application manifest files. The following minimal YAML example enables snapshots in the application. For more information about Backup resource options, see [backup resource](/reference/custom-resource-backup) in _Reference_.
+
+        **Example:**
 
         ```yaml
         apiVersion: velero.io/v1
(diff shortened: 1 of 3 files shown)


docs/vendor/snapshots-overview.md
```diff
@@ -18,21 +18,31 @@ The app manager also exposes hooks that can be used to inject scripts to execute
...
+* **Full snapshots (Recommended)**: Backs up the admin console and all application data, including KOTS-specific object-stored data. For embedded clusters, this also backs up the Docker registry, which is required for air gapped installations.
 
 * **Partial snapshots**: Backs up the application volumes and manifests only. This is a legacy feature and does not support disaster recovery because it only backs up the application.
 
-For limitations and considerations, see [Limitations and Considerations](/enterprise/snapshots-understanding#limitations-and-consoderations).
+For limitations and considerations, see [Limitations and Considerations](/enterprise/snapshots-understanding#limitations-and-considerations).
+
+### Pod Volume Data
+
+By default, you must opt-in to have pod volumes backed up using restic (restic is referred to as File System Backup or FSB in the Velero documentation). In the Backup resource that you configure to enable snapshots, you must annotate each specific volume that you want to back up.
```
jonquil2002, line 29: @sgalsaleh In this case, do we need to refer to this as node-agent instead of restic? Velero docs refer to this as FSB in this case, and we're not sure how and when these terms (FSB versus node-agent) should be used.

docs/vendor/snapshots-overview.md
```diff
@@ -18,21 +18,31 @@ The app manager also exposes hooks that can be used to inject scripts to execute
...
+
+You can use a Velero-compatible plug-in, such as the VMware Tanzu `velero-plugin-for-gcp`, instead of restic. In this case, Velero backs up everything, not just the annotated volumes that you want to include. Backing up everything can, in turn, cause incomplete restores. As a best practice when using plug-ins, after you opt-in the volumes that you want to include, add a label to exclude the volumes that you do not want to back up.
```
sgalsaleh, line 31: Using plug-ins other than restic to back up volumes is not supported, we should not tell vendors/customers to do that. So I don't think this sentence is necessary.

docs/vendor/snapshots-configuring-backups.md
```diff
@@ -64,9 +65,14 @@ The snapshots feature is a backup and restore option that lets you define a mani
     ```
 
 1. (Optional) Configure manifest exclusions. By default, Velero also includes snapshots of all of the Kubernetes objects in the namespace.
-To exclude any manifest, add a `velero.io/exclude-from-backup` label to the manifest to be excluded.
 
-    Example:
+  :::note
+  If you are using a Velero-compatible plug-in, such as the VMware Tanzu `velero-plugin-for-gcp`, add a label to exclude volumes that you do not want to backup. Otherwise, Velero backups up all volumes even if you only annotated a few volumes for inclusion, which can cause incomplete restores. For more information about plug-ins, see [Pod Volume Data](snapshots-overview/#pod-volume-data) in _About Backup and Restore_.
```
sgalsaleh, line 70: Using plug-ins other than restic to back up volumes is not supported, we should not tell vendors/customers to do that. So I don't think this sentence is necessary.

docs/vendor/snapshots-overview.md
jonquil2002: Ok, I removed this plug-in info in both topics and added a statement that we support only restic as a backup program.

docs/vendor/snapshots-configuring-backups.md
jonquil2002: Ok, I removed this plug-in info in both topics and added a statement that we support only restic as a backup program.

sgalsaleh (dismissed):

docs/vendor/snapshots-configuring-backups.md
```diff
@@ -1,12 +1,16 @@
 # Configuring Backups
 
-The snapshots feature is a backup and restore option that lets you define a manifest for executing snapshots and restoring previous snapshots.
+The snapshots feature is a backup and restore option that lets you define a manifest for creating snapshots and restoring previous snapshots. The backups include all of the annotated volumes in the archive. For more information, see [About Backup and Restore](snapshots-overview/).
 
-1. To enable snapshots:
+**Note:** If you are using multiple applications, repeat this procedure for each application. Every application must have its own Backup resource to be included in a full backup.
```
paigecalvert, line 5: :::note instead of **note ?

docs/vendor/snapshots-overview.md
```diff
@@ -18,21 +18,28 @@ The app manager also exposes hooks that can be used to inject scripts to execute
...
+
+### Pod Volume Data
+
+Replicated snapshots supports the restic backup program only.
+
+By default, Velero requires that you opt-in to have pod volumes backed up. In the Backup resource that you configure to enable snapshots, you must annotate each specific volume that you want to back up. For more information about including and excluding pod volumes, see [Configuring Backups](snapshots-configuring-backups). 
+
+### Object-Stored Data
 
+If other types of object-stored data (not KOTS-specific) is being used that does not use a persistentVolumeClaim (PVC), then you must write custom backup and restore hooks for your users to be able to back up object-stored data. For example, Rook and Ceph do not use PVCs and require custom backup and restore hooks. For more information, see [Configuring Backup and Restore Hooks](snapshots-hooks).
```
paigecalvert, line 35: What does KOTS-specific mean in this context? The object store that KOTS uses?

paigecalvert (dismissed):
Looks great! Just a couple small comments

docs/vendor/snapshots-overview.md
jonquil2002: Reverted this paragraph back to its original position for context, which seems KOTS add-on specific in that context. We can do a follow-up story to get that area better defined.

PR #1013, opened 2023-03-20
remove outdated minio vulnerability disclosures

cbodonnell:
This PR removes an outdated section for MinIO vulnerability disclosures.  We updated to the latest version of MinIO in [KOTS 1.95.0](https://docs.replicated.com/release-notes/rn-app-manager#improvements-1-95-0), so we can remove these.

The entire section has been removed since it only had MinIO disclosures.  It's possible we want to keep the `## Vulnerability Management Exceptions` header, so open to suggestions here.


tcalotta (approved):
looks good to me

paigecalvert (approved):
thanks @cbodonnell and @tcalotta

PR #1055, opened 2023-04-04
73080: Add a Change Private Registry for Air Gap Procedure

jonquil2002:
SC: https://app.shortcut.com/replicated/story/73080/document-how-to-change-private-registry-settings-for-air-gap-after-installation

Previews:
-https://deploy-preview-1055--replicated-docs.netlify.app/enterprise/image-registry-settings#configure-private-registries-for-online-clusters
- https://deploy-preview-1055--replicated-docs.netlify.app/enterprise/image-registry-settings#air-gap

### DIFF
--- a/docs/enterprise/image-registry-settings.md
+++ b/docs/enterprise/image-registry-settings.md
@@ -42,7 +42,7 @@ To change private registry settings for an air gap cluster:
 
 1. On the Registry Settings tab in the admin console, select the **Disable Pushing Images to Private Registry** checkbox, and click **Save changes**. This is a temporary action that allows you to edit the registry namespace and hostname.
 
-  If you only want to change the username or password, you can skip this step.
+  If you only want to change the username or password, you do not have to disable pushing the images.
 
 1. Edit the fields as needed, and click **Save changes**.
 
--- a/docs/enterprise/image-registry-settings.md
+++ b/docs/enterprise/image-registry-settings.md
@@ -1,3 +1,5 @@
+import ImageRegistrySettings from "../partials/image-registry/_image-registry-settings.mdx"
+
 # Using Private Registries
 
 This topic describes how to configure private registry settings in the Replicated admin console.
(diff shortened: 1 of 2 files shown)


cbodonnell (approved):

docs/enterprise/image-registry-settings.md
```diff
@@ -12,17 +14,39 @@ Private registries can be used with online or air gap clusters. For embedded clu
 
 Your domain must support a Docker V2 protocol. For more information, see [Private Registry Requirements](installing-general-requirements#private-registry-requirements) in _Requirements for Installation_.
 
-## Configure Private Registry Settings
+## Configure Private Registries for Online Clusters
```
paigecalvert, line 17: I really like the idea of using the verbs "Configure" and "Change" to indicate that for air gap installations, you would never just now be configuring the registry for the first time.

paigecalvert: Maybe only suggestion would be to use the preposition "in" instead of "for". When it says "for a cluster", I think it sounds like it's suggesting that the registry is for the whole cluster, rather than just being for the given application.

docs/enterprise/image-registry-settings.md
```diff
@@ -12,17 +14,39 @@ Private registries can be used with online or air gap clusters. For embedded clu
...
+
+1. Click **Test Connection** to test the connection between the admin console and the private registry host.
+
+1. Click **Save changes**.
+
+1. (Optional) Click **Stop using registry** if you want to remove the registry settings from the app manager.
+
+## Change Private Registries for Air Gap Clusters {#air-gap}
+
+During air gap installations, you configure a private registry for the app manager to push images to. You can change the private registry settings at any time in the admin console.
```
paigecalvert, line 39: suggestion: During air gap installations, you configure a private registry ~~for the app manager to push images to.~~ where the app manager pushes images.

docs/enterprise/image-registry-settings.md
```diff
@@ -12,17 +14,39 @@ Private registries can be used with online or air gap clusters. For embedded clu
...
 You configure the application to use a private registry in the admin console, unless you are using a kURL Registry add-on. You can also disable the connection or remove the registry settings if needed.
 
-To configure private registry settings:
+For information about changing registry settings in an air gap environment, see [Change Private Registries for Air Gap Clusters](#air-gap).
+
+To configure private registry settings for a online cluster:
 
 1. On the Registry Settings tab in the admin console, edit the fields:
 
      ![Registry Settings](/images/registry-settings.png)
```
paigecalvert, line 27: If it's not too much of a pain, recommend making this image a little smaller (If it's not easy to change the size of the actual png, then could use the HTLM img tag with a width attribute

paigecalvert (approved):

docs/enterprise/image-registry-settings.md
jonquil2002: Markdown kept fighting me on accepting the size earlier when I edited it in SnagIt - but now I tried the HTML and it worked! Never thought of that. We should check and see if we can make changes to our CSS so we don't have to mess around with HTML though.

PR #1062, opened 2023-04-06
Clarify steps to support both helm CLI and kots installations

paigecalvert:
Preview: https://deploy-preview-1062--replicated-docs.netlify.app/vendor/helm-image-registry (See the new section on supporting both types of installations. Also removed content in other sections to reduce noise and clutter)

Story: https://app.shortcut.com/replicated/story/73789/update-organization-to-reduce-noise-and-improve-skim-ability-of-the-using-private-registries-for-helm-cli-installations-topic

### DIFF
--- a/docs/vendor/helm-image-registry.md
+++ b/docs/vendor/helm-image-registry.md
@@ -12,7 +12,9 @@ Using an external private image registry or the Replicated private registry for 
 
 In addition to the image pull secret, using an external private registry also requires that you add credentials for the registry to the vendor portal so that Replicated can access the image through the Replicated proxy service.
 
-For installations that use the helm CLI rather than the app manager, Replicated cannot automatically inject an image pull secret nor patch the image name to reference the proxy service in the Helm chart for your application. You can complete the following procedures to use a private registry for helm CLI installations:
+For installations that use the helm CLI rather than the app manager, Replicated cannot automatically inject an image pull secret nor patch the image name to reference the proxy service in the Helm chart for your application.
+
+To use a private registry for helm CLI installations, complete the following procedures:
 1. [Deliver Image Pull Secrets](#pull-secret)
 1. [Reference the Proxy Service](#proxy-service)
 1. [(Optional) Support Both Helm CLI and App Manager Installations](#helm-and-kots)
@@ -117,9 +119,7 @@ To deliver customer-specific image pull secrets for a private registry:
 This procedure is required only for external private registries. Skip this procedure if you are using the Replicated private registry.
 :::
 
-To reference the proxy service, first create a field in the Helm chart `values.yaml` file with the URL of the location of the private image on `proxy.replicated.com`. Use template functions to reference this field from the Pod spec in the Helm chart. This allows Replicated to access the external registry through the proxy service when your users install with the helm CLI.
-
-Then, update the Replicated HelmChart custom resource manifest file with the URL for the image on your external registry. Adding this URL to the HelmChart custom resource ensures that the Replicated proxy service can automatically patch the image name to reference `proxy.replicated.com` during KOTS Install or Embedded Cluster installations.
+This procedure describes how to update your Helm chart to allow Replicated to access the external registry through the proxy service when your users install with the helm CLI.
 
 To update the image name to reference the proxy service:
 
@@ -175,7 +175,7 @@ To update the image name to reference the proxy service:
 
 ## (Optional) Step 3: Support Both Helm CLI and App Manager Installations {#helm-and-kots}
 
-To support both helm CLI and app manager (KOTS Install and Embedded Cluster) installations from the same release using the same image registry, you must add a static value with your registry URL to the HelmChart custom resource for app manager installations.
+To support both helm CLI and app manager installations with your private registry from the same release, you must update the release to add your registry URL as a static value in the HelmChart custom resource.
 
 This allows the Replicated proxy service to automatically patch the image name to reference `proxy.replicated.com` for app manager installations. Helm Install installations ignore this static value, and instead use the `proxy.replicated.com` URL that you added to the `values.yaml` file previously.
 
@@ -222,4 +222,4 @@ To support both helm CLI and app manager installations:
        apiImageRepository: quay.io/my-org/api:v1.0.1
    ```
 
-1. Save and promote the release to a development environment to test your changes.
+1. Save and promote the release to a development environment to test that you can install with both the helm CLI and the app manager (KOTS Install or Embedded Cluster).
--- a/docs/vendor/helm-image-registry.md
+++ b/docs/vendor/helm-image-registry.md
@@ -18,6 +18,7 @@ To use a private registry for helm CLI installations, complete the following pro
 1. [Deliver Image Pull Secrets](#pull-secret)
 1. [Reference the Proxy Service](#proxy-service)
 1. [(Optional) Support Both Helm CLI and App Manager Installations](#helm-and-kots)
+
 ## Step 1: Deliver Image Pull Secrets {#pull-secret}
 
 When the app manager installs an application, Replicated automatically uses the customer license to create and inject an image pull secret that is specific to a customer's license. For helm CLI installations, you must use a template function to inject a customer-specific image pull secret.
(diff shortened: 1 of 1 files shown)


ousbots (dismissed):
This looks good to me, better organized and much easier to understand when and why the registry value should be overriden in the HelmChart resource (using both cli and kots to install)

docs/vendor/helm-image-registry.md
```diff
@@ -1,37 +1,38 @@
...
 
-In addition to the image pull secret, using an external private registry also requires that you add credentials for the registry to the vendor portal so that Replicated can access the image through the Replicated proxy service. For kots CLI and Kubernetes installer installations, Replicated automatically patches the image name to reference the location of the image on the proxy service at `proxy.replicated.com`. For more information about this process, see [About Using an External Registry](/vendor/private-images-about).
+In addition to the image pull secret, using an external private registry also requires that you add credentials for the registry to the vendor portal so that Replicated can access the image through the Replicated proxy service.
 
 For installations that use the helm CLI rather than the app manager, Replicated cannot automatically inject an image pull secret nor patch the image name to reference the proxy service in the Helm chart for your application.
```
jonquil2002, line 15: I know it was written like this before. But, it feels like a sudden jump between this sentence and the next. This sentence, as it stands alone, makes it sound like nothing can be done. How about a segue of some kind, maybe something like "For Helm CLI installations, Replicated cannot automatically inject an image pull secret nor patch the image name to reference the proxy service in the Helm chart for your application, so additional configuration is needed."

docs/vendor/helm-image-registry.md
```diff
@@ -1,37 +1,38 @@
...
+## Step 1: Deliver Image Pull Secrets {#pull-secret}
 
-## Deliver Image Pull Secrets for a Private Registry {#pull-secret}
-
-To inject an image pull secret for an external private registry or the Replicated private registry, first add the Replicated `LicenseDockerCfg` template function to the Helm chart `values.yaml` file. The `LicenseDockerCfg` template function renders a value based on the unique customer license when the Helm chart is pulled. Write this rendered value to a pull secret, then reference the pull secret in the necessary template files for the Helm chart.
-
-For more information about the `LicenseDockerCfg` template function, see [LicenseDockerCfg](/reference/template-functions-license-context#licensedockercfg) in _License Context_.
+When the app manager installs an application, Replicated automatically uses the customer license to create and inject an image pull secret that is specific to a customer's license. For helm CLI installations, you must use a template function to inject a customer-specific image pull secret.
```
jonquil2002, line 24: I think this first sentence is not needed. It seems confusing since it's not related to the procedure, and the overview section already states this comparison. You could just go with "For helm CLI installations, you must use a template function to inject a customer-specific image pull secret." and it would likely be clear enough.

docs/vendor/helm-image-registry.md
```diff
@@ -1,37 +1,38 @@
...
 For installations that use the helm CLI rather than the app manager, Replicated cannot automatically inject an image pull secret nor patch the image name to reference the proxy service in the Helm chart for your application.
 
-To use private images with your application for helm CLI installations, use Helm's template functions to inject an image pull secret into the Helm chart. If you are using an external private registry, you must also update references to the image name in your Helm chart to use the location of the image on the Replicated proxy service.
+To use a private registry for helm CLI installations, complete the following procedures:
+1. [Deliver Image Pull Secrets](#pull-secret)
+1. [Reference the Proxy Service](#proxy-service)
+1. [(Optional) Support Both Helm CLI and App Manager Installations](#helm-and-kots)
 
-For more information, see [Deliver Image Pull Secrets for a Private Registry](#pull-secret) and [Update the Image Name to Reference the Proxy Service](#proxy-service) below.
+## Step 1: Deliver Image Pull Secrets {#pull-secret}
```
jonquil2002, line 22: We're introducing a new model here with Step 1, Step 2, etc. for the procedures in a process. We haven't done that elsewhere. In some ways I like it, but in some ways it's confusing. I think it works best in a tutorial rather than a generic procedure. I think it's particularly confusing at the end of this procedure when the last step tells them to go to Step 2 next. Food for thought.

docs/vendor/helm-image-registry.md
```diff
@@ -174,7 +170,19 @@ To update the image name to reference the proxy service:
...
 1. Package your Helm chart and add the packaged chart to a release in the Replicated vendor portal. For more information, see [Creating Releases with Helm Charts](helm-release).
 
-1. In the release, create or open the HelmChart custom resource manifest file. A HelmChart custom resource manifest file has `kind: HelmChart` and `apiVersion: kots.io/v1beta1`.
+1. Save and promote the release to a development environment to test your changes.
+
+1. To use the same private registry for both helm CLI and app manager installations, continue to [(Optional) Step 3: Support Both Helm CLI and App Manager Installations](#helm-and-kots) below.
+
+## (Optional) Step 3: Support Both Helm CLI and App Manager Installations {#helm-and-kots}
+
+To support both helm CLI and app manager installations with your private registry from the same release, you must update the release to add your registry URL as a static value in the HelmChart custom resource.
```
jonquil2002, line 179: Is there a use case that we can give for when they'd be supporting both types of installations? And is this the similar to when they might have a standard manifest release to which they add Helm charts (probably not similar, but had to ask)? I'm not sure if vendors would be confused by this, but I sure am wondering about the use case.

jonquil2002 (approved):
lgtm - had a few suggestions and questions.

docs/vendor/helm-image-registry.md
paigecalvert: Good call

docs/vendor/helm-image-registry.md
paigecalvert: I like that-- I was struggling for a good segue here

docs/vendor/helm-image-registry.md
paigecalvert: Do you feel like that description would be helpful? And/or moving the conceptual info from that helm CLI topic would be helpful?

jonquil2002: Cool, thanks for the explanation. And feel free to move that content if you feel it fits there. It sounds right.

jonquil2002: Oh yes, re-reading that https://docs.replicated.com/vendor/helm-install#about sections, it's more about supporting multiple types of installation and might really fit better in the About Deploying Helm Charts topic.

docs/vendor/helm-image-registry.md
```diff
@@ -174,7 +170,19 @@ To update the image name to reference the proxy service:
...
+1. Save and promote the release to a development environment to test your changes.
+
+1. To use the same private registry for both helm CLI and app manager installations, continue to [(Optional) Support Both Helm CLI and App Manager Installations](#helm-and-kots) below.
+
+## (Optional) Support Both helm CLI and App Manager Installations {#helm-and-kots}
+
+As an application vendor, you can support both helm CLI and app manager installations from the same release. For more information, see [About Supporting helm CLI installations](helm-install) in _Supporting helm CLI Installations (Beta)_.
+
+To support both types of installations with your private registry from a single release, you must update the release to add your registry URL as a static value in the HelmChart custom resource. This allows the Replicated proxy service to automatically patch the image name to reference `proxy.replicated.com` for app manager installations. Helm Install installations ignore this static value, and instead use the `proxy.replicated.com` URL that you added to the `values.yaml` file previously.
```
cbodonnell, line 181: Only thing I noticed.  It's KOTS that patches the image to pull through the proxy. This one makes it sounds like the proxy service is what is changing the image reference.
```suggestion
To support both types of installations with your private registry from a single release, you must update the release to add your registry URL as a static value in the HelmChart custom resource. This allows the app manager to automatically patch the image name to reference `proxy.replicated.com`. Helm Install installations ignore this static value, and instead use the `proxy.replicated.com` URL that you added to the `values.yaml` file previously.
```

paigecalvert: good catch, thank you!

jonquil2002 (approved):
lgtm

PR #1063, opened 2023-04-07
72532: Improve Content in Optional Value Keys

jonquil2002:
SC: https://app.shortcut.com/replicated/story/72532/rewrite-helm-topic-optional-value-keys

Preview: https://deploy-preview-1063--replicated-docs-upgrade.netlify.app/vendor/helm-optional-value-keys

### DIFF
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -73,11 +73,12 @@ spec:
     - when: "repl{{ ConfigOptionEquals `mariadb_type` `external`}}"
       recursiveMerge: false
       values:
-        externalDatabase.host: "repl{{ ConfigOption `external_db_host`}}"
-        externalDatabase.user: "repl{{ ConfigOption `external_db_user`}}"
-        externalDatabase.password: "repl{{ ConfigOption `external_db_password`}}"
-        externalDatabase.database: "repl{{ ConfigOption `external_db_database`}}"
-        externalDatabase.port: "repl{{ `external_ db_port`}}"
+        externalDatabase:
+          host: "repl{{ ConfigOption `external_db_host`}}"
+          user: "repl{{ ConfigOption `external_db_user`}}"
+          password: "repl{{ ConfigOption `external_db_password`}}"
+          database: "repl{{ ConfigOption `external_db_database`}}"
+          port: "repl{{ `external_ db_port`}}"
 
 
   # builder values provide a way to render the chart with all images
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -66,17 +66,17 @@ spec:
   # these values will be supplied to helm template
   values:
     mariadb:
-      enabled: repl{{ ConfigOptionEquals `mariadb_type` `embeddedDatabase`}}
+      enabled: repl{{ ConfigOptionEquals `external_db_type` `embeddedDatabase`}}
 
   optionalValues:
-    - when: "repl{{ ConfigOptionEquals `mariadb_type` `embeddedDatabase`}}"
+    - when: "repl{{ ConfigOptionEquals `external_db_type` `embeddedDatabase`}}"
       recursiveMerge: false
       values:
-        externalDatabase.host: "repl{{ ConfigOption `external_mariadb_host`}}"
-        externalDatabase.user: "repl{{ ConfigOption `external_mariadb_user`}}"
-        externalDatabase.password: "repl{{ ConfigOption `external_mariadb_password`}}"
-        externalDatabase.database: "repl{{ ConfigOption `external_mariadb_database`}}"
-        externalDatabase.port: "repl{{ `external_mariadb_port`}}"
+        externalDatabase.host: "repl{{ ConfigOption `external_db_host`}}"
+        externalDatabase.user: "repl{{ ConfigOption `external_db_user`}}"
+        externalDatabase.password: "repl{{ ConfigOption `external_db_password`}}"
+        externalDatabase.database: "repl{{ ConfigOption `external_db_database`}}"
+        externalDatabase.port: "repl{{ `external_ db_port`}}"
 
 
   # builder values provide a way to render the chart with all images
@@ -91,9 +91,17 @@ The HelmChart YAML above results in the following `values.yaml` if a user select
 mariadb:
   enabled: false
   externalDatabase:
-    host: "repl{{ ConfigOption `external_mariadb_host`}}"
-    user: "repl{{ ConfigOption `external_mariadb_user`}}"
-    password: "repl{{ ConfigOption `external_mariadb_password`}}"
-    database: "repl{{ ConfigOption `external_mariadb_database`}}"
-    port: "repl{{ `external_mariadb_port`}}"
+    host: "repl{{ ConfigOption `external_db_host`}}"
+    user: "repl{{ ConfigOption `external_db_user`}}"
+    password: "repl{{ ConfigOption `external_db_password`}}"
+    database: "repl{{ ConfigOption `external_db_database`}}"
+    port: "repl{{ `external_db_port`}}"
 ```
+The HelmChart YAML above results in the following `values.yaml` if the user selects `embeddedDatabase`:
+
+```
+mariadb:
+  enabled: true
+```
+
+For more information about the `optionalValues` property, including details about the when and `recursiveMerge` fields, see [`optionalValues`](https://docs.replicated.com/reference/custom-resource-helmchart#optionalvalues) in _HelmChart_.
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -17,7 +17,7 @@ Some advanced use cases involve writing values to a values file only if there is
 
 **Example:**
 
-For example, in the Bitnami Wordpress [chart.yaml.](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/Chart.yaml) file, there is a reference to `mariadb`. This is configured through the [values.yaml](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/values.yaml#L1086) file:
+For example, in the Bitnami Wordpress [chart.yaml.](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/Chart.yaml) file, there is a reference to `mariadb`. This reference is configured through the [values.yaml](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/values.yaml#L1086) file:
 
 ```yaml
 mariadb:
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -25,10 +25,11 @@ mariadb:
   ## To use an external database set this to false and configure the `externalDatabase.*` parameters
   ##
   enabled: true
-  ## External Database Configuration
-  ## All of these values are only used if `mariadb.enabled=false`
-  ##
-  externalDatabase:
+...
+## External Database Configuration
+## All of these values are only used if `mariadb.enabled=false`
+##
+externalDatabase:
   ## @param externalDatabase.host External Database server host
   ##
   host: localhost
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -9,7 +9,7 @@ It is possible to either remove or include values when certain conditions are me
 
 If the Helm chart `values.yaml` contains a static value that must be removed when deploying using the Replicated app manager, set the value to `"null"` (including the quotation marks) in the HelmChart custom resource manifest file.
 
-Typically this situation happens when you are including a community chart as a dependency in your own chart. You cannot control how the community chart is built and structured, and you might want to change some of the default behavior that the community chart does not easily expose. For example, Kubernetes only supports one livenessProbe handler. You can set a default livenessProbe handler to `"null"` and enable your preferred handler. For more information, see [Deleting a Default Key](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key) in the Helm documentation.
+Typically this situation happens when you are including a community chart as a dependency in your own chart. You cannot control how the community chart is built and structured, and you might want to change some of the default behavior that the community chart does not easily expose. For more information, see [Deleting a Default Key](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key) in the Helm documentation.
 
 ## Include Values
 
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -17,7 +17,7 @@ Some advanced use cases involve writing values to a values file only if there is
 
 **Example**
 
-For example, in the Bitnami Wordpress[chart.yaml.](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/Chart.yaml) chart, there is a reference to mariadb. This is configured through the [values.yaml](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/values.yaml#L1086):
+For example, in the Bitnami Wordpress [chart.yaml.](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/Chart.yaml) chart, there is a reference to `mariadb`. This is configured through the [values.yaml](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/values.yaml#L1086):
 
 ```yaml
 mariadb:
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -78,7 +78,7 @@ spec:
           user: "repl{{ ConfigOption `external_db_user`}}"
           password: "repl{{ ConfigOption `external_db_password`}}"
           database: "repl{{ ConfigOption `external_db_database`}}"
-          port: "repl{{ `external_ db_port`}}"
+          port: "repl{{ ConfigOption `external_ db_port`}}"
 
 
   # builder values provide a way to render the chart with all images
--- a/docs/vendor/helm-optional-value-keys.md
+++ b/docs/vendor/helm-optional-value-keys.md
@@ -97,7 +97,7 @@ mariadb:
     database: "repl{{ ConfigOption `external_db_database`}}"
     port: "repl{{ `external_db_port`}}"
 ```
-The HelmChart YAML above results in the following `values.yaml` if the user selects `embeddedDatabase`:
+The HelmChart YAML above results in the following `values.yaml` if the user selects a `mariadb_type` of `embedded`:
 
 ```
 mariadb:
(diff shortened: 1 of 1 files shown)


docs/vendor/helm-optional-value-keys.md
```diff
@@ -13,18 +13,18 @@ For more information, see [Deleting a Default Key](https://helm.sh/docs/chart_te
...
 
-Helm charts such as Postgres do not treat an empty value (`""`) in the same way as if the key were missing from the chart. This can be necessary for applications packaged as a Helm chart that include a reference to another chart in the `requirements.yaml`.
+Additionally, Helm charts such as Postgres do not treat an empty value (`""`) in the same way as if the key were missing from the chart. This can be necessary for applications packaged as a Helm chart that include a reference to another chart in the `requirements.yaml`.
 
 For example, in the [requirements.yaml](https://github.com/helm/charts/blob/e64112e0913db99227926b49fa0ae59158c9c9d9/stable/sentry/requirements.yaml) of the [Sentry](https://github.com/helm/charts/tree/master/stable/sentry) chart, there is a reference to postgres. This is configured through the [Sentry chart values.yaml](https://github.com/helm/charts/blob/e64112e0913db99227926b49fa0ae59158c9c9d9/stable/sentry/values.yaml#L192):
 
 ```yaml
 # PostgreSQL chart configs
 postgresql:
```
jonquil2002, line 24: This YAML is now configure differently in the Sentry spec, and I'm wondering whether we should update our example and if it impacts our explanation. The new YAML is this:

```yaml
postgresql:
  enabled: true
  nameOverride: sentry-postgresql
  postgresqlDatabase: sentry
  postgresqlUsername: postgres
  # Only used when internal PG is disabled
  # postgresqlHost: postgres
  # postgresqlPassword: postgres
  # postgresqlPort: 5432
```

jonquil2002:
@cbodonnell Which static values in the Helm chart values.yaml  must be removed when deploying using the Replicated app manager. How would a vendor know which ones to set to `"null"`?

docs/vendor/helm-optional-value-keys.md
cbodonnell: We may want to update these examples in general since `requirements.yaml` is part of the older "v1" helm API that was used with Helm v2 (the [repo](https://github.com/helm/charts) even states that it's no longer maintained).

Bitnami hosts a repository with many "v2" API charts (that go with Helm v3): https://github.com/bitnami/charts

docs/vendor/helm-optional-value-keys.md
```diff
@@ -13,18 +13,18 @@ For more information, see [Deleting a Default Key](https://helm.sh/docs/chart_te
 
 ## Include Values
 
-Some advanced use cases involve writing values to a values file only if there is a value to apply.
+Some advanced use cases involve writing values to a values file only if there is a value to apply. For example, if a customer has the option to configure a different value than the default value in your file, such as setting the database option they want to use. You can include an optional value key so that your application can dynamically add and remove the default value key based on a `when` condition.
```
cbodonnell, line 16: The general idea as I understand it is that optional values are either included or not included depending on whether the condition is satisfied.  If the condition is not satisfied, the chart value itself is not removed, but instead the optional value provided is not used and it will fall back to whatever the default is (which might be nothing if there is no default).

```suggestion
Some advanced use cases involve writing values to a values file only if there is a value to apply. For example, if a customer has the option to configure a different value than the default value in your file, such as setting the database option they want to use. You can include an optional value key so that your application can dynamically include it based on a `when` condition.
```

cbodonnell:
> @cbodonnell Which static values in the Helm chart values.yaml must be removed when deploying using the Replicated app manager. How would a vendor know which ones to set to `"null"`?
> 
> Also, please see my other questions. Thanks.

@jonquil2002 Usually this would happen if a vendor was including a community chart as a dependency in their own chart.  They cannot control how that community chart is built and structured, but they might want to change some of the default behavior in ways that the chart does not easily expose.  The helm docs have an example: https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key

In that example the community chart defines a default liveness probe:
```
livenessProbe:
  httpGet:
    path: /user/login
    port: http
  initialDelaySeconds: 120
```

The vendor wants to change the definition of the liveness probe by passing `--set livenessProbe.exec.command=[cat,docroot/CHANGELOG.txt]`, however that would result in an invalid liveness probe since Kubernetes only supports one liveness probe handler (cannot have both `httpGet` and `exec`):
```
livenessProbe:
  httpGet:
    path: /user/login
    port: http
  exec:
    command:
    - cat
    - docroot/CHANGELOG.txt
  initialDelaySeconds: 120
```

So they are able to remove the `httpGet` section all together by also passing `--set livenessProbe.httpGet=null`.

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
   # these values will be supplied to helm template
   values:
-    postgresql:
-      enabled: repl{{ ConfigOptionEquals `postgres_type` `embedded_postgres`}}
+    mariadb:
+      enabled: repl{{ ConfigOptionEquals `external_db_type` `embeddedDatabase`}}
 
   optionalValues:
-    - when: "repl{{ ConfigOptionEquals `postgres_type` `external_postgres`}}"
+    - when: "repl{{ ConfigOptionEquals `external_db_type` `embeddedDatabase`}}"
```
jonquil2002, line 72: Based on the change in the values above, I changed this value to `external_db_type` also.  Is that correct?

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
   # builder values provide a way to render the chart with all images
   # and manifests. this is used in replicated to create airgap packages
   builder:
-    postgresql:
+    mariadb:
       enabled: true
 ```
-
-The above YAML will result in the following `values.yaml` if the user selects `external_postgres`:
+The HelmChart YAML above results in the following `values.yaml` if a user selects `external_db_type`:
```
jonquil2002, line 88: @cbodonnell Again, I change this to `external_db_type` for consistency. Is it correct?

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
-  postgresqlPassword: "repl{{ ConfigOption `external_postgres_password`}}"
-  postgresqlPort: "repl{{ ConfigOption `external_postgres_port`}}"
+  externalDatabase:
+    host: "repl{{ ConfigOption `external_db_host`}}"
+    user: "repl{{ ConfigOption `external_db_user`}}"
+    password: "repl{{ ConfigOption `external_db_password`}}"
+    database: "repl{{ ConfigOption `external_db_database`}}"
+    port: "repl{{ `external_db_port`}}"
 ```
+The HelmChart YAML above results in the following `values.yaml` if the user selects `embeddedDatabase`:
```
jonquil2002, line 100: @cbodonnell  I added this in to mirror what we had in the original topic and to clarify what happens when a user selects the embedded database option.

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
 
 If the Helm chart `values.yaml` contains a static value that must be removed when deploying using the Replicated app manager, set the value to `"null"` (including the quotation marks) in the HelmChart custom resource manifest file.
 
-For more information, see [Deleting a Default Key](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key) in the Helm documentation.
+Typically this situation happens when you are including a community chart as a dependency in your own chart. You cannot control how the community chart is built and structured, and you might want to change some of the default behavior that the community chart does not easily expose. For example, Kubernetes only supports one livenessProbe handler. You can set a default livenessProbe handler to `"null"` and enable your preferred handler. For more information, see [Deleting a Default Key](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key) in the Helm documentation.
```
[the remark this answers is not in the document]
cbodonnell: I like this piece.  I'd only suggest removing the last sentences about liveness probes since that is a specific example that may vary from chart to chart.  The helm docs show that example and give context around it as it pertains to a specific example chart, so I think the first bit is probably all we need here.

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
+  port: 3306
+  ## @param externalDatabase.user External Database username
+  ##
+  user: bn_wordpress
+  ## @param externalDatabase.password External Database user password
+  ##
+  password: ""
+  ## @param externalDatabase.database External Database database name
+  ##
+  database: bitnami_wordpress
```
cbodonnell, line 46: `externalDatabase` is it's own values section outside of `mariadb`, so I just modified the indentations for a few lines.  It's also not _right_ next to the `mariadb.enabled` part, so in my suggestion I added an ellipses to indicate this.  Not sure if we have prior examples of where we omit sections for brevity like that.
```suggestion
mariadb:
  ## @param mariadb.enabled Deploy a MariaDB server to satisfy the applications database requirements
  ## To use an external database set this to false and configure the `externalDatabase.*` parameters
  ##
  enabled: true
...
## External Database Configuration
## All of these values are only used if `mariadb.enabled=false`
##
externalDatabase:
  ## @param externalDatabase.host External Database server host
  ##
  host: localhost
  ## @param externalDatabase.port External Database server port
  ##
  port: 3306
  ## @param externalDatabase.user External Database username
  ##
  user: bn_wordpress
  ## @param externalDatabase.password External Database user password
  ##
  password: ""
  ## @param externalDatabase.database External Database database name
  ##
  database: bitnami_wordpress
```

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
+    name: wordpress
+    chartVersion: 15.3.2
 
   # values are used in the customer environment, as a pre-render step
   # these values will be supplied to helm template
   values:
-    postgresql:
-      enabled: repl{{ ConfigOptionEquals `postgres_type` `embedded_postgres`}}
+    mariadb:
+      enabled: repl{{ ConfigOptionEquals `external_db_type` `embeddedDatabase`}}
```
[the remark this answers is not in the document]
cbodonnell: I think having `mariadb_type` makes sense because the config option would be a radio button toggle where the end-user would choose either `embeddedDatabase` or `externalDatabase`.  If it's `embeddedDatabase`, then this would evaluate to `true` and the wordpress chart would deploy it's own mariadb instance.  If not, then it would evaluate to `false`, the chart would not deploy a mariadb instance and the optional values below would come into play.

docs/vendor/helm-optional-value-keys.md
cbodonnell: Same as above, regarding the field name.

For this one, we want to check if the config option equals `externalDatabase` since we only want these optional values being applied if the user chose to use their own external mariadb instance.

```suggestion
    - when: "repl{{ ConfigOptionEquals `external_db_type` `externalDatabase`}}"
```

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
-          postgresqlDatabase: "repl{{ ConfigOption `external_postgres_database`}}"
-          postgresqlUsername: "repl{{ ConfigOption `external_postgres_username`}}"
-          postgresqlHost: "repl{{ ConfigOption `external_postgres_host`}}"
-          postgresqlPassword: "repl{{ ConfigOption `external_postgres_password`}}"
-          postgresqlPort: "repl{{ ConfigOption `external_postgres_port`}}"
+        externalDatabase.host: "repl{{ ConfigOption `external_db_host`}}"
+        externalDatabase.user: "repl{{ ConfigOption `external_db_user`}}"
+        externalDatabase.password: "repl{{ ConfigOption `external_db_password`}}"
+        externalDatabase.database: "repl{{ ConfigOption `external_db_database`}}"
+        externalDatabase.port: "repl{{ `external_ db_port`}}"
```
cbodonnell, line 79:
```suggestion
        externalDatabase:
          host: "repl{{ ConfigOption `external_db_host`}}"
          user: "repl{{ ConfigOption `external_db_user`}}"
          password: "repl{{ ConfigOption `external_db_password`}}"
          database: "repl{{ ConfigOption `external_db_database`}}"
          port: "repl{{ `external_ db_port`}}"
```

docs/vendor/helm-optional-value-keys.md
cbodonnell: This would result if a user selects an `external_db_type` of `embeddedDatabase`.

Or a `mariadb_type` of `embeddedDatabase` if we go with the original name.

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
-  postgresqlUsername: "repl{{ ConfigOption `external_postgres_username`}}"
-  postgresqlHost: "repl{{ ConfigOption `external_postgres_host`}}"
-  postgresqlPassword: "repl{{ ConfigOption `external_postgres_password`}}"
-  postgresqlPort: "repl{{ ConfigOption `external_postgres_port`}}"
+  externalDatabase:
+    host: "repl{{ ConfigOption `external_db_host`}}"
+    user: "repl{{ ConfigOption `external_db_user`}}"
+    password: "repl{{ ConfigOption `external_db_password`}}"
+    database: "repl{{ ConfigOption `external_db_database`}}"
+    port: "repl{{ `external_db_port`}}"
```
cbodonnell, line 98: just some indentation fixes similar to the other one above since `externalDatabase` is it's own section
```suggestion
externalDatabase:
  host: "repl{{ ConfigOption `external_db_host`}}"
  user: "repl{{ ConfigOption `external_db_user`}}"
  password: "repl{{ ConfigOption `external_db_password`}}"
  database: "repl{{ ConfigOption `external_db_database`}}"
  port: "repl{{ `external_db_port`}}"
```

docs/vendor/helm-optional-value-keys.md
cbodonnell:
```suggestion
The HelmChart YAML above results in the following `values.yaml` if a user selects a `mariadb_type` of `external`:
```

docs/vendor/helm-optional-value-keys.md
cbodonnell:
```suggestion
The HelmChart YAML above results in the following `values.yaml` if the user selects a `mariadb_type` of `embedded`:
```

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,99 @@ It is possible to either remove or include values when certain conditions are me
...
-
-```yaml
-postgresql:
+```
+mariadb:
   enabled: true
 ```
-For more information about the `optionalValues` property, including details about the `when` and `recursiveMerge` fields, see [optionalValues](../reference/custom-resource-helmchart#optionalvalues) in _HelmChart_.
+
+For more information about the `optionalValues` property, including details about the when and `recursiveMerge` fields, see [`optionalValues`](https://docs.replicated.com/reference/custom-resource-helmchart#optionalvalues) in _HelmChart_.
```
cbodonnell, line 107: small one
```suggestion
For more information about the `optionalValues` property, including details about the `when` and `recursiveMerge` fields, see [`optionalValues`](https://docs.replicated.com/reference/custom-resource-helmchart#optionalvalues) in _HelmChart_.
```

docs/vendor/helm-optional-value-keys.md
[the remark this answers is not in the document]
cbodonnell: I think having `mariadb_type` makes sense because the config option would be a radio button toggle _named_ `mariadb_type` where the end-user would choose either the value `embeddedDatabase` or `externalDatabase`.  Thinking about it more, this is somewhat confusing since the field in the `values.yaml` is also named `externalDatabase`.  How about we change the example so that the config option is named `mariadb_type` and the possible values are just `embedded` or `external` to make that differentiation more clear?

If it's `embedded`, then this template function would evaluate to `true` and the wordpress chart would deploy it's own mariadb instance.  If it's `external`, then it would evaluate to `false`, the chart would not deploy a mariadb instance and the optional values below would come into play.

```suggestion
      enabled: repl{{ ConfigOptionEquals `mariadb_type` `embedded`}}
```

docs/vendor/helm-optional-value-keys.md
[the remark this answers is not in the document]
jonquil2002: Cool, done, thanks. It's good to have a second opinion.

docs/vendor/helm-optional-value-keys.md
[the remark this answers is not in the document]
jonquil2002: Thank you so much for the ideas and for explaining the logic. This is what I was struggling to understand.

cbodonnell (dismissed):

docs/vendor/helm-optional-value-keys.md
```diff
@@ -1,4 +1,4 @@
-# Including Optional Value Keys
+# Configuring Optional Value Keys
 
 > This topic applies to both native Helm and Replicated Helm installations.
```
paigecalvert, line 3: I did a search, and it looks like this is the only topic that has a  `> This topic applies to both native Helm and Replicated Helm installations.` blurb. Perhaps this could be combined with `The Helm chart values.yaml file is not a static mapping in an application. It is possible to either remove or include values when certain conditions are met` into more of the standard topic sentence format that we use elsewhere?

docs/vendor/helm-optional-value-keys.md
```diff
@@ -9,91 +9,101 @@ It is possible to either remove or include values when certain conditions are me
 
 If the Helm chart `values.yaml` contains a static value that must be removed when deploying using the Replicated app manager, set the value to `"null"` (including the quotation marks) in the HelmChart custom resource manifest file.
```
paigecalvert, line 10: The heading of this section is "Remove Values", but I think it sounds like it's actually _excluding_ the values from a given deployment. In contrast, removing might suggest that something already was deployed and now we want to remove it. Maybe the section could be called Exclude Values instead? This would also mirror the format of the "Include Values"

Similarly: 
>If the Helm chart `values.yaml` contains a static value that must be ~~removed~~ excluded when deploying using the Replicated app manager

cbodonnell (dismissed):

paigecalvert (dismissed):
Great updates @jonquil2002 , just a couple comments

docs/vendor/helm-optional-value-keys.md
jonquil2002: Good idea.

docs/vendor/helm-optional-value-keys.md
jonquil2002: You're right, it's not being removed. Helm actually calls this "deleting a default key", then describes it as "overriding". It's a good idea to change it, and I've chosen "Override Values". Thanks for catching this.

jonquil2002:
P.S. Checked cross-references and nothing is broken by changing from "Remove Values" to "Override Values".

PR #1068, opened 2023-04-12
70155: Add Native Helm Deployment Flowchart

jonquil2002:
SC: https://app.shortcut.com/replicated/story/70155/provide-a-flow-chart-explaining-how-kots-post-processes-helm-chart-resources-through-helm-kustomize

Preview: https://deploy-preview-1068--replicated-docs-upgrade.netlify.app/vendor/helm-overview#app-manager


paigecalvert (commented):
Flowchart looks good 👍 

Only note is that, especially since the five boxes in the diagram don't match the 5 steps in the process, I think this probably just needs a bit of a description of what exactly is being depicted in each part of the flowchart.

Something like, "As shown in the diagram above, given a Helm chart, the app manager uses Kustomize to merge instructions from Replicated and the end user to chart resources, packages the resulting manifest files into a new Helm chart, then deploys the Helm chart." (This could probably use some editing and SME review).

For context, it looks like steps 2,3,4 in the process correspond to the Kustomize box.
Step 5 corresponds to the manipulated helm chart and deploy boxes in the diagram.

Might help to frame this a little more for users so they understand what the diagram is showing, and then also that the process is giving a little more color to what is going on in the Kustomzie phase.

paigecalvert (approved):

PR #1074, opened 2023-04-13
73576: Add CLI procedures for Admin Console Setup

jonquil2002:
SC: https://app.shortcut.com/replicated/story/73576/add-cli-steps-to-install-topics

Previews:

Please see step 4 in this procedure for the additional CLI step that was added:
- https://deploy-preview-1074--replicated-docs.netlify.app/enterprise/installing-embedded-cluster#online

And also minor changes are in these topics:

- https://deploy-preview-1074--replicated-docs.netlify.app/enterprise/installing-app-setup
- https://deploy-preview-1074--replicated-docs.netlify.app/enterprise/installing-existing-cluster
- https://deploy-preview-1074--replicated-docs.netlify.app/enterprise/installing-existing-cluster-airgapped#air-gap
- https://deploy-preview-1074--replicated-docs.netlify.app/vendor/packaging-using-tls-certs#get-the-tls-secret


docs/enterprise/installing-embedded-cluster.md
```diff
@@ -64,7 +62,30 @@ To install the application:
...
 1. Note the `Kotsadm` and `Login with password (will not be shown again)` fields in the output of the installation command. 
 
-1. Log in to the admin console to complete the application setup, run preflight checks, and deploy. See [Completing Application Setup and Deploying](installing-app-setup).
+1. Install the application using one of the following methods:
+
+    - **Admin console:** Use the `kotsadm` and password from the previous step to log into the admin console. For information about using the admin console, see [Deploying the Application using the Admin Console](installing-app-setup).
+
+    - **kots CLI:**
+
+        1. Install the kots CLI. See [Install without Root Access](/reference/kots-cli-getting-started#install-without-root-access) in the _Installing the kots CLI_ reference section.
```
sgalsaleh, line 71: this is installed automatically as part of the `kotsadm` add-on, so it doesn't need to be installed.

jonquil2002: Thanks for catching that. I'm not sure why it was part of the embedded cluster install to begin with, but removing that from both.

docs/enterprise/installing-existing-cluster-airgapped.md
```diff
@@ -99,6 +99,6 @@ To install the application:
 
     After the `kots install` command installs the admin console and the application on the cluster, it creates a port forward to the admin console. The admin console is exposed internally on the cluster and can only be accessed using a port forward.
```
paigecalvert, line 100: This does not appear in the changed files view, but I noticed that the existing cluster air gap procedure says "Install the admin console using the images that you pushed in the previous step:" in step 5. For consistency, maybe "admin console" should be changed to "app manager"

paigecalvert (dismissed):
Looks good @jonquil2002 , just a couple comments with minor suggestions for the sake of consistency

docs/enterprise/installing-existing-cluster-airgapped.md
jonquil2002: I'm a little hesitant to make this change because the previous step says "Run the following command to extract admin console container images and push them into a private registry:" and has the command `kubectl kots admin-console push-images`. In fact, the entire procedure talks about installing the admin console. I suggest we create a separate SC to ask some questions and see if we can make these changes safely.

PR #1076, opened 2023-04-14
Fix typo

red-patience:
Fix a misspelling of "registry" in an example command.


jonquil2002 (approved):
lgtm - Thank you so much for catching that, @red-patience ! Much appreciated.

PR #1080, opened 2023-04-14
72547: Improve HelmChart Reference

jonquil2002:
SC: https://app.shortcut.com/replicated/story/72547/rewrite-the-helmchart-reference-topic

### DIFF
--- a/docs/reference/custom-resource-helmchart.md
+++ b/docs/reference/custom-resource-helmchart.md
@@ -2,16 +2,13 @@ import VersionLimitation from "../partials/helm/_helm-version-limitation.mdx"
 
 # HelmChart
 
-A HelmChart custom resource enables the Replicated app manager to process and deploy Helm charts as part of an application. The app manager can process Helm charts using the Replicated Helm installation for existing applications. You can also use a native Helm installation for new applications or newly added Helm charts. For more information, see [About Deploying Helm Charts](/vendor/helm-overview).
-
-HelmChart custom resources are required for the app manager to deploy Helm charts. HelmChart custom resources are not required if only raw Kubernetes manifests are deployed.
+HelmChart custom resources are required for the app manager to process and deploy Helm charts for the supported Helm installation types. For more information about Helm installation types, see [About Deploying Helm Charts](/vendor/helm-overview).
 
 The HelmChart custom resource manifest file references a required `.tgz` export of the Helm chart resources and provides the necessary instructions for processing and preparing the chart for deployment.
 
-By default, the HelmChart custom resource uses the Replicated Helm installation, which uses the app manager to render and deploy Helm charts. For new installations, you can set `useHelmInstall: true` in the manifest to use the native Helm installation.
-
-**Deploying multiple instances of the same chart**:
-You must add an additional HelmChart custom resource with a unique [release name](custom-resource-helmchart#chartreleasename) for each instance of the chart that is to be deployed as part of the application. However, only one `.tgz` of the chart needs to be included in the release.
+:::note
+To deploy multiple instances of the same chart, you must add an additional HelmChart custom resource with a unique [release name](custom-resource-helmchart#chartreleasename) for each instance of the chart that is to be deployed as part of the application. However, only one `.tgz` of the chart needs to be included in the release.
+:::
 
 The following is an example manifest file for the HelmChart custom resource:
 
@@ -110,15 +107,13 @@ Acceptable values are `v2` or `v3`. `v3` is the default when no value is specifi
 
 ## useHelmInstall
 
-Identifies whether this Helm chart will use the Replicated Helm installation (`false`) or native Helm installation (`true`).
-Default is `false`.
-It is recommended that you set this to `true`, because native Helm is preferred and Replicated Helm will be removed at some point.
+Identifies whether this Helm chart will use the Replicated Helm installation (`false`) or native Helm installation (`true`). We recommend that you set this to `true`, because native Helm is the preferred method. **Default:** `false`
+
 Native Helm installations always deploy using Helm v3.
 
 ## weight
 
-Determines the order to apply charts that have `useHelmInstall: true`. Charts are applied by weight in ascending order, with lower weights applied first.
-Acceptable values are positive and negative integers, with a default value of 0.
+Determines the order to apply charts that have `useHelmInstall: true`. Charts are applied by weight in ascending order, with lower weights applied first. Acceptable values are positive and negative integers. **Default:** `0`
 
 For more information, see [Defining Installation Order for Native Helm Charts](/vendor/helm-native-helm-install-order).
 


docs/reference/custom-resource-helmchart.md
```diff
@@ -82,43 +83,39 @@ The `chart` key allows for a mapping between the data in this definition and the
...
+
+The release name must be unique across all charts deployed in the namespace. Specifying a unique release name allows you to deploy multiple instances of the same Helm chart.
+
 Must be a valid Helm release name that matches regex `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` and is no longer than 53 characters.
 
 ## helmVersion
 
-Identifies the Helm Version used to render the chart.
-Acceptable values are `v2` or `v3`. `v3` is the default when no value is specified.
+Identifies the Helm Version used to render the chart. **Supported value:** `v3`
```
jonquil2002, line 104: @paigecalvert WDYT of this change?  I think it doesn't make sense to list V2 as an acceptable or supported value, especially with the note.

docs/reference/custom-resource-helmchart.md
```diff
@@ -1,17 +1,18 @@
...
+HelmChart custom resources are required for the app manager to process and deploy Helm charts for the supported Helm installation types. For more information about Helm installation types, see [About Deploying Helm Charts](/vendor/helm-overview).
 
 The HelmChart custom resource manifest file references a required `.tgz` export of the Helm chart resources and provides the necessary instructions for processing and preparing the chart for deployment.
 
-By default, the HelmChart custom resource uses the Replicated Helm installation, which uses the app manager to render and deploy Helm charts. For new installations, you can set `useHelmInstall: true` in the manifest to use the native Helm installation.
+:::note
+To deploy multiple instances of the same chart, you must add an additional HelmChart custom resource with a unique [release name](custom-resource-helmchart#chartreleasename) for each instance of the chart that is to be deployed as part of the application. However, only one `.tgz` of the chart needs to be included in the release.
+:::
```
paigecalvert, line 13: Perhaps this note could be removed, relocated, or reduced (which sounds like a very depressing evolution of "reduce reuse recycle")

One important piece of information in this note that could exist at a high level is "one HelmChart CR per helm chart". Maybe that could just be a sentence tacked on to the paragraph above.

There's also some information about how to use release name to deploy multiple instances of the same chart within a single release? I don't really understand that use case honestly, but maybe that info could go down in the release name section?

docs/reference/custom-resource-helmchart.md
paigecalvert: Good question...I see what you mean. I suppose it's _technically_ supported by app manager in that using v2 won't break anything, even if Helm ended support for v2. I might lean towards including v2 since there are people who might still have it in there , but let's see what the App Installer team thinks on this. I'm not sure I'm up-to-date on the latest in terms of what we'll support for helm v2

paigecalvert: And looking at the helmInstall details below, it also looks like it says that native helm always uses v3. I wonder if that means that if I put `useHelmInstall: true` and `helmVersion: v2` that my release won't work? That could also factor into how we express which versions are supported. (I.e., sounds like it could be accurate to say it's definitely not supported for native helm)

docs/reference/custom-resource-helmchart.md
```diff
@@ -155,11 +151,15 @@ For more options, see the [Allow deletion of a previous values file key](https:/
...
+The  attribute is a value for making optional charts. The `exclude` attribute can be parsed by template functions. 
 
 When the app manager processes Helm charts, it excludes the entire chart if the output of the `exclude` field can be parsed as a boolean evaluating to `true`.
 
-For more information about how the app manager processes Helm charts, see [Helm processing](../vendor/helm-processing).
+For more information about optional charts, template functions, and how app manager processes Helm charts, see:
+
+* [Optional Charts](/vendor/helm-optional-charts)
+* [About Template Function Contexts](template-functions-about)
+* [Helm Processing](/vendor/helm-processing)
```
paigecalvert, line 162: Nice link consolidation 😏

docs/reference/custom-resource-helmchart.md
```diff
@@ -190,7 +192,7 @@ For more information about the syntax for template functions, see [About Templat
 
 The `recursiveMerge` boolean defines how the app manager merges the `values` and `optionalValues` datasets when the conditional statement in the `when` field is true.
 
-The admin console uses the values from this merged dataset and from the Helm chart `values.yaml` file when deploying the application.         
+The admin console uses the values from this merged dataset and from the Helm chart `values.yaml` file when deploying the application. For an example of recursive merge, see [Example: Recursive Merge](/vendor/helm-optional-value-keys#example-recursive-merge) in _Configuring Optional Value Keys_.
```
paigecalvert, line 195: Good calling pulling out that example. Really simplifies the content here

docs/vendor/helm-overview.md
```diff
@@ -143,12 +143,35 @@ For more information about how to package an application with Replicated so that
...
 
-To create an `.airgap` bundle for a release that uses Helm charts, the Replicated vendor portal renders templates of the Helm charts with `helm template`. To specify which values from the Helm chart `values.yaml` file are included in the `.airgap` bundle, you add a `builder` key in the HelmChart custom resource manifest file. For more information, see [builder](/reference/custom-resource-helmchart#builder) in _HelmChart_.
+To create an `.airgap` bundle for a release that uses Helm charts, the Replicated vendor portal renders templates of the Helm charts with `helm template`. To specify which values from the Helm chart `values.yaml` file are included in the `.airgap` bundle, you add a `builder` key in the HelmChart custom resource manifest file. For more information, see [builder](/reference/custom-resource-helmchart#builder) in the _HelmChart_ reference.
 
 :::note
 The helm CLI installation method does not support installations into air gap environments. See [helm CLI Limitations](#helm-cli-limitations) below.
 :::
 
+<HelmBuilderRequirements/>
```
paigecalvert, line 152: It looks like this topic is missing the import HelmBuilderRequirements line at the top, so the partial is not rendering (I really like that you've brought this content over here. )

paigecalvert (dismissed):

docs/reference/custom-resource-helmchart.md
jonquil2002: Good idea.

docs/reference/custom-resource-helmchart.md
jonquil2002: @cbodonnell WDYT of the way it's written now? I changed it so that v2 is the only supported value and that native Helm only uses v3. Because f the nte, it really impies that anyone still using v2 should upgrade to v3. But is there a reason why I should still list v2 as an acceptable/supported value?

docs/reference/custom-resource-helmchart.md
```diff
@@ -82,43 +79,39 @@ The `chart` key allows for a mapping between the data in this definition and the
...
 
 > Introduced in Replicated app manager v1.73.0
 
-Specifies the release name to be used when installing this instance of the Helm chart.
+Specifies the release name to use when installing this instance of the Helm chart.
 Defaults to the chart name.
-The release name must be unique across all charts deployed in the namespace.
-Specifying a unique release name allows you to deploy multiple instances of the same Helm chart.
+
+The release name must be unique across all charts deployed in the namespace. To deploy multiple instances of the same Helm chart in a release, you must add an additional HelmChart custom resource with a unique [release name](custom-resource-helmchart#chartreleasename) for each instance of the Helm chart.
```
jonquil2002, line 94: @cbodonnell I moved the note into this paragraph.I also did a little rewrite about the multiple instances and moved the info about the unique name down to the `chart.releasename` section. PTAL and see if it's ok.

docs/reference/custom-resource-helmchart.md
```diff
@@ -1,17 +1,14 @@
...
+# HelmChart
 
-The HelmChart custom resource manifest file references a required `.tgz` export of the Helm chart resources and provides the necessary instructions for processing and preparing the chart for deployment.
+HelmChart custom resources are required for the app manager to process and deploy Helm charts for the supported Helm installation types. For more information about Helm installation types, see [About Deploying Helm Charts](/vendor/helm-overview).
 
-By default, the HelmChart custom resource uses the Replicated Helm installation, which uses the app manager to render and deploy Helm charts. For new installations, you can set `useHelmInstall: true` in the manifest to use the native Helm installation.
+The HelmChart custom resource manifest file references a required `.tgz` export of the Helm chart resources and provides the necessary instructions for processing and preparing the chart for deployment. To deploy multiple instances of the same Helm chart, you must add an additional HelmChart custom resource for each instance. However, only one `.tgz` of the Helm chart needs to be included in the release.
```
jonquil2002, line 9: @cbodonnell I moved the note into this paragraph.I also did a little rewrite about the multiple instances and moved the info about the unique name down to the `chart.releasename` section. PTAL and see if it's ok.

docs/reference/custom-resource-helmchart.md
jonquil2002: @craig - WDYT about the supported value being onlyv3 and the clarification that native Helm only supports v3? I made this change because we recommend that they migrate to v3 from v2 in the note. Is there a reason why we'd want to say that v2 is an acceptable/supported value?

docs/reference/custom-resource-helmchart.md
cbodonnell: This looks good to me.  Only thing I would call out is that the note about the unique release name in the `chart.releasename` section has a link to that same section, which is probably not necessary since the reader would already be there.

docs/reference/custom-resource-helmchart.md
cbodonnell: @jonquil2002 We haven't dropped support for `v2` _yet_ in the code, so I would hesitate to remove it entirely from docs.  If/when we do remove support, we can probably just remove the `helmVersion` field all together since there wouldn't be a use for it at that point.

To @paigecalvert's question:

> I wonder if that means that if I put `useHelmInstall: true` and `helmVersion: v2` that my release won't work?

I don't think there is anything built into KOTS that would stop a user from doing this, but it's likely to result in issues.  With this combination, KOTS would be using Helm v2 to render out the chart manifests, but then would use Helm v3 to deploy them to the cluster.  If it was an older chart with an api version of v1, then it might fail to deploy as expected.  If it was a newer chart with an api version of v2, then it may fail during the rendering phase.  In general, it's not recommended and not something we would support given the mismatched versions of Helm.

docs/reference/custom-resource-helmchart.md
```diff
@@ -214,107 +212,16 @@ The following table describes how `values` and `optionalValues` are merged based
...
-As shown above, `postgresql` is included only when the user selects the `embedded_postgres` option.
-
-To ensure the vendor portal includes this conditional `postgresql` resource in `.airgap` bundles, add the same `postgresql` value to the `builder` key with `enabled` set to `true`:
-
-```yaml
-  builder:
-    postgresql:
-      enabled: true
-```
+For more information about using the `builder` key, see [Air Gap](/vendor/helm-overview#air-gap) in _About Deploying Helm Charts_.
```
cbodonnell, line 227: This line is almost the same as the last sentence on line 223 above.  Not sure if it's intentional, but wanted to call it out just in case.  Maybe it makes more sense with the partial in between, but I'm just reading on GitHub.

docs/reference/custom-resource-helmchart.md
jonquil2002: Fixed.

docs/reference/custom-resource-helmchart.md
jonquil2002: Ok, I reverted it to include `v2` as an acceptable value, and added that `v3` is required for native Helm installations.

docs/reference/custom-resource-helmchart.md
```diff
@@ -82,43 +79,39 @@ The `chart` key allows for a mapping between the data in this definition and the
...
 <VersionLimitation/>
 :::
 
 ## useHelmInstall
 
-Identifies whether this Helm chart will use the Replicated Helm installation (`false`) or native Helm installation (`true`).
-Default is `false`.
-It is recommended that you set this to `true`, because native Helm is preferred and Replicated Helm will be removed at some point.
+Identifies whether this Helm chart will use the Replicated Helm installation (`false`) or native Helm installation (`true`). Replicated recommends that you set this to `true` because native Helm is the preferred method. **Default:** `false`
+
```
paigecalvert, line 109: Not a big deal, but noticing that there are only a couple fields that use the **Default** and **Supported values** formatting. (It could just be that there are not many fields that have defaults or supported values to document, but just wanted to raise it)

paigecalvert (approved):

PR #1089, opened 2023-04-20
installing-general-requirements.mdx - kURL Firewall Openings

CpuID:
I just did a test kURL installation, and found we're pulling the packages down from `s3.kurl.sh` also. We should reference this on the list of domains for people adding firewall/proxy exceptions etc.

picked up as part of https://github.com/replicated-collab/bigid-kots/issues/149#issuecomment-1515714077


jonquil2002:
@CpuID I think you might also need to update this section of the kurl.sh docs, but I'm not sure: https://kurl.sh/docs/install-with-kurl/system-requirements#firewall-openings-for-online-installations

jonquil2002 (approved):
lgtm - I think you might also need to update this section of the kurl.sh docs, but I'm not sure: https://kurl.sh/docs/install-with-kurl/system-requirements#firewall-openings-for-online-installations

CpuID:
thanks :)

CpuID:
thanks :) 

> lgtm - I think you might also need to update this section of the kurl.sh docs, but I'm not sure: https://kurl.sh/docs/install-with-kurl/system-requirements#firewall-openings-for-online-installations

https://github.com/replicatedhq/kurl.sh/pull/962 :) thx, good catch

PR #1094, opened 2023-04-21
item validation docs

pavansokkenagaraj:
https://app.shortcut.com/replicated/story/75137/kots-config-validation-failing-when-the-item-is-required-and-has-default-value-and-or-when-hidden


paigecalvert (approved):
thanks @pavansokkenagaraj

PR #1102, opened 2023-04-24
74588-NEW: Improve SemVer Content/Findability

jonquil2002:
SC: https://app.shortcut.com/replicated/story/74588/improve-find-ability-of-docs-on-how-to-enable-semver

Previews: 

- https://deploy-preview-1102--replicated-docs-upgrade.netlify.app/vendor/releases-about#semantic-versioning
- https://deploy-preview-1102--replicated-docs-upgrade.netlify.app/vendor/releases-about-channels#about-the-channels-page
- https://deploy-preview-1102--replicated-docs-upgrade.netlify.app/vendor/releases-about-channels#about-the-channels-page
- https://deploy-preview-1102--replicated-docs-upgrade.netlify.app/reference/replicated-cli-release-create
- https://deploy-preview-1102--replicated-docs-upgrade.netlify.app/reference/replicated-cli-release-promote


docs/vendor/releases-about.md
```diff
@@ -28,10 +28,17 @@ Semantic versioning is available with the Replicated app manager v1.58.0 and lat
 
 - For existing applications created before February 23, 2022, semantic versioning is disabled by default on all channels.
 
-When you enable semantic versioning on a channel, the version label for a release promoted to that channel is verified to ensure that it is a valid semantic version. For more information about valid semantic versions, see [Semantic Versioning 2.0.0](https://semver.org).
+To use semantic versioning:
```
jonquil2002, line 31: @paigecalvert Curious to see if you think this addition is now superfluous because I made changes to the paragraph that follows this numbered list. The paragraph might suffice, but I'm not entirely sure.

paigecalvert: Good question...I think it is good info to include. There might be a couple other pieces of info that would be good to have in this section as well, given that it's now the "hub" for all/most the semver info:

- What does it mean to "use" semantic versioning? This could be important just to differentiate from the fact that anyone can use whatever version label they want of course, but the real _feature_ here is that you can enforce it
- And then one step further: why would a vendor want to enable semver on a channel? (I think it's because semver is recommended because it makes releases more predictable for users and it's useful to have it enforced so that no one accidentally uses an incorrect version)
- Finally, in addition to your explanation about how you use semver labels when you promote releases after you enable semver on a channel, we could also write a sentence to explain how to enable semver: "To enable semver on a channel, click the gear icon and enable the [Toggle Name] toggle." We talk about enabling semver in this section, though we don't just say how. It could be useful to include this instruction so that people don't have to search for how to enable it, as it might not be obvious that it's in the channel settings.

paigecalvert (dismissed):
Looks good! My main comment was just a couple more suggestions for pieces of info that we could include in the semver section

PR #1106, opened 2023-04-25
Fix important admonitions

paigecalvert:
https://app.shortcut.com/replicated/story/75359/bug-fix-admonitions


paigecalvert:
Both showing good on the preview, so going to merge

PR #1112, opened 2023-04-28
clean up. remove commented out config lines that are no longer needed

St0rmz1:
remove commented out config lines that are no longer needed


paigecalvert (approved):
thank you!

PR #1127, opened 2023-05-05
76117: Move Annotations, Kubernetes Jobs, and Kubernetes Objects in TOC

jonquil2002:
SC: https://app.shortcut.com/replicated/story/76117/move-annotations-and-cleaning-up-kubernetes-jobs-higher-up-in-toc

Moved Kubernetes Operators to higher level near bottom. Also created a new TOC category called "Managing Objects" and moved the Annotations and Cleaning Up Kubernetes Jobs under the new category.

Preview: https://deploy-preview-1127--replicated-docs-upgrade.netlify.app/

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -206,7 +206,7 @@ const sidebars = {
     },
     {
       type: 'category',
-      label: 'Managing Objects',
+      label: 'Managing Resources and Objects',
       items: [
         'vendor/packaging-include-resources',
         'vendor/packaging-cleaning-up-jobs',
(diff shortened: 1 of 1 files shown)


paigecalvert (approved):
Nice, I like it!

Only question is around the term _Objects_ in this context. As in, I wonder if including/excluding optional _resources_ would fall under the category of managing _objects_. Maybe it's a subtle distinction, or really no distinction at all, but could be a good q for Dex

paigecalvert (dismissed):
Nice, I like it!

Only question is around the term _Objects_ in this context. As in, I wonder if including/excluding optional _resources_ would fall under the category of managing _objects_. Maybe it's a subtle distinction, or really no distinction at all, but could be a good q for Dex

PR #1136, opened 2023-05-10
Add clarification between HelmChart and Helm chart files

paigecalvert:
Preview: https://deploy-preview-1136--replicated-docs-upgrade.netlify.app/vendor/helm-install#add-the-admin-console-chart

Adds a comment at the top of each code block to clarify which file it is
Adds header content to the top of each code block to clarify which file it is 
Cleans up the examples to make them consistent and reduce the overall length of the steps (each step has only one code block now).

### DIFF
--- a/docs/vendor/helm-install.md
+++ b/docs/vendor/helm-install.md
@@ -96,6 +96,10 @@ To conditionally include and exclude the `admin-console` Helm chart:
 1.  In the Helm chart `Chart.yaml` file, add the `admin-console` Helm chart to the `dependencies` field:
 
    ```yaml
+    # Helm chart Chart.yaml file
+
+    apiVersion: v2
+    ...
     dependencies:
     - name: admin-console
       version: "1.83.0"
@@ -131,8 +135,10 @@ To conditionally include and exclude the `admin-console` Helm chart:
 1. In the `values.yaml` file, add the following fields that map to the `condition` field in the `Chart.yaml` file:
 
    ```yaml
+   # Helm chart values.yaml file
+
    admin-console:
-    enabled: true
+     enabled: true
    ```
 
    Helm Install installations read this `true` value from the `values.yaml` file and include the `admin-console` Helm chart in the deployment.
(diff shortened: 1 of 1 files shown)


ousbots (approved):

docs/vendor/helm-install.md
```diff
@@ -131,101 +137,117 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
+        ...
         values:
           admin-console:
-           enabled: false  
-      ```     
+            enabled: false
+      ```
 
-   1. If the HelmChart custom resource includes an `optionalValues` field, set `recursiveMerge` to `true`. This prevents the `admin-console` field in `values` from being overwritten by the fields in `optionalValues`.
+      KOTS Install and Embedded Cluster installations read this static value of `false` from the HelmChart custom resource, which prevents the `admin-console` Helm chart from deploying. Helm Install installations ignore this static value, and instead read the value of `true` from the `values.yaml` file that you added in a previous step.
```
jonquil2002, line 186: Re: KOTS Install and Embedded Cluster installations, we've never capitalized those previously. And we generally don't use "install" unless it's a verb, we use "installation" instead, the exception being Helm Install. Not sure how I feel about the capitalization, and "Helm Install installations" feels uncomfortable. 

Also, are we moving to embedded cluster officially? I think we need to make some decisions, but that could take awhile with PM. In the meantime, how do you feel about sticking with the current naming we've been using and then maybe it'll be easier to do search/replace if the terminology changes?

docs/vendor/helm-install.md
```diff
@@ -131,101 +137,117 @@ To conditionally include and exclude the `admin-console` Helm chart:
...
           admin-console:
-           enabled: false              
+            enabled: false
       ```
+
+      Values in the `builder` field provide a way to render the chart with all images and manifests. A common use case for the `builder` field in Replicated is to create packages for delivering an application to an air gap environment.
+
+      For more information, see [builder](/reference/custom-resource-helmchart#builder) in _HelmChart_.
+
+  **Example:**
```
jonquil2002, line 228: I think this example should be indented instead of inline with the numbered steps.

jonquil2002 (approved):
lgtm. The additions of examples is a great idea. A few questions and minor nits.

docs/vendor/helm-install.md
paigecalvert: Yeah you're absolutely right. Those terms come directly from the vendor portal, which felt like the cleanest way to keep everything straight since there are three different options. (All of this is defined in the intro paragraphs in this topic).

It's definitely worth a revisit, though I expect this topic to change heavily for the foundation plan launch, so it might not be worth it in the short term. When the SDK launches and we rewrite this topic or delete it, maybe editing the terminology at that point would make sense?

PR #1141, opened 2023-05-15
HOLD FOR RELEASE 76191: Add HelmChart v2 Reference Topic

jonquil2002:
SC: https://app.shortcut.com/replicated/story/76191/add-v2-helmchart-reference-topic

Previews:

- https://deploy-preview-1141--replicated-docs.netlify.app/reference/custom-resource-helmchart-v2
- https://deploy-preview-1141--replicated-docs.netlify.app/reference/custom-resource-helmchart


docs/reference/custom-resource-helmchart-v2.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+## namespace
+
+<Namespace/>
+
+## builder
+
+The `builder` field is required for the following use cases:
+
+* To support online installations that are configured to push and pull from a private registry, where `disable image push` is _not_ selected. You must provide the necessary values in the `builder` field to render the helm chart with all of the necessary images.
```
jonquil2002, line 135: @ajp-io @cbodonnell I believe this first applies only to v2, yes?

docs/reference/custom-resource-helmchart-v2.mdx
```diff
@@ -0,0 +1,160 @@
...
+`postgresql` is defined as a conditional resource in the `values` key of the HelmChart custom resource:
+
+```yaml
+  values:
+    postgresql:
+      enabled: repl{{ ConfigOptionEquals `postgres_type` `embedded_postgres`}}
+```
+As shown above, `postgresql` is included only when the user selects the `embedded_postgres` option.
+
+To ensure the vendor portal includes this conditional `postgresql` resource in the installation file, add the same `postgresql` value to the `builder` key with `enabled` set to `true`:
```
jonquil2002, line 154: @ajp-io @cbodonnell In this v2 sentence, I wrote "installation file". In the v1 version, it says "`airgap bundle`. Is it ok to say `installation file` in the v1 version instead of `airgap bundle`? If so, I can share the same content in a partial instead of maintaining two versions of almost identical content.

docs/reference/custom-resource-helmchart-v2.mdx
cbodonnell: Yup, that's only for v2 and for that particular use-case described.

Related note - If a vendor currently supports airgap installs, they will already have filled out this field, so there would be no additional changes needed on their end.  Additional effort would only be needed for vendors that do not support airgap installs, but do want to support the use-case described here.  They would need to fill out the `builder` field if they haven't already.

docs/reference/custom-resource-helmchart-v2.mdx
cbodonnell: Interested to hear what @ajp-io thinks here.  In v2, the `builder` field is used by the Vendor Portal (same as v1), but it's also used by the app manager.  Both use those values to template out the Helm chart and find all necessary images.  Maybe we can make the wording more generic so that it's not necessarily referencing the vendor portal or app manager, but gets the point across that `builder` is used to make sure that all conditional resources are included in the installation?

docs/reference/custom-resource-helmchart-v2.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+### chart.name
+
+<ChartName/>
+
+### chart.chartVersion
+
+<ChartVersion/>
+
+### chart.releaseName
```
cbodonnell, line 95: `releaseName` is a top-level property of the spec in v2 instead of being nested within `chart` because it's not really part of the "mapping between the data in this definition and the chart archive itself" as we state in the `chart` description.

The example yaml spec above looks correct.

```suggestion
## releaseName
```

docs/partials/helm/_helm-cr-weight.mdx
```diff
@@ -0,0 +1,3 @@
+Determines the order to apply charts that have `useHelmInstall: true`. Charts are applied by weight in ascending order, with lower weights applied first. **Supported values:** Positive or negative integers. **Default:** `0`
```
cbodonnell, line 1: There's a reference to `useHelmInstall: true` here, which wouldn't apply in v2.  Could we re-word it to just reference "native helm" somehow?

docs/partials/helm/_helm-cr-upgrade-flags.mdx
```diff
@@ -0,0 +1,16 @@
+:::note
+`helmUpgradeFlags` is available in the app manager v1.75.0 and later.
+:::
+
+Specifies additional flags to pass to the `helm upgrade` command for charts that have `useHelmInstall: true`. These flags are passed in addition to any flags the app manager passes by default. The values specified here take precedence if the app manager already passes the same flag. The `helmUpgradeFlags` attribute can be parsed by template functions. For more information about template functions, see [About template function contexts](template-functions-about).
```
cbodonnell, line 5: There's a reference to `useHelmInstall: true` here, which wouldn't apply in v2.  Could we re-word it to just reference "native helm" somehow?

docs/reference/custom-resource-helmchart-v2.mdx
jonquil2002: I can put some language in about that, I meant to earlier.

docs/reference/custom-resource-helmchart-v2.mdx
jonquil2002: Oof, I didn't even see that. I'll just have to make this two different field descriptions instead of sharing a partial.

docs/partials/helm/_helm-cr-weight.mdx
jonquil2002: Oof, I didn't even see that. I'll just have to make this two different field descriptions instead of sharing a partial.

jonquil2002: Yep, did it in the partial for both v1 & v2.

docs/partials/helm/_helm-cr-upgrade-flags.mdx
jonquil2002: Yep, did it in the partial for v1 & v2

docs/reference/custom-resource-helmchart-v2.mdx
jonquil2002: Made the changes. PTAL

docs/reference/custom-resource-helmchart-v2.mdx
jonquil2002: Made this change. PTAL

docs/partials/helm/_helm-cr-chart-release-name.mdx
```diff
@@ -0,0 +1,8 @@
+> Introduced in Replicated app manager v1.73.0
```
jonquil2002, line 1: @ajp-io  Do we need a release number here for v2? Can we delete this language altogether for both v1 & v2?

docs/partials/helm/_helm-cr-chart-release-name.mdx
```diff
@@ -0,0 +1,8 @@
...
+
+Specifies the release name to use when installing this instance of the Helm chart.
+Defaults to the chart name.
```
jonquil2002, line 4: @cbodonnell Seems like this line should be deleted for v2 - yes? Does any other language in this field description need to change for v2?

docs/partials/helm/_helm-cr-chart-release-name.mdx
ajp-io: I left a similar comment lower down for the v2 resource. We shouldn't mention which version fields were introduced for v2, since they were all released now (whatever our next KOTS version is). It could still be helpful for v1, though those versions are getting older and hopefully people are using sufficiently recent versions. That might be a docs decision.

But for v2, we should just say when the v2 resource was added to KOTS, not individual fields.

docs/reference/custom-resource-helmchart-v2.mdx
ajp-io: I think we should make it clearer why this is needed. Just a suggestion, doesn't have to be this wording.
```suggestion
* To support online installations that are configured to push and pull from a private registry, where `disable image push` is _not_ selected. You must provide the necessary values in the `builder` field to render the Helm chart with all of the necessary images so that KOTS knows where to pull the images from to push them into the private registry.
```

docs/partials/helm/_helm-cr-chart-release-name.mdx
ajp-io: I might be misunderstanding the question, but I think this statement is still true. It's just that `releaseName` moved from a subfield of `chart` in v1 to a top-level field in v2. But I don't know if I'm missing something.

ajp-io (changes requested):

docs/partials/helm/_helm-cr-chart-release-name.mdx
cbodonnell: > I might be misunderstanding the question, but I think this statement is still true. It's just that `releaseName` moved from a subfield of `chart` in v1 to a top-level field in v2. But I don't know if I'm missing something.

Right, we just moved this field.  The behavior is the same, so I think we can keep the same language too.

docs/reference/custom-resource-helmchart-v2.mdx
```diff
@@ -0,0 +1,152 @@
...
+import OptionalValuesWhen from "../partials/helm/_helm-cr-optional-values-when.mdx"
+import OptionalValuesRecursiveMerge from "../partials/helm/_helm-cr-optional-values-recursive-merge.mdx"
+import Namespace from "../partials/helm/_helm-cr-namespace.mdx"
+import Intro from "../partials/helm/_helm-cr-intro.mdx"
+import BuilderAirgapIntro from "../partials/helm/_helm-cr-builder-airgap-intro.mdx"
+import BuilderExample from "../partials/helm/_helm-cr-builder-example.mdx"
+
+# HelmChart v2
+
+The HelmChart v2 custom resource supports the native Helm v2 installation. This v2 installation method does a Helm install or upgrade on your Helm chart. Because this method bypasses existing Kustomize logic to rewrite images, inject pull secrets, and add backup labels and annotations, you must configure a combination of Replicated and Helm templating for the images, pull secrets, and backup labels and annotations.
```
cbodonnell, line 21: I did some digging late last week and it looks like the annotations are an old thing we have always added in KOTS and are used to identify resources that are managed by KOTS.  However, in this install method, all resources would be managed by Helm, so I'm fairly sure vendors would not need to add annotations.
```suggestion
The HelmChart v2 custom resource supports the native Helm v2 installation. This v2 installation method does a Helm install or upgrade on your Helm chart. Because this method bypasses existing Kustomize logic to rewrite images, inject pull secrets, and add backup labels, you must configure a combination of Replicated and Helm templating for the images, pull secrets, and backup labels.
```

ajp-io (changes requested):
TBH for the sake of time I just read the parts that I thought were new here—changes to `builder` the intro about v2 installations, etc. I figure the rest of the fields are copy/paste from v1, but let me know if I missed something.

docs/reference/custom-resource-helmchart-v2.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+<Namespace/>
+
+## builder
+
+The `builder` field is required for the following use cases:
+
+* <BuilderAirgapIntro/>
+
+* To support online installations that are configured to push and pull from a private registry, where `disable image push` is _not_ selected. You must provide the necessary values in the `builder` field to render the Helm chart with all of the necessary images so that KOTS knows where to pull the images from to push them into the private registry.
```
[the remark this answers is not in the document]
jonquil2002: Lol, I thought you were just trying to make that name change happen! I fully supported :)

docs/reference/custom-resource-helmchart-v2.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+<Namespace/>
+
+## builder
+
+The `builder` field is required for the following use cases:
+
+* <BuilderAirgapIntro/>
+
+* To support online installations that are configured to push and pull from a private registry, where `disable image push` is _not_ selected. You must provide the necessary values in the `builder` field to render the Helm chart with all of the necessary images so that the app manager knows where to pull the images from to push them into the private registry.
```
paigecalvert, line 137: I reck'n `disable image push` might be referring to the `--disable-image-push` flag when the user runs the kots install command? Or is it referring to a field somewhere?I think this could be helpful to clarify. When I read it the first time, I thought "when it is not selected _where_?"

paigecalvert (dismissed):
one comment!

ajp-io (dismissed):
I made a couple small changes. This is good to me if we add Beta and include the KOTS version this will be released in.

docs/reference/custom-resource-helmchart-v2.mdx
jonquil2002: Good call.

@cbodonnell - Pls check and see if the wording for this bullet is now correct? I added the proper flag syntax and added the reference to `kots install`.

If it's correct, please approve the PR. Thank you! :)

cbodonnell: That is accurate, but it can also be configured in the UI.  My bad for the incorrect wording when I added that to the doc (was going off memory).  Here's what it looks like in the UI:
Screen Shot 2023-05-17 at 11 48 00 AM

So maybe something along the lines of "where the `--disable-image-push` flag is _not_ passed to the `kots install` command or 'Disable Pushing Images to Registry' is _not_ selected"?

jonquil2002: @cbodonnell Ah, thanks. Made some changes. What do you think now?

cbodonnell (dismissed):

docs/reference/custom-resource-helmchart-v2.mdx
cbodonnell: Looks good to me!

docs/reference/custom-resource-helmchart-v2.mdx
```diff
@@ -0,0 +1,149 @@
...
+import OptionalValuesWhen from "../partials/helm/_helm-cr-optional-values-when.mdx"
+import OptionalValuesRecursiveMerge from "../partials/helm/_helm-cr-optional-values-recursive-merge.mdx"
+import Namespace from "../partials/helm/_helm-cr-namespace.mdx"
+import Intro from "../partials/helm/_helm-cr-intro.mdx"
+import BuilderAirgapIntro from "../partials/helm/_helm-cr-builder-airgap-intro.mdx"
+import BuilderExample from "../partials/helm/_helm-cr-builder-example.mdx"
+
+# HelmChart v2 (Beta)
+
+> Introduced in Replicated app manager v1.xx.xx
```
cbodonnell, line 21: @jonquil2002 this will be v1.99.0

jonquil2002: Thank you! :)

cbodonnell (approved):

paigecalvert (approved):

PR #1154, opened 2023-05-18
update velero version compatibility matrix

cbodonnell:
updates the velero version compatibility matrix to indicate compatibility through version 1.10.2 because of this known upstream issue: https://docs.replicated.com/enterprise/snapshots-troubleshooting-backup-restore#partial-snapshot-restore-finishes-with-warnings

### DIFF
--- a/docs/enterprise/snapshots-understanding.md
+++ b/docs/enterprise/snapshots-understanding.md
@@ -82,4 +82,4 @@ The app manager is based on the open source KOTS project, which is maintained by
 |------|-------------|
 | 1.15 to 1.20.2 | 1.2.0 |
 | 1.20.3 and later | 1.5.1 through 1.9.x |
-| 1.94.1 and later | 1.6.x and later |
+| 1.94.1 and later | 1.6.x through 1.10.2 |


docs/enterprise/snapshots-understanding.md
```diff
@@ -82,4 +82,4 @@ The app manager is based on the open source KOTS project, which is maintained by
 |------|-------------|
 | 1.15 to 1.20.2 | 1.2.0 |
 | 1.20.3 and later | 1.5.1 through 1.9.x |
-| 1.94.1 and later | 1.6.x and later |
\ No newline at end of file
+| 1.94.1 and later | 1.6.x through 1.10.2 |
```
ajp-io, line 85: Should this be 1.10.x in case there are additional patch releases for 1.10? Or is 1.10.2 definitely the last minor for that major?

I assume this bug wouldn't get included in subsequent 1.10 patch releases if they're created…

pavansokkenagaraj: `1.10.3` has the bug too. So it needs to be `1.10.2`

cbodonnell: @ajp-io, unfortunately the bug also exists in 1.10.3 in addition to 1.11.0.  1.10.2 is the latest version that does not produce these warnings.

ajp-io: Ok, I'll approve then.

ajp-io (approved):

paigecalvert (approved):

PR #1156, opened 2023-05-18
77302: Created partials for generating support bundles content

jonquil2002:
SC: https://app.shortcut.com/replicated/story/77302/make-generating-support-bundles-cli-into-partials

Previews: https://deploy-preview-1156--replicated-docs.netlify.app/enterprise/troubleshooting-an-app#generating-a-support-bundle-using-the-cli


docs/enterprise/troubleshooting-an-app.mdx
```diff
@@ -0,0 +1,44 @@
...
+import GenerateBundleCliAdminConsole from "../partials/support-bundles/_generate-bundle-cli-admin-console.mdx"
+import GenerateBundleHost from "../partials/support-bundles/_generate-bundle-host.mdx"
+import GenerateBundleDefaultKots from "../partials/support-bundles/_generate-bundle-default-kots.mdx"
+
+
+# Generating Support Bundles
+
+The Replicated admin console includes a Troubleshoot page where you can generate an analysis and review remediation suggestions for troubleshooting an application. You can download a support bundle to share with your vendor.
+
+Alternatively, you can generate a support bundle using the support-bundle CLI. For Kubernetes installer clusters, you can also generate a host support bundle.
```
jonquil2002, line 13: I changed the first sentence here just a little because they can do more than generate a bundle on running clusters.

docs/enterprise/troubleshooting-an-app.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+## Generating a Bundle using the CLI
+
+You can generate a support bundle using the command line (CLI) instead of the admin console. For example, the admin console might not be available if you are using a Kubernetes installer (embedded installer) and the installation fails. Or perhaps you need to use a more recent version of the support-bundle plugin than what is embedded in the admin console.
+
+### Prerequisite: Install the Plugin {#plugin}
+
+<InstallPlugin/>
+
+### Generate a Merged Support Bundle
```
jonquil2002, line 27: @paigecalvert I went back and forth on what the headings should be in this section. The capabilities in this section are more extensive and more powerful than were originally documented, and I think this is why we struggled back then to get that across. Hopefully this is a better clarification, but wonder what you think.

jonquil2002: I also moved this entire section higher in the list of CLI procedures, and of course we'll handle it differently for the new TOC/vendor side of things. But SWYT.

docs/partials/support-bundles/_generate-bundle-admin-console.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+  The analysis executes the support bundle plugin. After the analysis completes, the bundle is available on the Troubleshoot tab in the admin console. If any known issues are detected, they are highlighted with possible remediation suggestions.
+
+  :::note
+  No data leaves the cluster. Data is never sent across the internet or to anyone else.
+  :::
+
+  ![Analysis](/images/analysis.png)
+
+1. (Optional) Click **Download bundle** to download the support bundle. You can send the bundle to your vendor for assistance.
```
paigecalvert, line 17: Noting that when we want to add these partials to the vendor docs we'll need to find language that works for both.

Maybe part of a separate iteration?

paigecalvert (commented):
Looks great @jonquil2002 , some suggestions and thoughts in the comments!

docs/partials/support-bundles/_generate-bundle-admin-console.mdx
jonquil2002: Yes, a separate iteration it what I had in mind. I was thinking that maybe the intro sentence for some of it might be specific for a vendor or a customer and sit outside of the partial. Specifically thinking about where we explain to the customer that the vendor will tell them about any non-cluster specs, like URLs.

docs/enterprise/troubleshooting-an-app.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+Alternatively, you can generate a support bundle using the support-bundle CLI. For Kubernetes installer clusters, you can also generate a host support bundle.
+
+## Generate a Bundle using the Admin Console
+
+<GenerateBundleAdminConsole/>
+
+## Generating a Bundle using the CLI
+
+You can generate a support bundle using the command line (CLI) instead of the admin console. For example, the admin console might not be available if you are using a Kubernetes installer (embedded installer) and the installation fails. Or perhaps you need to use a more recent version of the support-bundle plugin than what is embedded in the admin console.
```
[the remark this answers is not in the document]
jonquil2002: Re: updating the support-bundle plug-in, that was recommended by Ada, as I recall. I didn't get a use case, but I'll ask. I also reworded this whole section and on the Troubleshoot site, it looks like the procedures have changed for installing krew, so I'll get a technical review.

docs/partials/support-bundles/_install-plugin.mdx
```diff
@@ -0,0 +1,42 @@
+For existing clusters, do a first-time installation or update the support-bundle plugin (a kubectl plugin) before you generate a support bundle using the CLI. The plugin is required to generate a support bundle.
```
jonquil2002, line 1: @xavpaice I updated this section bc it looks like the krew install instructions have changed on the Troubleshoot site (originally we were told to use a curl command).

Also, what use case would require that they update the support-bundle plugin if it's already installed? Would that be a case where they updated KOTS at some point on an existing cluster but never updated the support-bundle plug-in? How would a customer know which version of the plugin to update to?

docs/partials/support-bundles/_generate-bundle.mdx
```diff
@@ -0,0 +1,36 @@
+Support bundles are generated from one or more support bundle specifications. Specification locations include YAML manifest files in the cluster, URLs, stdin, and Secrets or ConfigMap objects embedded in a Kubernetes cluster. For more information, see [Collecting a Support Bundle](https://troubleshoot.sh/docs/support-bundle/collecting/) and [Discover Cluster Specs](https://troubleshoot.sh/docs/support-bundle/discover-cluster-specs/) in the Troubleshoot documentation.
+
+The following examples show some possible CLI commands to generate a support bundle for one or more specifications and locations:
+
+- Automatically discover specifications in the cluster that are contained in Secrets or ConfigMaps:
```
jonquil2002, line 5: @xavpaice This content was added awhile ago, but it was at the bottom of the TOC, so you didn't see the --load-cluster-specs info. I've reordered the procedure in the TOC to make it prevalent, and I've reformatted the content so that it hopefully makes mroe sense. PTAL and SWYT. Thanks!

docs/partials/support-bundles/_install-plugin.mdx
xavpaice: you don't need to use Krew, but it does make it easier.  I find just running `kubectl krew upgrade` does wonders for enabling new features.  I can't tie that to KOTS though, I would suggest just always using the most recent.

For install instructions, you have options.  Krew is easy, but you need to install krew first.  You can also curl the release directly from Github (I suspect that's the one that got removed?).

docs/partials/support-bundles/_generate-bundle.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+  ```bash
+  kubectl support-bundle --load-cluster-specs
+  ```
+- Automatically discover specifications found in the cluster that match a custom label:
+
+  ```bash
+  kubectl support-bundle --load-cluster-specs -l troubleshoot.io/kind=LABEL
```
xavpaice, line 13:
```suggestion
  kubectl support-bundle --load-cluster-specs -l LABEL
```

docs/partials/support-bundles/_install-plugin.mdx
jonquil2002: @xavpaice  So if I understand this correctly, we should recommend always upgrading to the latest support-bundle plugin even if they're running KOTS? This will give them the latest Troubleshoot features if they're supported by their current version of KOTS? And if the Troubleshoot features are not supported by their current version of KOTS, it won't cause their support bundle to fail?

jonquil2002: For the record, Xav helped me sort this in Slack. https://replicated.slack.com/archives/C02BPJQQP28/p1685460186426479

We are good to go.

PR #1159, opened 2023-05-19
77116: Remove App Layer Identity Service

jonquil2002:
SC: https://app.shortcut.com/replicated/story/77116/remove-the-app-level-authentication-information-from-the-beta-identity-service-related-docs

Previews:

- This topic needed improvements to help call out the admin console identity option: https://deploy-preview-1159--replicated-docs-upgrade.netlify.app/vendor/releases-creating-customer
- This topic needed a little more context, and since it talks about the identity enablement, it seemed apropos to add a little something - and also fixed some links that weren't pointing to the right thing: https://deploy-preview-1159--replicated-docs-upgrade.netlify.app/vendor/licenses-using-builtin-fields

Also removed the identity-service-configuring.md topic after confirming with Salah that it was app layer only. I set up a redirect for the deleted topic.


docs/vendor/licenses-using-builtin-fields.md
```diff
@@ -7,7 +7,7 @@ custom license fields for a customer, see [Managing Custom License Fields](licen
 
 All applications that are packaged and delivered through the Replicated vendor
 portal include built-in license fields. Built-in fields are reserved field names
-and are available in every customer license.
+and are available in every customer license. These license fields can be configured using the Vendor API or the Identity Context template functions. For more information, see [Using the Vendor API v3](/reference/vendor-api-using) and [Identity Context](/reference/template-functions-identity-context).
```
sgalsaleh, line 10: so, not all these license fields are configurable. the ones that are, can be configured via the API or the vendor portal, not the template functions. they can be _used_ via template functions.

also, it would be the license context template function, not the identity context one.

docs/vendor/licenses-using-builtin-fields.md
```diff
@@ -22,9 +22,9 @@ customer.
...
 | `isAirgapSupported` | If a license supports air gapped installations, then this field is set to `true`. If air gapped is not supported, this field is missing. When you enable this feature, the `.rli` file will have license meta data embedded in it, and must be re-downloaded. |
-| `isGeoaxisSupported` | If a license supports integration with GeoAxis, this field is set to `true`. If GeoAxis is not supported, this field is either `false` or missing. **Note**: This field requires that the vendor has the GeoAxis entitlement. It also requires that the vendor has access to the Identity Service entitlement. See [Enabling and Configuring Identity Service (Beta)](identity-service-configuring). |
+| `isGeoaxisSupported` | If a license supports integration with GeoAxis, this field is set to `true`. If GeoAxis is not supported, this field is either `false` or missing. **Note**: This field requires that the vendor has the GeoAxis entitlement. It also requires that the vendor has access to the Identity Service entitlement. For information about the identity service option, see [Creating a Customer)](releases-creating-customer). |
```
sgalsaleh, line 25: is creating a customer the best resource we can point to for more information around configuring the identity service?

docs/vendor/releases-creating-customer.md
```diff
@@ -15,18 +11,44 @@ To create a customer:
...
+
+    The available license options are:
+
+    <table>
+        <tr>
+          <th width="30%">Name</th>
+          <th width="70%">Description</th>
+        </tr>
+        <tr>
+          <td>Air gap</td>
```
sgalsaleh, line 29: this is how it's shown in the vendor portal
```suggestion
          <td>Airgap Download Enabled</td>
```

docs/vendor/releases-creating-customer.md
```diff
@@ -1,12 +1,8 @@
 # Creating a Customer
 
-After you promote a release to a channel, you create a customer.
+After you promote your first release to a channel, you create a customer. Additional customers can be added as needed.
```
paigecalvert, line 3: Good call editing this language...perhaps we could just delete these sentences all together? (I could also see a "This topic describes how to create a customer" type sentence)

netlify.toml
```diff
@@ -100,6 +100,10 @@
 [[redirects]]
   from="https://docs.replicated.com/vendor/team-management-rbac-about"
   to="https://docs.replicated.com/vendor/team-management-rbac-configuring"
+
+[[redirects]]
+  from="https://docs.replicated.com/vendor/identity-service-configuring"
+  to="https://docs.replicated.com/vendor/releases-creating-customer"
```
paigecalvert, line 106: I know we haven't done this before, but I am wondering if it might be better to let this topic 404, as the content is totally going away? Alternatively, we could remove it from the TOC and then add a warning at the top of the page to let people know why it's being removed.

As is, the Creating a Customer topic doesn't contain any of the info from the topic that is being removed (which is intentional), so it could be disorienting to land there if you are expecting to find identity service info. I have been thinking about this as it pertains to the helm CLI install docs being removed, and so it was top of mind: do we want to redirect someone at all if the content they are looking for has been entirely and intentionally removed?

docs/vendor/releases-creating-customer.md
```diff
@@ -15,18 +11,44 @@ To create a customer:
...
+          <td>Enables new installations with this license to install from an air gap package or from an online installation.</td>
+        </tr>
+        <tr>
+          <td>Gitops Enabled</td>
+          <td>Enables existing cluster and embedded cluster installations to transition to a GitOps workflow for deployment.</td>
+        </tr>
+        <tr>
+          <td>Identity Service Enabled</td>
+          <td>Enables installations to integrate with third-party identity service providers to provision RBAC for customer authenticating from the admin console.</td>
+        </tr>
```
paigecalvert, line 39: customer authenticating > customer**s** authenticating ?

or

customer authenticating > customer authentication ?

paigecalvert: "from the admin console" > does this mean that there would be RBAC rules applied upon logging in to the admin console? Or that the user would be able to provision RBAC for authentication to the app from within the admin console?

paigecalvert (dismissed):
looks good @jonquil2002 , just some questions

docs/vendor/releases-creating-customer.md
jonquil2002: @ salah Can you confirm - Does this mean that the customer would be able to provision RBAC for authentication to the app from within the admin console using these two procedures:

- https://docs.replicated.com/enterprise/auth-identity-provider
- https://docs.replicated.com/enterprise/auth-configuring-rbac

Also, we use the language "Enables installations to integrate with third-paty identity providers" but I think it would be clearer to say "Enables customers to..."

docs/vendor/releases-creating-customer.md
jonquil2002: Good call.

netlify.toml
jonquil2002: Interesting considerations @paigecalvert 

Maybe we ought to explain that we're no longer supporting this but we do support identity service for the admin console authentication instead, and then remove the topic altogether once the Release Notes say that the Beta feature is removed from the code.

Without the Release Notes announcement, the 404 error might seem like just a broken link. I guess let's try the former?

paigecalvert: Nice, I like that plan (removing the topic once there's the release note that says that the feature is removed). I think the main concern was just removing things without letting the users know somehow that they've been removed. Redirects are of course very useful when we move content around, but are less useful when the content is removed all together.

docs/vendor/releases-creating-customer.md
sgalsaleh: > customer authenticating > customers authenticating ?

"customers authenticating" sounds right to me. maybe we can use "users authenticating"? i'd fine with either.

> "from the admin console" > does this mean that there would be RBAC rules applied upon logging in to the admin console? Or that the user would be able to provision RBAC for authentication to the app from within the admin console?

the latter, but it's for authentication to the admin console, not the app.

> Also, we use the language "Enables installations to integrate with third-paty identity providers" but I think it would be clearer to say "Enables customers to..."

agreed.

sgalsaleh (approved):

docs/reference/custom-resource-identity.md
```diff
@@ -1,3 +1,7 @@
+:::important
+This topic has been deleted from the product documentation because this Beta feature being deprecated.
```
paigecalvert, line 2: A couple suggestions for present tense:

>This topic has been deleted

This topic is removed/deleted?

>this Beta feature being deprecated

this Beta feature is deprecated?

paigecalvert (approved):
lgtm! there's a small typo in the important note, and also added a couple minor suggestions to that note for tense

PR #1169, opened 2023-05-24
remove k8s 1.27 support in KOTS

cbodonnell:
There is an issue with native Helm installs in KOTS on 1.27 - it will be resolved by the next patch release, but we should remove this support from docs until then


ajp-io (approved):

ajp-io:
@replicatedhq/replicated-docs

paigecalvert (approved):
LGTM

Was this a bug that customers noticed? Just wondering if we need to add a notice to tell people that we previously said it was support but it's not anymore

cbodonnell:
> LGTM
> 
> Was this a bug that customers noticed? Just wondering if we need to add a notice to tell people that we previously said it was support but it's not anymore

The incompatibility was discovered internally.  No customers have reported this issue yet.

PR #1175, opened 2023-05-25
add coming soon note to kurl OS events

dexhorthy:


paigecalvert:
Closing. We're going to remove these and add them in when they are released

PR #1205, opened 2023-06-02
TEST---77337: Builders Plan Preflights and Support Bundles

jonquil2002:
SC: https://app.shortcut.com/replicated/story/77337/foundation-plan-create-vendor-support-bundle-section

Previews: 
- 
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/support-online-support-bundle-specs
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/preflight-support-bundle-about
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/support-modular-support-bundle-specs
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/preflight-sb-helm-templates-about
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/preflight-helm-defining
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/preflight-kots-defining
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/preflight-running
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/support-bundle-helm-customizing
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/support-bundle-kots-customizing
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/support-bundle-kots-secrets-configmaps
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/support-bundle-generating
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/reference/custom-resource-preflight
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/reference/custom-resource-redactor (no changes made to content - just moved it in the TOC)
- https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/vendor/support-host-support-bundles


jonquil2002:
@xavpaice  Ready for a first pass review. This Helm content - and subsequent changes that were needed for KOTS - are in this PR. Some of the bigger changes are also copy/pasted into this Gdoc in case that's easier for you to mark up: https://docs.google.com/document/d/1K1kPHSF3dagXqZFe63P_lisCATWsQLkffsEdW0ZBFbA/edit

We can also walk through and discuss this tomorrow.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,316 @@
...
+      - "pods/log"
+    verbs:
+      - "get"
+      - "list"
+```
+
+
+### ClusterRoleBinding
+
+The following example shows RBAC settings for `ClusterRoleBinding` with preflight checks. It also shows `preflight-install` and `preflight-upgrade` hooks and weights.
```
jonquil2002, line 281: @xavpaice Can you provide more context around how and when to use RBAC with preflight checks? I'm unclear about this and think that vendors might need more guidance.

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,279 @@
...
+## Helm Specification Guidance {#helm}
+
+For Helm charts, the following additional guidance can help you decide which specification types to use.
+
+
+### Preflight Guidance
+
+The following tables can help guide you in deciding which options to use for preflight checks.
+
+#### Secrets Without Hooks
```
jonquil2002, line 103: @xavpaice Should we add tables for Preflights using ConfigMaps with and without hooks? If so, would the info be the same, except for the fact that ConfigMaps shouldn't be used to collect private info from the cluster?

docs/vendor/support-online-support-bundle-specs.md
```diff
@@ -1,17 +1,18 @@
...
 
 You can keep the online specification current with collectors and analyzers to manage bug fixes. When a customer generates a support bundle, the online specification can detect those potential problems in the cluster and let them know know how to fix it. Without the URI link option, you must wait for the next time your customers update their applications or Kubernetes versions to get notified of potential problems. The URI link option is particularly useful for customers that do not update their application routinely.
 
-If you are using a modular approach to designing support bundles, you can use multiple online specifications. Each specification supports one URI link. For more information about modular specifications, see [About Creating Modular and Discoverable Support Bundles](support-modular-support-bundle-specs).
+If you are using a modular approach to designing support bundles, you can use multiple online specifications. Each specification supports one URI link. For more information about modular specifications, see [About Modular Support Bundles](support-modular-support-bundle-specs).
 
 ## Example: URI Linking to a Source Repository
```
jonquil2002, line 11: @xavpaice It would be good to add a Helm chart example. I'd really like to replace this KOTS/KUSL example as well after July 31 because it shows Replicated "dog fooding" its own product. That is great in theory, but in communicating through technical documentation, it can be confusing to users. User can think "Are you telling me this is how I should use your product?" It would be better to give them a genericized real world example, but I would need your help with that.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,281 @@
...
+ This topic provides a basic understanding and some key considerations about Helm preflight checks to help guide you in enabling them for your application.
+
+## About Helm Preflight Checks
+
+You define preflight checks based on your application needs. Replicated recommends that you use prelight checks to help ensure that your customers' environments support your application _before_ running the installation. 
+
+Preflight checks are not included by default, so you must enable them.
+Preflight checks run before running the `helm install` command, without hooks, to confirm the target cluster has the resources required for a successful install.
+
+When you supply the output of `helm template` to the preflight command as stdin, the command filters the stream and searches for and runs the following specifications:
```
jonquil2002, line 19: @xavpaice Why would someone want to use stdin with Helm preflights? Is stdin just a default process when running the `helm template` command? It seems so, since it's not specified with the `helm template oci://REGISTRY/APP_NAME/CHART --values FILENAME.yaml | kubectl preflight -` command. 

And do they have other options than using stdin? SOunds like stdin is the best practice and we'd only want to document that, but a lot of mention was made of stdin in your draft document and I am trying to make sure I understand it and communicate the why/use case clearly.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,281 @@
...
+      <th>Supports Templating?</th>
+      <td><p>Yes</p></td>
+    </tr>
+    <tr>
+      <th>Limitations</th>
+      <td><p>Not applicable</p></td>
+    </tr>
+</table>
+
+### Preflight Custom Resource
```
jonquil2002, line 53: Is Preflight custom resource a spec type that we want to document? It seems that we really want to steer vendors to using Secrets with templating, which would provide any conditional statements they need. It doesn't seem like the Preflight CR offers any advantage to Helm charts versus Secrets.

If Preflights custom resource is something we want to de-emphasize, could we remove it from the Replicated docs and document it on community instead?

docs/vendor/preflight-helm-defining.mdx
jonquil2002, line 19: Also, it feels like this whole paragraph can go away. Is it important for vendors to know what specification types stdin is looking for? And isn't this `helm template` command used for running any of the preflight spec types, including URLs or YAML files?

docs/vendor/preflight-running.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+## Run Preflights Checks
+
+Before running the `helm install` command, your customers run preflight checks using the local values file that you provide. This lets customers verify that their environments meet the installation requirements before they run the actual installation. You can ask them to send you the results of the preflight checks if needed.
+
+Run the preflights using a local values file to override chart defaults:
+
+```
+helm template oci://REGISTRY/APP_NAME/CHART --values FILENAME.yaml | kubectl preflight -
```
jonquil2002, line 26: Does this command work with running preflights from specs that don't support templating, like URLs and YAML files?

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,281 @@
...
+      - fail:
+          when: connected == false
+          message: Cannot connect to MySQL server
+      - fail:
+          when: version < 8.x
+          message: The MySQL server must be at least version 8
+      - pass:
+          message: The MySQL server is ready
+```
+## Define a Preflight Custom Resource as a Template
```
jonquil2002, line 241: @xavpaice Related to my earlier question, do we want to de-emphasize this spec type? If so, what do you think about moving this procedure over to a Community post?

docs/vendor/preflight-kots-defining.md
```diff
@@ -0,0 +1,98 @@
...
+import PreflightsKotsGuidance from "../partials/preflights/_preflights-kots-guidance.mdx"
+
+
+# Define KOTS Preflight Checks
+
+You define preflight checks based on your application needs. Preflight checks are not included by default. This procedure provides a basic understanding and some key considerations to help guide you. For more information about defining preflight checks, see [Preflight Checks](https://troubleshoot.sh/docs/preflight/introduction/), [Collecting Data](https://troubleshoot.sh/docs/collect/), and [Analyzing Data](https://troubleshoot.sh/docs/analyze/) in the Troubleshoot documentation.
+
+<PreflightsKotsGuidance/>
+
+To define preflight checks for KOTS, create a Preflight custom resource manifest file (`kind: Preflight`) in your release.
```
jonquil2002, line 12: @xavpaice This whole topic was reorganized to show a simple but complete example, followed by conceptual guidance, such as recommended collectors to consider.

I used this model for all of the Helm and KOTS preflight and support bundles, instead of using numbered procedures.

docs/vendor/support-bundle-helm-customizing.mdx
```diff
@@ -0,0 +1,286 @@
...
+      <th>Supports Templating?</th>
+      <td>Yes</td>
+    </tr>
+    <tr>
+      <th>Limitations</th>
+      <td><p>Not applicable</p></td>
+    </tr>
+  </table>
+
+#### ConfigMaps
```
jonquil2002, line 38: @xavpaice @paigecalvert WDYT about adding a ConfigMap table to preflights. Or removing this table altogether?

docs/vendor/support-bundle-helm-customizing.mdx
```diff
@@ -0,0 +1,286 @@
...
+<ConfigmapNote/>
+
+To customize a support bundle as a Secret, create a specification using `kind: Secret`. You must include the following fields:
+
+- `Label` - Label the secret definition as `troubleshoot.sh/kind: support-bundle`
+- `stringData` - Specify a `stringData` field with a key named `support-bundle-spec`
+
+The following example shows a support bundle specification as a Secret. For more examples, see the [Troubleshoot example repository](https://github.com/replicatedhq/troubleshoot/tree/main/examples/support-bundle) in GitHub.
+
+```yaml
```
jonquil2002, line 120: @xavpaice I took your basic example with your annotations and added more collectors and analyzers to them from the repo links your sent. The repo links are KOTS/KURL oriented, so I removed some. But I ket a few in here thinking they could work if the URIs are updated. PTAL at everything here. I'll call out the URIs specifically as well.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,280 @@
...
+      <th>Supports Templating?</th>
+      <td>Does not support templating, unless templating is being used on the server side.</td>
+    </tr>
+    <tr>
+      <th>Limitations</th>
+      <td><p>Does not work for air gap environments.</p><p>Cannot easily add hooks in the future.</p><p>Burdensome for the customer if there are multiple URLs that must be run.</p></td>
+    </tr>
+  </table>
+
+## Define Preflight Checks as Secrets
```
paigecalvert, line 121: Thinking about the way the content is split up on the page, and I wanted to propose an alternative for your consideration.

One of the main workloads for vendors here is _writing the preflight spec_. That includes writing the Collectors and the Analyzers (and sort of the Redactors) to define your own preflight checks. This is the exact same, regardless of wherever you end up saving the YAML you write. It's also the same for KOTS and Helm distributions. I think this is the really challenging part for vendors where we could put in most of our documentation effort around explanations and examples. 

The other part is _where do I save this preflight spec that I write?_ . And for Helm distributions, that's either a Secret or a Preflight CR, as you've described. 

Maybe there is a way for the headings to reflect these two phases? First, pick where you are going to write it. Then, write the collectors and analyzers and here's how you do it

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,280 @@
...
+and [Analyzing Data](https://troubleshoot.sh/docs/analyze/) in the Troubleshoot documentation.
+
+To define preflight checks as a Secret, create a specification using `kind: Secret`. You must include the following fields:
+
+- `Label` - Label the secret definition as `troubleshoot.sh/kind: preflight`
+- `stringData` - Specify a `stringData` field with a key named `preflight.yaml` so that the preflight binary can use this Secret when it runs from the CLI
+
+The following example shows preflight checks using a Secret. For more examples, see the [Troubleshoot example repository](https://github.com/replicatedhq/troubleshoot/tree/main/examples/preflight) in GitHub.
+
+```yaml
```
paigecalvert, line 137: I think this example needs a lot more explanation, especially because it's pretty complex. For example, what are each of the collectors collecting? What are analyzers doing?

Then the other big piece I think is the template functions, which are designed to customize the checks based on the specific environment. I wonder if we might need a separate topic or section about templating. To steal the name of the lab that Chuck is working on, maybe that's something like "Adapting Preflights and Support Bundles to Customer Configurations"

As is, I found it a little overwhelming to see this big example up front, without understanding the pieces.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,280 @@
...
+Add collectors to define information to be collected for analysis during the analyze phase. For example, you can collect information about the MySQL version that is running in a cluster:
+
+```yaml
+   spec:
+      collectors:
+        - mysql:
+        collectorName: mysql
+        uri: 'repl{{ ConfigOption "db_user" }}:repl{{ConfigOption "db_password" }}@tcp(repl{{ ConfigOption "db_host" }}:repl{{ConfigOption "db_port" }})/repl{{ ConfigOption "db_name" }}'
+```
+Replicated recommends replacing using a template function for the URI to avoid exposing sensitive information. For more information about template functions, see [About Template Functions](/reference/template-functions-about).
```
paigecalvert, line 208: These template functions do not work with Helm installation (ConfigOption is a replicated template function). These would need to be pointing to something in the Helm values file.

Could also use some additional explanation about what's going on with the template functions, and how they map to the values.yaml file.

Maybe we could also have some simple examples that do not include templating?

paigecalvert: Do you think it would also be helpful to point to the list of collectors in the troubleshoot docs here? https://troubleshoot.sh/docs/collect/all/? As is, I don't have anywhere else to go from here. So, if I wanted to check the version of MySQL then I'm good. But I'm not empowered to learn more

docs/vendor/support-bundle-generating.mdx
```diff
@@ -0,0 +1,31 @@
+import InstallPlugin from "../partials/support-bundles/_install-plugin.mdx"
+import GenerateBundle from "../partials/support-bundles/_generate-bundle.mdx"
+
+
+# Generating Support Bundles
+
+For Helm charts and KOTS installations, support bundles can be generated using these CLI commands.
```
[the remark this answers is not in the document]
jonquil2002: Yes indeed. :)

docs/vendor/preflight-running.md
```diff
@@ -0,0 +1,52 @@
...
+
+You can run Helm preflight checks before running `helm install` and during the installation of an application. Additionally, your preflight checks can include one or more specification types and resources.
+
+## Prerequisite
+
+Customers must install the preflight plug-in, which is a client-side utility that adds a single binary to their path:
+
+```
+helm registry login registry.replicated.com --username USERNAME --password PASSWORD
+curl https://krew.sh/preflight | bash
```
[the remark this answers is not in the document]
jonquil2002: @xavpaice The Troubleshoot Gdoc had this entire command listed for setting up the preflight plug in, followed by another command to run the preflights.

But typically user would just run `curl https://krew.sh/preflight | bash` to install the plug-in. So is there any reason why they'd want to run `helm registry login registry.replicated.com --username USERNAME --password PASSWORD` with the curl command to install the preflight plugin? It doesn't make sense to me to log into a registry at this point.

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,91 @@
...
+### Analyzers
+Analyzers use the post-redacted data from the collectors to identify issues. The outcomes that you specify are displayed to customers.
+
+Analyzer outcomes for preflight checks differ from the outcomes for support bundles in terms of how they are used:
+
+- Preflight checks use analyzers to determine pass, fail, and warning outcomes, and display messages to a customer. For example, you can define a fail or warning outcome if the Kubernetes version on the cluster does not meet the minimum version that your application supports.
+
+- Support bundles use analyzers to help identify potential problems. When a support bundle is uploaded to the vendor portal, it is extracted and automatically analyzed. The goal of this process is to surface known issues or hints of what might be a problem. Analyzers produce outcomes that contain custom messages to explain what the problem might be.
+
+## Specification Types
```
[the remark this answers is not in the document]
jonquil2002: Agreed, it was conflated.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,280 @@
...
+                  message: Cannot connect to MySQL server
+              - fail:
+                  when: version < 8.x
+                  message: The MySQL server must be at least version 8
+              - pass:
+                  message: The MySQL server is ready
+        {{ end }}
+```
+
+### Collectors
```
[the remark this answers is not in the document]
jonquil2002: It does. They were indented because I think the Preflight custom resource section might go away. But I can bump them out for now, and then reorder them if the other section stays. It does apply to both.

docs/vendor/preflight-helm-defining.mdx
jonquil2002: I changed out the templating to Helm, added an About Helm Templates topic, and added more xrefs to the collectors and analyzers throughout all of the Define/Customize topics using a partial.

docs/vendor/preflight-helm-defining.mdx
jonquil2002: I changed out the templating to Helm, added an About Helm Templates topic, and added more xrefs to the collectors and analyzers throughout all of the Define/Customize topics using a partial.

I also got rid of the big example up front and went with smaller ones for these sections. Then added the big example at the end with an xref to the Troubleshoot site for collector and analyzer explanations.

docs/vendor/preflight-sb-helm-templates-about.md
```diff
@@ -0,0 +1,111 @@
+# About Helm Templates
```
jonquil2002, line 1: @xavpaice I added this new topic to give a more in-depth explanation about using Helm templates with Replicated. I pulled some of your info and examples from the Gdoc. Please review this topics carefully. I am  uncertain whether the examples and explanations are entirely correct.

docs/vendor/preflight-sb-helm-templates-about.md
```diff
@@ -0,0 +1,111 @@
...
+                   when: connected == false
+                   message: Cannot connect to MySQL server
+               - fail:
+                   when: version < 8.x
+                   message: The MySQL server must be at least version 8
+               - pass:
+                   message: The MySQL server is ready
+{{ end }}
+
+{{- $_ := mergeOverwrite .Values (include "myApplication.preflights" . | fromYaml) -}}
```
jonquil2002, line 68: @xavpaice Is there anything specific or additional context the a vendor needs to know about this mergeOverwrite template?

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+spec:
+      {{ if eq .Values.global.mariadb.enabled false }}
+      collectors:
+          - mysql:
+            collectorName: mysql
+            uri: '{{ .Values.global.externalDatabase.user }}:{{ .Values.global.externalDatabase.password }}@tcp({{ .Values.global.externalDatabase.host }}:{{ .Values.global.externalDatabase.port }})/{{ .Values.global.externalDatabase.database }}?tls=false'
+      {{ end }}
+```
+
+Replicated recommends using a Helm template for the URI to avoid exposing sensitive information. The Helm template maps to the `values.yaml` file to populate the connection string for the MySQL database.
```
jonquil2002, line 123: @xavpaice Is this accurate - that we recommend using the template for URI to avoid exposing sensitive info? I know it's true for KOTS, but not sure about Helm.

xavpaice: We'd recommend using templates to avoid exposing sensitive things anywhere in the chart - not just the URI.  Depends on the URI though - unlikely that http://kots.io is sensitive, for example

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+      <th>URLs</th>
+      <td><p>Specifications can be hosted online at a URL. Easy to iterate against as a team.</p><p>This also allows updating support bundle specifications in between application updates to notify customers of potential problems and fixes.</p></td>
+    </tr>
+    <tr>
+      <th>OCI Registry</th>
+      <td><p>Specifications are located in an image registry.</p></td>
+    </tr>
+    <tr>
+      <th>stdin</th>
+      <td><p>Specifications are provided as stdin from a Helm chart using <code>-</code> as the argument. Can be used by Helm installations, and is compatible with using Helm templates.</p></td>
```
paigecalvert, line 94: >Can be used by Helm installations

To confirm: can the others _not_ be used by Helm installations?

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+Preflight and support bundle specifications can be located and accessed in various ways. 
+
+Additionally, using multiple specifications and multiple specification locations are supported. For more information about using:
+
+- Multiple specifications for preflight or support bundle specifications, see [About Modular Specifications](support-modular-support-bundle-specs).
+- URLs for support bundles, see [About Online Support Bundle Specifications](support-online-support-bundle-specs).
+
+The following table gives a brief description of the available options to store and access your preflight and support bundle specifications:
+
+<table>
```
paigecalvert, line 75: Something I found myself wondering overall with this table is which apply to KOTS installs, which to Helm installs, and which to both?

Here's my best guess on that:

- (Both Types of Install?) User can run `kubectl preflight` (and pass a yaml file, a URL, or the location of the spec at an image registry)
- (Helm Install Only) User can run `helm template` command with `| preflight -` when the specs are included in your helm chart
- (KOTS Install Only) Preflights automatically run in the admin console or kots CLI during install/upgrade. Users can collect support bundles from the admin console

Perhaps an extra column to call this out, or just adding to the description to clarify?

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+      <th>OCI Registry</th>
+      <td><p>Specifications are located in an image registry.</p></td>
+    </tr>
+    <tr>
+      <th>stdin</th>
+      <td><p>Specifications are provided as stdin from a Helm chart using <code>-</code> as the argument. Can be used by Helm installations, and is compatible with using Helm templates.</p></td>
+    </tr>
+  </table>
+
+  For information about how to write the specifications, see:
```
paigecalvert, line 98: This felt a little easy to miss under the table--maybe it does above?

docs/vendor/preflight-running.md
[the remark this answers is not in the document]
paigecalvert: @jonquil2002 assuming the vendor has their preflight specs in their helm chart, then the user would need to log in the registry in order to authenticate and run the helm template command against that chart

You can see in the command below it's telling the user to run `helm template oci://REGISTRY/APP_NAME/CHART`. They need to log in to that registry first before they can run helm template

docs/vendor/preflight-running.md
```diff
@@ -0,0 +1,52 @@
...
+    helm template oci://REGISTRY/APP_NAME/CHART --values FILENAME.yaml | kubectl preflight -
+    ```
+    Replace:
+
+    - `REGISTRY` with the registry domain. For example, registry.domain.com
+    - `APP_NAME` with the name of the application.
+    - `CHART` with the name of the Helm chart.
+    - `FILENAME` with the name of the local values file.
+
+- To run preflights with templating, such as with a Secret or a Preflight custom resource:
```
paigecalvert, line 39: I don't think I understand these two options. It looks like the only difference is that the user is setting a value from the command line in the second one.

docs/vendor/preflight-support-bundle-about.md
xavpaice: Only preflight supports stdin - support-bundle does not.

Anything that generates a stream of text (`cat file.yaml`) can be used as the input.  Helm is just an example.

docs/vendor/support-online-support-bundle-specs.md
```diff
@@ -1,17 +1,27 @@
...
+You can make the definition of one or more support bundle specifications available online in a source repository and link to it from the specifications in the cluster. This approach lets you update collectors and analyzers outside of the application release and notify customers of potential problems and fixes in between application updates.
 
-You can keep the online specification current with collectors and analyzers to manage bug fixes. When a customer generates a support bundle, the online specification can detect those potential problems in the cluster and let them know know how to fix it. Without the URI link option, you must wait for the next time your customers update their applications or Kubernetes versions to get notified of potential problems. The URI link option is particularly useful for customers that do not update their application routinely.
+The schema supports a `uri://` field that, when set, causes the support bundle generation to use the online specification. If the URI is unreachable or unparseable, any collectors or analyzers in the specification in the cluster are used as a fallback.
```
xavpaice, line 5:
```suggestion
The schema supports a `uri:` field that, when set, causes the support bundle generation to use the online specification. If the URI is unreachable or unparseable, any collectors or analyzers in the specification in the cluster are used as a fallback. 
```

docs/vendor/support-online-support-bundle-specs.md
```diff
@@ -1,17 +1,27 @@
...
+You update collectors and analyzers in the online specification to manage bug fixes. When a customer generates a support bundle, the online specification can detect those potential problems in the cluster and let them know know how to fix it. Without the URI link option, you must wait for the next time your customers update their applications or Kubernetes versions to get notified of potential problems. The URI link option is particularly useful for customers that do not update their application routinely.
+
+If you are using a modular approach to designing support bundles, you can use multiple online specifications. Each specification supports one URI link. For more information about modular specifications, see [About Modular Support Bundles](support-modular-support-bundle-specs).
+
+## Limitations
+
+The following limitations apply to using URLs to store support bundle specifications:
+
+- URLs must be manually typed in to generate a support bundle, and there might be multiple strings and long strings.
+- Vendors must provide customers with the URL strings.
```
xavpaice, line 16: the URL is part of the spec - as long as the customer has the spec that's all they need.  If the URL is in the spec installed by the application, just running `support-bundle --load-cluster-specs` would use the URL field in the collected specs to update each one if possible.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+## Create a Secret
+
+<SecretsConfig/> 
+
+You must include the following:
+
+- `Label` - Label the secret definition as `troubleshoot.sh/kind: preflight`
+- `stringData` - Specify a `stringData` field with a key named `preflight.yaml` so that the preflight binary can use this Secret when it runs from the CLI
+- `{{ .Release.Namespace }}` to ensure that the specification is scoped to the release namespace.
```
paigecalvert, line 44: I notice that the example below has `.Release.Name` Is this supposed to say `.Release.Namespace` or `.Release.Name`? (I suspect name? Since the field is `name:`)

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+    {{ end }}
+    ```
+
+1. Configure your `values.yaml` file with Helm templating to support the Preflight custom resource.
+
+  **Example:**
+
+  ```yaml
+  renderpreflights: example-name
+  ```
```
paigecalvert, line 96: I'm not sure if this matters, but should this show `renderpreflights` set to `true`? Maybe `{{ if .Values.renderpreflights }}` just checks to see if `renderpreflights` is set at all, in which case any value would work, but for the sake of ease of understanding, maybe `true` is easier to interpret?

docs/vendor/preflight-kots-defining.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+## About KOTS Preflight Checks
+
+<PreflightsAbout/> 
+
+Preflight are included in your application package and run automatically before installing the application.
+
+<PreflightsSpecLocations/>
+
+## Input Kind
```
paigecalvert, line 24: I don't think I follow what this is saying for KOTS. When packaging your application for KOTS install, you just add the Preflight CR to your release, right?

docs/vendor/preflight-kots-defining.mdx
```diff
@@ -0,0 +1,281 @@
...
+When you run preflight checks with KOTS, you must use the Preflight custom resource (`kind: Preflight`).
+
+You can wrap the Preflight custom resource in the following inputs:
+
+- Secret (kind: Secret)
+- ConfigMap (kind: ConfigMap)
+
+Secrets and Config maps are installed in the cluster. Replicated recommends using Secrets to protect secure information in the cluster.
+
+## Create a Secret
```
paigecalvert, line 35: Similar to above, I think this section can be removed from the KOTS-specific topic, right?

docs/vendor/support-bundle-helm-customizing.mdx
```diff
@@ -0,0 +1,282 @@
...
+## Input Kind
+
+You can create support bundle specification using the following kinds:
+
+- Secret (`kind: Secret`)
+- Support Bundle custom resource (`kind: SupportBundle`)
+
+## Limitations
+
+YAML file and URL locations do not support Helm templating.
```
paigecalvert, line 34: Recommend adding a bit more detail here. I'm not sure I follow what "the YAML file and URL locations" means in this context, and it might be assuming that someone has read the table in the Overview topic in detail. 

I think this is saying that, if you want to use templating in your specs, then you have to either have the spec in a Secret so that you can run `helm template` to render it? If so, this might actually just be something that can be covered in the Helm Templating topic that you added

docs/vendor/support-bundle-helm-customizing.mdx
```diff
@@ -0,0 +1,282 @@
...
+<PreflightsSbNote/>
+
+<PreflightsSpecLocations/>
+
+## Input Kind
+
+You can create support bundle specification using the following kinds:
+
+- Secret (`kind: Secret`)
+- Support Bundle custom resource (`kind: SupportBundle`)
```
paigecalvert, line 30: It looks like the only option shown in this topic is "Define Support Bundles as Secrets". If we don't want to explain the option of creating a SupportBundle kind that you share through a URL, file, or at a registry link, then maybe this section can be deleted?

docs/vendor/support-bundle-helm-customizing.mdx
```diff
@@ -0,0 +1,282 @@
...
+    apiVersion: troubleshoot.sh/v1beta2
+    kind: SupportBundle
+    metadata:
+      name: support-bundle
+    spec:
+      collectors: []
+      analyzers: []
+```
+
+## Define the Support Bundle Specification
```
paigecalvert, line 71: All the detail you have in this section and the corresponding preflight section is awesome!

paigecalvert (commented):
sending over a few more comments :)

docs/vendor/preflight-sb-helm-templates-about.md
```diff
@@ -0,0 +1,111 @@
...
+          - fail:
+              when: "== Init:Error"
+              message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in an Init:Error state.` }}
+          - fail:
+              when: "== Init:CrashLoopBackOff"
+              message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in an Init:CrashLoopBackOff state.` }}
+          - fail:
+              when: "!= Healthy" # Catch all unhealthy pods. A pod is considered healthy if it has a status of Completed, or Running and all of its containers are ready.
+              message: {{ `Pod {{ .Namespace }}/{{ .Name }} is unhealthy with a status of {{ .Status.Reason }}.` }}
+```
```
xavpaice, line 111: This example is from the values.yaml from the library chart usage - we can't use that now.

Here:

```yaml
# This is the support bundle config secret that will be used to generate the support bundle
apiVersion: v1
kind: Secret
metadata:
  labels:
    troubleshoot.sh/kind: support-bundle
  name: {{ .Release.Name }}-support-bundle
  namespace: {{ .Release.Namespace }}
type: Opaque
stringData:
  # This is the support bundle spec that will be used to generate the support bundle
  # Notes: we use {{ .Release.Namespace }} to ensure that the support bundle is scoped to the release namespace
  # We can use any of Helm's templating features here, including {{ .Values.someValue }}
  support-bundle-spec: |
    apiVersion: troubleshoot.sh/v1beta2
    kind: SupportBundle
    metadata:
      name: support-bundle
    spec:
      collectors:
        - clusterInfo: {}
        - clusterResources: {}
        - logs:
            selector:
              - app=someapp
            namespace: {{ .Release.Namespace }}
      analyzers:
        - clusterPodStatuses:
            name: unhealthy
            namespaces:
              - default
              - myapp-namespace
            outcomes:
              - fail:
                  when: "== CrashLoopBackOff"
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in a CrashLoopBackOff state.` }}
              - fail:
                  when: "== ImagePullBackOff"
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in a ImagePullBackOff state.` }}
              - fail:
                  when: "== Pending"
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in a Pending state.` }}
              - fail:
                  when: "== Evicted"
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in a Evicted state.` }}
              - fail:
                  when: "== Terminating"
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in a Terminating state.` }}
              - fail:
                  when: "== Init:Error"
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in an Init:Error state.` }}
              - fail:
                  when: "== Init:CrashLoopBackOff"
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is in an Init:CrashLoopBackOff state.` }}
              - fail:
                  when: "!= Healthy" # Catch all unhealthy pods. A pod is considered healthy if it has a status of Completed, or Running and all of its containers are ready.
                  message: {{ `Pod {{ .Namespace }}/{{ .Name }} is unhealthy with a status of {{ .Status.Reason }}.` }}


```

xavpaice: again, the spec itself could be done differently to make it clearer, but this is an example.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+## About Helm Preflight Checks
+
+<PreflightsAbout/>
+
+Preflight checks run using a `helm template` command before running the installation to confirm the target cluster has the resources required for a successful installation.
+
+## Input Kind
+
+When you run `helm template` command as stdin, the command filters the stream and runs the following inputs:
```
xavpaice, line 26: you run `helm template  <args>` to produce a stream of stdout, and pipe the result of that to `preflight -`.  The `preflight` command filters the stream.

docs/partials/preflights/_secrets-config.mdx
```diff
@@ -0,0 +1,8 @@
+Replicated recommends writing your preflight checks specification in one or more Secrets. Using Secrets protects secure information in a cluster.
+
+:::note
+Alternatively, you can use a ConfigMap (`kind: configMap`) if the specification will not collect private information from the cluster.
```
xavpaice, line 4:
```suggestion
Alternatively, you can use a ConfigMap (`kind: configMap`) if the specification will not contain private information.
```

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+    spec:
+      collectors:[]
+      analyzers:[]
+```
+
+Next, define the preflight checks specification by adding collectors and analyzers. Optionally, you can add to the default redactors. For more information, see [Define the Preflight Checks Specification](#preflights).
+
+## Create a Preflight Custom Resource
+
+The Preflight custom resource lets you provide preflight checks in a Helm template. If you use this input, you must use conditional statements in the custom resource because it is not installed in the cluster. Instead, the custom resource is read using stdin.
```
xavpaice, line 68: it is not able to be installed in the cluster because the definition is not installed.

xavpaice, line 68:
```suggestion
The Preflight custom resource lets you provide preflight checks in a Helm template without using `stringData`. If you use this input, you must use conditional statements in the custom resource because it is not able to be installed in the cluster.
```

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+<PreflightsCrNote/>
+
+To define a Preflight custom resource as a template:
+
+1. Create a Preflight custom resource YAML file using `kind: Preflight`. Wrap the custom resource template in an `{{ if` so that it is only rendered when specified in the `values.yaml` file.
```
xavpaice, line 74:
```suggestion
1. Create a Preflight custom resource YAML file using `kind: Preflight`. Wrap the custom resource template in an `{{ if` so that it is only rendered when specified in the `values.yaml` file or on the `helm` CLI.
```

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+    kind: Preflight
+    metadata:
+      name: example-name
+    spec:
+      collectors: []
+      analyzers: []
+    {{ end }}
+    ```
+
+1. Configure your `values.yaml` file with Helm templating to support the Preflight custom resource.
```
xavpaice, line 90:
```suggestion
1. Configure your `values.yaml` file to enable the Preflight custom resource.
```

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+  **Example:**
+
+  ```yaml
+  renderpreflights: example-name
```
xavpaice, line 95:
```suggestion
  renderpreflights: true
```

docs/vendor/preflight-helm-defining.mdx
xavpaice, line 90: alternately, rather than edit the values.yaml:

```
helm template oci://chartlocation --values values.yaml --set renderpreflights=true | preflight -
```

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+          when: version < 8.x
+          message: The MySQL server must be at least version 8
+      - pass:
+          message: The MySQL server is ready
+```
+<PreflightsDefineXref/>
+
+<AnalyzersNote/>
+
+### `strict` Analyzers
```
xavpaice, line 157: `strict` is for KOTS only, it does nothing when installing with Helm

docs/partials/preflights/_preflights-add-strict.mdx
```diff
@@ -0,0 +1,9 @@
+You can set any preflight analyzers to `strict: true` if you want to enforce requirements for the chosen analyzers. For more information about strict preflight checks, see [`strict`](https://troubleshoot.sh/docs/analyze/#strict) in the Troubleshoot documentation.
```
xavpaice, line 1: `strict` only applies to KOTS - not Helm.

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,260 @@
...
+```
+<PreflightsDefineXref/>
+
+## Redactors
+
+<RedactorsAbout/>
+
+## Example
+
+The following example shows a preflight support bundle specification in a Secret for a Helm installation. For more examples, see the [Troubleshoot example repository](https://github.com/replicatedhq/troubleshoot/tree/main/examples/preflight) in GitHub.
```
xavpaice, line 192:
```suggestion
The following example shows a preflight specification in a Secret for a Helm installation. For more examples, see the [Troubleshoot example repository](https://github.com/replicatedhq/troubleshoot/tree/main/examples/preflight) in GitHub.
```

docs/vendor/preflight-kots-defining.mdx
xavpaice, line 35: KOTS doesn't use a secret for preflights

docs/vendor/preflight-kots-defining.mdx
```diff
@@ -0,0 +1,281 @@
...
+              message: The MySQL server must be at least version 8
+          - pass:
+              message: The MySQL server is ready
+```
+
+<PreflightsDefineXref/>
+
+## Example
+
+The following example shows a preflight support bundle specification wrapped in a Secret for a KOTS installation. For more examples, see the [Troubleshoot example repository](https://github.com/replicatedhq/troubleshoot/tree/main/examples/preflight) in GitHub.
```
xavpaice, line 164: no need for the secret for KOTS

docs/vendor/support-bundle-helm-customizing.mdx
xavpaice, line 34: this is a bit confusing in this location.  I think what you're getting at is that running support-bundle with a spec that is given via a URL or a file, means that there's no opportunity to use Helm to template it?  It lost me because we haven't mentioned that in this doc, it's probably better in the About: section

docs/vendor/support-bundle-helm-customizing.mdx
```diff
@@ -0,0 +1,282 @@
...
+
+### Pod Log Collectors
+
+<SupportBundleAddLogs/>
+
+The following example shows a Pod log collector for nginx scoped to a specific namespace using `{{ .Release.Namespace }}`:
+
+```yaml
+spec:
+      collectors:
```
xavpaice, line 105: indentation: should be 2 spaces here

docs/vendor/support-bundle-helm-customizing.mdx
```diff
@@ -0,0 +1,282 @@
...
+            outcomes:
+              - fail:
+                  message: Didn't say hi.
+              - pass:
+                  message: Said hi!
+```
+
+## Next Steps
+
+- When using Secrets, you can also add Helm templates for the specification in the `values.yaml` file.
```
xavpaice, line 280: you can't put Helm templates in a values.yaml file - maybe just delete this sentence?

docs/vendor/support-bundle-kots-secrets-configmaps.md
```diff
@@ -0,0 +1,64 @@
+# Configure KOTS Discoverable Support Bundles
```
xavpaice, line 1: I don't think we need this page at all - KOTS uses configMap under the hood, but there's no need for a Vendor to be doing that when they're using KOTS. `kubectl supportbundle --load-cluster-specs` is equivalent to what happens in KOTS now, there's no difference and certainly no need to list secrets and specify them all on the CLI

docs/vendor/preflight-running.md
xavpaice: that's right, the second example just avoids having a values.yaml with those settings.  Both use templating, the `--set` just overrides what's in `values.yaml`

xavpaice:
Couple of notes that don't appear to be in this change, but probably should be:

https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/reference/custom-resource-preflight - "Default support bundles are automatically included with releases, and can be customized." - only for KOTS, this is in a generic section.

Same doc: we give an example of `strict` in a preflight, but that's KOTS only

https://deploy-preview-1205--replicated-docs-upgrade.netlify.app/reference/custom-resource-redactor "The default redactors cannot be disabled" - only when run from the Admin Console, if you use the CLI you can disable them.

Note also the `kind: Redactor` also can't be added directly to the cluster, so might not be applicable to Helm installs at this stage.  Maybe we need to make these sections KOTS only?

docs/vendor/preflight-running.md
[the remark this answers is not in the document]
jonquil2002: Thanks, I'm groking it now. Finally!

docs/vendor/preflight-sb-helm-templates-about.md
jonquil2002: Thanks :)

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+### Analyzers
+Analyzers use the post-redacted data from the collectors to identify issues. The outcomes that you specify are displayed to customers.
+
+Analyzer outcomes for preflight checks differ from the outcomes for support bundles in terms of how they are used:
+
+- Preflight checks use analyzers to determine pass, fail, and warning outcomes, and display messages to a customer. For example, you can define a fail or warning outcome if the Kubernetes version on the cluster does not meet the minimum version that your application supports.
+
+- Support bundles use analyzers to help identify potential problems. When a support bundle is uploaded to the vendor portal, it is extracted and automatically analyzed. The goal of this process is to surface known issues or hints of what might be a problem. Analyzers produce outcomes that contain custom messages to explain what the problem might be.
+
+## Specification Locations and Access
```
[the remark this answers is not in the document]
jonquil2002: Love it. Thank you!

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+
+<table>
+    <tr>
+      <th width="25%">Type</th>
+      <th width="75%">Description</th>
+    </tr>
+    <tr>
+      <th>YAML Files</th>
+      <td>Specifications can be distributed directly to your customers and stored on their local file systems. Can be useful for air gap installations.</td>
+    </tr>
```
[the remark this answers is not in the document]
jonquil2002: Good observations/suggestions. I revised this section substantially. I think it may actually be usable now, but let's pair and see.

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+    <tr>
+      <th>URLs</th>
+      <td><p>Specifications can be hosted online at a URL. Easy to iterate against as a team.</p><p>This also allows updating support bundle specifications in between application updates to notify customers of potential problems and fixes.</p></td>
```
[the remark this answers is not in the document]
jonquil2002: Only online/URL specs can be updated automatically in between application releases. The word "automatically" is missing here. There are dependencies - the URI must exist in the local support bundle spec to begin with, but we xref them to that topic for that deeper info.

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+    </tr>
+    <tr>
+      <th>OCI Registry</th>
+      <td><p>Specifications are located in an image registry.</p></td>
```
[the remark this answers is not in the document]
jonquil2002: Added new sections below this to guide them to the commands for running and related topics.

docs/vendor/preflight-support-bundle-about.md
```diff
@@ -0,0 +1,103 @@
...
+    </tr>
+    <tr>
+      <th>stdin</th>
+      <td><p>Specifications are provided as stdin from a Helm chart using <code>-</code> as the argument. Can be used by Helm installations, and is compatible with using Helm templates.</p></td>
+    </tr>
```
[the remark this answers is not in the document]
jonquil2002: I don't think so, because that's about how you generate the support bundle, not about distributing the spec. But...your thought has led to the idea of adding sections here to give an overview about running the specs.

docs/vendor/support-online-support-bundle-specs.md
jonquil2002: Changed manually. Diff was outdated.

docs/vendor/preflight-kots-defining.mdx
jonquil2002: Ugh, yes, you do. It's the SupportBundle CR that can be wrapped in a Secret for KOTS.

docs/vendor/preflight-kots-defining.mdx
jonquil2002: Yep.

docs/vendor/support-bundle-helm-customizing.mdx
jonquil2002: Added the section. I thought we wanted to list all of the possibilities and then just fully document the ones we recommend. But, there is value in adding more info about this.

docs/vendor/support-bundle-helm-customizing.mdx
jonquil2002: Teamwork makes the dream work! Thanks for the compliment :)

docs/vendor/preflight-sb-helm-templates-about.md
jonquil2002: Thanks! The library chart example was just a placeholder, but I didn't get a chance to point it out for you. Glad you caught it :)

docs/vendor/preflight-helm-defining.mdx
jonquil2002: Thanks. I changed "it" to "custom resource definition". So, it ends up as "If you use this input, you must use conditional statements in the custom resource because the custom resrouce definition is not installed in the cluster."

docs/partials/preflights/_preflights-add-strict.mdx
jonquil2002: I was wondering about that. Since it's on the Troubleshoot site, I thought it might apply more broadly than for KOTS. I've removed this section from Helm.

docs/vendor/preflight-helm-defining.mdx
jonquil2002: Fixed

docs/vendor/preflight-helm-defining.mdx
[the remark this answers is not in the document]
jonquil2002: Not sure I understand which one you're referring to? Is  this the one that isn't required? It is in the example under `metadata`

 "`name`: Specify a `"{{ .Release.Namespace }}-NAME` to ensure that the specification is scoped to the release namespace. Replace `NAME` with the namespace where the release will be deployed.

docs/vendor/preflight-helm-defining.mdx
jonquil2002: Fixed.

docs/vendor/preflight-helm-defining.mdx
jonquil2002: Paige and I decided to only document one way of running it for now, in the topic "Running Preflights for Helm Installation". For this configuration topic, we think the developer will already know that they can use the --set flag as an alternative method to editing.

docs/vendor/preflight-kots-defining.mdx
jonquil2002: Yep, too many moving parts. Thanks for catching that.

docs/vendor/support-bundle-kots-secrets-configmaps.md
jonquil2002: Ok, based on our discussion, this page has been removed.

docs/vendor/preflight-helm-defining.mdx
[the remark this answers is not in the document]
jonquil2002: We talked and this has been removed now.

docs/vendor/support-online-support-bundle-specs.md
jonquil2002: I commented out these limitations.

docs/vendor/preflight-kots-defining.mdx
xavpaice: For KOTS, we don't want this wrapped in a Secret - KOTS will not find it.

docs/vendor/preflight-running.md
```diff
@@ -0,0 +1,44 @@
...
+
+Before running the `helm install` command, your customers run preflight checks using the local values file that you provide. This lets customers verify that their environments meet the installation requirements before they run the actual installation. 
+
+The output shows the success, warning, or fail message each preflight check, depending on how they were configured.You can ask customers to send you the results of the preflight checks if needed.
+
+To run the preflights using a local values file to override chart defaults:
+
+    ```
+    helm template oci://REGISTRY/APP_NAME/CHART --values FILENAME.yaml | kubectl preflight -
+    ```
```
xavpaice, line 38: something went odd with the formatting here - it quoted all the replace parts too

docs/vendor/preflight-running.md
```diff
@@ -0,0 +1,44 @@
...
+    helm registry login REGISTRY_DOMAIN --username USERNAME --password PASSWORD
+    ```
+
+Replace:
+
+- `REGISTRY_DOMAIN` with the registry domain that contains the Helm chart.
+- `USERNAME` with the username that has access to the registry.
+- `PASSWORD` with the password for the registry.
+
+1. Run the following command to install the preflight plug-in using krew:
```
xavpaice, line 23: probably don't want to restart the numbering, maybe format it as something else

docs/vendor/preflight-running.md
```diff
@@ -0,0 +1,44 @@
...
+    ```
+    curl https://krew.sh/preflight | bash
+    ```
+
+## Run Preflights Checks
+
+Before running the `helm install` command, your customers run preflight checks using the local values file that you provide. This lets customers verify that their environments meet the installation requirements before they run the actual installation.
```
xavpaice, line 30: local values is optional - you could specify `--values xyz.yaml`, or `--set something=value`, or just skip that entirely.

xavpaice (commented):
A few minor changes to make, otherwise looks great

docs/vendor/preflight-kots-defining.mdx
jonquil2002: Good catch. Fixed that wording and removed the Secrets header.

docs/vendor/preflight-running.md
jonquil2002: Fixed. Thanks!

docs/vendor/preflight-running.md
jonquil2002: Fixed.

docs/vendor/preflight-running.md
jonquil2002: Made some changes here to reflect that.

jonquil2002: Also, we don't need to tell them every Helm flag option, as you mentioned when we talked. But my changes give them a couple of ways to do it, which seem like the two most likely ways they would need, I think.

docs/vendor/preflight-sb-helm-templates-about.md
```diff
@@ -0,0 +1,151 @@
+# About Helm Templates
```
paigecalvert, line 1: This topic is great. The title almost downplays how useful it is (ie, I wouldn't necessarily think there are some examples in here). What do you think about something like "About Using Helm Templates in Your Specifications" or "Using Template Functions in Your Specifications"?

paigecalvert (dismissed):
Amazing @jonquil2002 !!!! Just 2 small comments

docs/vendor/preflight-sb-helm-templates-about.md
jonquil2002: Good idea. Done!

docs/vendor/preflight-helm-defining.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+This topic provides a basic understanding and some key considerations about preflight checks for Helm installations to help guide you in defining them for your application.
+
+## About Helm Preflight Checks
+
+<PreflightsAbout/>
+
+Preflight checks run using a `helm template` command before running the installation to confirm the target cluster has the resources required for a successful installation.
+
+## Input Kind
```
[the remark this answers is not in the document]
jonquil2002: Nice, thanks!

xavpaice (approved):
This looks great, thanks for all the hard work!

PR #1209, opened 2023-06-07
Fix private registries content

paigecalvert:
https://deploy-preview-1209--replicated-docs.netlify.app/vendor/packaging-private-images#configure-access-to-an-external-registry

https://app.shortcut.com/replicated/story/79489/add-docs-about-setting-up-gcp-as-private-registry

### DIFF
--- a/docs/vendor/packaging-private-images.md
+++ b/docs/vendor/packaging-private-images.md
@@ -43,22 +43,134 @@ To configure access to your private images in an external registry using the ven
 
    <img src="/images/add-external-registry.png" alt="/images/add-external-registry.png" width="400px"></img>
 
-1. Click the **Other Registry** or **DockerHub** tab.
-1. Complete the fields in the dialog:
-<table>
-  <tr>
-    <th width="30%">Field</th>
-    <th width="70%">Instructions</th>
-  </tr>
-  <tr>
-    <td>Endpoint</td>
-    <td>Enter the endpoint, such as quay.io, index.docker.io, or gcr.io.</td>
-  </tr>
-  <tr>
-    <td>Username and Password</td>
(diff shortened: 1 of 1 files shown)


docs/vendor/packaging-private-images.md
```diff
@@ -43,22 +43,134 @@ To configure access to your private images in an external registry using the ven
...
+          <td>Endpoint</td>
+          <td>Enter the endpoint for the registry, such as example.registry.com.</td>
+        </tr>
+        <tr>
+          <td>Username and Password</td>
+          <td>Enter the username and password for an account that has pull access to the registry.</td>
+        </tr>
+      </table>
+
+1. For **Image name & tag**, enter the image name and image tag and click **Test** to confirm that the vendor portal can access the image. For example, `api:v1.0.1`.
```
paigecalvert, line 171: If users have an image located here: `my-registry.example.com/app-name/image:tag`, where would they indicate the `app-name` part? Does that go in the endpoint field, or in the image name and tag field?

docs/vendor/packaging-private-images.md
```diff
@@ -43,22 +43,134 @@ To configure access to your private images in an external registry using the ven
...
+
+      <table>
+        <tr>
+          <th width="30%">Field</th>
+          <th width="70%">Instructions</th>
+        </tr>
+        <tr>
+          <td>Endpoint</td>
+          <td>Enter the endpoint for the registry, such as index.docker.io.</td>
+        </tr>
```
paigecalvert, line 84: For the **Endpoint** field for each provider, I wasn't sure what the endpoint is for each provider and/or if it's possible that the endpoint varies for a given provider. As in, for GCR, is the endpoint _always_ gcr.io? Or might it vary? I suspect it might vary since the UI doesn't automatically populate the field.

docs/vendor/packaging-private-images.md
alicenstar: Eh.. I'm actually not sure. Maybe @marccampbell or @divolgin would know the answer.

alicenstar (dismissed):
This looks good to me, but I am unsure about the app name/endpoint question.

PR #1214, opened 2023-06-09
Don't mention overrides

ajp-io:


docs/partials/custom-domains/_use-custom-domain.mdx
```diff
@@ -28,4 +28,4 @@ To assign a custom domain to a channel:
 
 1. In the vendor portal, go to **Channels** and click the settings icon for the target channel.
 
-1. Under **Custom domain overrides**, in the drop-down for the target Replicated endpoint, select the domain to use for the channel. Select **No override** to use the default custom domain, or to use the default Replicated domain if no default custom domain is set. For more information about channel settings, see [Channel Settings](releases-about-channels#channel-settings) in _About Channels_.
\ No newline at end of file
+1. Under **Custom domains**, in the drop-down for the target Replicated endpoint, select the domain to use for the channel. Select **No override** to use the default custom domain, or to use the default Replicated domain if no default custom domain is set. For more information about channel settings, see [Channel Settings](releases-about-channels#channel-settings) in _About Channels_.
```
ajp-io, line 31: @paigecalvert These can be released with @miaawong's changes.

I updated the channel settings name, but I'll leave the wording for the rest of the paragraph. We took your advice and aren't using overrides anymore. I can record a quick demo for you, but I just uncovered something else I'm digging into.

paigecalvert (approved):

docs/partials/custom-domains/_use-custom-domain.mdx
paigecalvert: thanks for making the edit! approved so it's ready to merge

ajp-io: Oh I worded my message incorrectly. I meant that I'll leave the rest of the wording to you as far as how to word it so it doesn't say override.

paigecalvert: Ahh okay yeah I misunderstood. Sounds good, I'll do the following:
- Remove overrides in the paragraph/wherever else it appears
- Update the screenshot

ajp-io: Yeah, I figured you could come up with better wording than if I tried to rewrite that without overrides.

Because, to be clear, the concept of overrides still does exist. We're just messaging it differently as you mentioned in your first review.

paigecalvert: Right. So: The _concept_ still exists; the _terminology_ is being removed from the UI.

And yeah whenever available, a loom or eng story/screenshots would be helpful to make sure I have the right experience

PR #1224, opened 2023-06-20
fix(TLSCert): clarify the nil should be ()

DexterYan:
related to [sc-80547](https://app.shortcut.com/replicated/story/80547/replicated-release-lint-throw-an-helm-chart-missing-error-when-tlscert-return-empty-result-which-cause-chart-yaml-rendered)


paigecalvert (approved):
thanks @DexterYan !

PR #1253, opened 2023-07-03
82091: Custom Domains for replicated.app

jonquil2002:
SC: https://app.shortcut.com/replicated/story/82091/using-custom-domains-for-replicated-app

Preview: https://deploy-preview-1253--replicated-docs.netlify.app/vendor/custom-domains

### DIFF
--- a/docs/partials/custom-domains/_limitations.mdx
+++ b/docs/partials/custom-domains/_limitations.mdx
@@ -2,6 +2,6 @@ Using custom domains has the following limitations:
 
 - A single custom domain cannot be used for multiple endpoints. For example, a single domain can map to `registry.replicated.com` for any number of applications, but cannot map to both `registry.replicated.com` and `proxy.replicated.com`, even if the applications are different.
 
-- Custom domains cannot be used to alias api.replicated.com (platform market API) or other services.
+- Custom domains cannot be used to alias api.replicated.com (legacy customer-facing APIs) or other services.
 
 - Multiple custom domains can be configured, but only one custom domain can be the default for each Replicated endpoint. All configured custom domains work whether or not they are the default.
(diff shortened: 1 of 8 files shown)


docs/partials/custom-domains/_limitations.mdx
```diff
@@ -2,6 +2,6 @@ Using custom domains has the following limitations:
 
 - A single custom domain cannot be used for multiple endpoints. For example, a single domain can map to `registry.replicated.com` for any number of applications, but cannot map to both `registry.replicated.com` and `proxy.replicated.com`, even if the applications are different.
 
-- Custom domains cannot be used to alias replicated.app (release manifests), api.replicated.com (platform market API), or other services.
+- Custom domains cannot be used to alias api.replicated.com (platform market API) or other services.
```
dexhorthy, line 5: The phrase "platform market API" is verging on an internal use term. Even in the platform days, we tried hard not to use the phrase "market" w/ customers. I would recommend this:

```suggestion
- Custom domains cannot be used to alias api.replicated.com (legacy customer-facing APIs) or other services.
```

jonquil2002: Agreed, I'm going with the first one because it is more explicit for KOTS users.

jonquil2002:
@ajp-io Ok, agreed. I thought that the intent was to reuse this content as the overview, but I misunderstood the Slack/SC description. Also, that whole Private Registries section is such a mixed bag, not only do we need a separate Overview topic, but the TOC itself could use a little reorg. I'll fix this up.

jonquil2002:
Closing this PR in favor of starting a new PR. I'm going to make substantial TOC changes, which means I need to pull in the most current content from the main branch, esp. the new section [Reuse a Custom Domain](https://docs.replicated.com/vendor/custom-domains#reuse-a-custom-domain-for-another-application)

PR #1256, opened 2023-07-05
Change secret name to be more specific

ajp-io:
This secret name collides with a secret the SDK creates. So we should name it something more specific.


jonquil2002 (approved):
lgtm. Since this secret name is required, should we call that out rather than just having it in the example?

PR #1265, opened 2023-07-07
improve repeatable config items example

cbodonnell:
This PR updates the config group name for the repeatable items examples so that it's clear that the group name should be used in the `valuesByGroup` field. Previously the name was `ports`, which also was the name of the yaml field that was being used as well, so the example was not immediately clear from the example that the key in `valuesByGroup` needs to match the parent group name (we do doc this [here](https://docs.replicated.com/reference/custom-resource-config#repeatable-items)).

### DIFF
--- a/docs/reference/custom-resource-config.md
+++ b/docs/reference/custom-resource-config.md
@@ -655,7 +655,7 @@ A repeatable config item copies a YAML array entry or YAML document for as many 
 To make an item repeatable, set `repeatable` to true:
 
 ```yaml
-    - name: ports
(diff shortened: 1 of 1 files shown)


docs/reference/custom-resource-config.md
cbodonnell, line 1: Could use a sanity check to make sure I changed all of the group names and `valuesByGroup` fields from `ports` to `ports_group`

jonquil2002 (approved):
lgtm - searched the source code and didn't find any missing examples

PR #1275, opened 2023-07-13
document new cli feature which lists customers and instances for a specific application version

St0rmz1:
document new cli feature which lists customers and instances for a specific application version.

added a new flag for ls called --appversion

For example:
replicated customer ls --appversion 0.1.0

### DIFF
--- a/docs/reference/replicated-cli-customer-ls.md
+++ b/docs/reference/replicated-cli-customer-ls.md
@@ -4,7 +4,7 @@ import App from "../partials/replicated-cli/_app.mdx"
 
 # customer ls
 
-List all customers in your application.
+List the customers for your application.
 
 ## Usage
 ```bash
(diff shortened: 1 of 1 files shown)


ajp-io:
Can this be `--app-version` instead of `--appversion`? Looks like Marc suggested something similar [here](https://github.com/replicatedhq/replicated/pull/251/files#r1237386178).

docs/reference/replicated-cli-customer-ls.md
```diff
@@ -4,7 +4,7 @@ import App from "../partials/replicated-cli/_app.mdx"
 
 # customer ls
 
-List all customers in your application.
+List the customers for your application.
```
paigecalvert, line 7: edited since it's no longer necessarily "all" customers that you're listing

paigecalvert (approved):
thanks @St0rmz1 !

PR #1294, opened 2023-07-19
83209: Update Helm Install dialog and instructions

paigecalvert:


erquhart (approved):
Looks accurate 👍 There are differences when you're on a dev license in a builders team, but they're covered under the integration mode docs. I guess we could mention that and link here, not sure how much cross referencing we want to do.

PR #1297, opened 2023-07-21
Updating the planning questionnaire

diamonwiggins:
Refocuses the planning questionnaire to be focused on onboarding the Vendor vs. solely gathering details from the Vendor's end customer

### DIFF
--- a/docs/vendor/planning-questionnaire.md
+++ b/docs/vendor/planning-questionnaire.md
@@ -12,7 +12,7 @@ as well as your customers' technical expertise and expectations.
 
     - If your application is containerized:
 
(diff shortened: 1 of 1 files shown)


dexhorthy:
@paigecalvert @diamonwiggins while I agree this is a better document for onboarding a vendor, I'm wondering if there's a strong desire to also **retire** the existing end-customer deployment questionnaire. I think this doc has value in other contexts, even if it ends up buried deeper in the documentation, and I'd prefer not to remove it

diamonwiggins:
Closing this one as this page will be addressed as part of a larger effort to revamp our onboarding docs

PR #1309, opened 2023-07-26
Edit release flows for builders

paigecalvert:
https://deploy-preview-1309--replicated-docs.netlify.app/vendor/releases-creating-releases


static/images/helm-install-method-vendor-portal.png
paigecalvert, line 1: This was an out of date image from before the v2 option was added. Never had deleted it

docs/vendor/releases-creating-releases.mdx
```diff
@@ -24,11 +22,28 @@ To create and promote a release in the vendor portal:
...
+   
+     For any Helm charts that you add, in the **Select Installation Method** dialog, select the version of the HelmChart custom resource that KOTS will use to install the chart. kots.io/v1beta2 is recommended. For more information about the HelmChart custom resource, see [Configuring the HelmChart Custom Resource](helm-native-v2-using).
+
+     <img src="/images/helm-select-install-method.png" alt="select installation method dialog" width="550px"/>
 
-  You can click the plus icon to add a new, untitled YAML file.
+     [View a larger image](/images/helm-select-install-method.png)
 
-1. Edit the YAML files as needed, and click **Save release**. This saves a draft that you can continue to edit until you promote it.
+1. Click **Save release**. This saves a draft that you can continue to edit until you promote it.
```
paigecalvert, line 46: This ^ was all the changes that we needed to make to the flow. There's the new drop down and a dialog or dragging the Helm chart for Helm-only releases.

docs/vendor/releases-creating-releases.mdx
```diff
@@ -24,11 +22,28 @@ To create and promote a release in the vendor portal:
...
+1. Add your files to the release:
+   * For Helm-only releases, in the **Upload Helm Chart** dialog, drag and drop or select a Helm chart `.tgz` file.
+   
+     <img src="/images/upload-helm-chart-modal.png" alt="upload helm chart dialog" width="550px"/>
+
+     [View a larger image](/images/upload-helm-chart-modal.png)
+
+   * For KOTS releases, drag and drop your application files to the file directory in the YAML editor. You can also click the plus icon to add a new, untitled YAML file.
+   
+     For any Helm charts that you add, in the **Select Installation Method** dialog, select the version of the HelmChart custom resource that KOTS will use to install the chart. kots.io/v1beta2 is recommended. For more information about the HelmChart custom resource, see [Configuring the HelmChart Custom Resource](helm-native-v2-using).
```
paigecalvert, line 40: I kept this description fairly lightweight so as to not overload this procedure, then linked to another topic for more information. Let me know if you think this is sufficient

docs/vendor/releases-creating-releases.mdx
```diff
@@ -41,18 +56,19 @@ To create and promote a release in the vendor portal:
...
-      Enter a version label.
+      <p>Enter a version label.</p>
+      <p>If you have one or more Helm charts in your release, the vendor portal automatically populates this field. You can change the version label to any <code>version</code> specified in any of the <code>Chart.yaml</code> files included in the release.</p>
     </td>
   </tr>
   <tr>
     <td>Requirements</td>
     <td>
-      Select <strong>Prevent this release from being skipped during upgrades</strong> to mark the release as required. Applies only to releases that support installation with Replicated KOTS.
+      (KOTS Only) Select the <strong>Prevent this release from being skipped during upgrades</strong> to mark the release as required for KOTS installations. This option does not apply to installations with Helm.
```
paigecalvert, line 71: One other benefit of condensing this information into one topic is that there were previously two different places where the promote dialog options were explained (my fault 🤦‍♀️  ). So, I was able to make some minor updates here to capture that other info.

docs/vendor/replicated-sdk-using.md
```diff
@@ -38,18 +35,4 @@ To declare the Replicated SDK as a dependency:
...
-
-### Using the replicated CLI
-
-<HelmReleaseStepsCLI/>
-
-### Using the Vendor Portal
-
-<HelmReleaseSteps/>
\ No newline at end of file
+After you package your Helm chart with the Replicated SDK as a dependency, add the `.tgz` file to a release. For more information, see [Creating a Release with Your Helm Chart](helm-install-release).
```
paigecalvert, line 38: Linking rather than duplicating

docs/vendor/helm-install-release.md
```diff
@@ -23,18 +21,63 @@ Supporting Helm installations has the following requirements:
 
 * <FeatureFlag/>  
 
-## Package the Helm Chart
+## Package a Helm Chart and Create a Release {#release}
```
paigecalvert, line 24: Honestly, this topic was accomplishing much of the same goal as the "Packaging your helm chart" topic. It looks like the vendor portal links to this topic (and specifically the release anchor link), so I decided to keep this one over the packaging topic. 

A couple questions for you:
- I decided to keep this as a separate topic rather than incorporating it into the procedures in the Managing Releases with the Vendor Portal or CLI topics because it feels like it actually comes _before_ you create the release, rather than being part of a a release itself. Also, there's of course the people who might create a release without Helm charts and don't need to see these steps. WDYT?
- Since this topic no longer has release steps, do you think the title/headings in the topic need to be updated at all?
- These steps are indeed repeated over in the Using the SDK topic, albeit in a much lighter weight format. I thought it might still be helpful to keep the Using the SDK topic around because the framing over there is more focused on the SDK and "how do I implement/use this thing". Also, I could see that topic potentially expanding in the future if there are other steps we need to add. But WDYT?

docs/vendor/helm-release.md
paigecalvert, line 1: This was the Creating a Release with Your Helm Chart for KOTS topic. It was a URL that's been around for awhile that kept getting repurposed. Revisiting this topic, it didn't look like it was providing any additional value compared to the Configuring the HelmChart Custom Resource topic. 

The only piece of info that was missing from elsewhere in the docs was that this topic explained the Select Helm Install method dialog that pops up when you add a Helm chart.

So, I added mention of this dialog to the Managing Releases with the Vendor Portal steps and then linked to the Configuring the HelmChart Custom Resource topic for more info.

Do you notice anything else that isn't captured elsewhere? Do you think this change make sense?

docs/vendor/releases-creating-cli.mdx
```diff
@@ -33,7 +33,11 @@ To create and promote a new release:
 
 1. (Helm Charts Only) If your release contains any Helm charts:
 
-   1. Package the Helm charts into `.tgz` files. See [Creating a Helm Chart Package](helm-release-creating-package).
+   1. Package the Helm charts into `.tgz` files:
+
+      ```
+      helm package . -u
+      ```
```
paigecalvert, line 40: It's such a small step, might as well add it here

docs/vendor/releases-creating-releases.mdx
```diff
@@ -10,9 +10,7 @@ This topic describes how to use the Replicated vendor portal to create and promo
 
 ## Create a Release
 
-You can use standard manifest files or Helm charts to create a release. 
-
-If you are using Helm charts, make sure that you first create the Helm chart package. For more information, see [Creating a Helm Chart Package](helm-release-creating-package).
+You can create a release with one or more Helm charts, Kubernetes manifests, or a combination. You must package a Helm chart to a `.tgz` file to add it to a release. For more information, see [Create a Release with Your Helm Chart](helm-install-release).
```
jonquil2002, line 13: I'm not clear what "a combination" means here. Can we really create a release using Helm charts and standard manifests? Or does this mean a release can support Helm and KOTS installations for a Helm chart?

docs/vendor/releases-creating-releases.mdx
jonquil2002: I think this works. But in the topic it links to, which wasn't part of these edits, I happened to notice that the Overview section is really more of a task: https://deploy-preview-1309--replicated-docs.netlify.app/vendor/helm-native-v2-using

I suggest remaining Overview to Configure the HelmChart Custom Resource so that vendors quickly realize that there's a workflow in them thar bullets and then the rest of the sections in the topic make sense at a high level.

docs/vendor/releases-creating-releases.mdx
jonquil2002: Eh, no biggie - we're humans juggling a lot of changes. And I like the changes you made.

docs/vendor/helm-install-release.md
jonquil2002: I think keeping this topic separate from the Managing Releases release topics is a good idea.

Yep, the title of this one should be changed as it's no onger about creating a release.

One other point of confusion: this topic and the [Using the SDK with Your Application (Beta)](https://deploy-preview-1309--replicated-docs.netlify.app/vendor/replicated-sdk-using) topic are now basically redundant. I think one of them can be deleted, though you'd need to check and make sure that all is accounted and present.

I might suggest this order in the TOC:

**Distributing with Helm (Beta)**
  - About Distributing with Helm (Beta)
  - About the Replicated SDK (Beta)
  - Create a Release with Helm Charts (Beta)(combined with SDK info to delete the redundant topic)
  - Installing an Application with Helm (Beta)
  
  Followed by the existing section for Distributing Helm Charts with KOTS

docs/vendor/helm-release.md
jonquil2002: I think this makes sense. I don't see anything missing. The only thing is, as I mentioned in a different comment, I would rename the Overview section to Configure the HelmChart Custom Resource to match what the bulleted list is conveying. I think that's the overriding goal of this section.

Good condensing!!!

docs/vendor/releases-creating-cli.mdx
jonquil2002: Agreed. Why is it `-u` here, but elsewhere there is no flag ? And I think it's `-update` in one place too.

jonquil2002: I think if a flag is needed, it might help to describe what the flag is doing so they don't have to go look it up.

But good idea. :)

docs/vendor/helm-install-release.md
paigecalvert: @jonquil2002 here's what I was thinking on in terms of keeping them both:

>These steps are indeed repeated over in the Using the SDK topic, albeit in a much lighter weight format. I thought it might still be helpful to keep the Using the SDK topic around because the framing over there is more focused on the SDK and "how do I implement/use this thing". Also, I could see that topic potentially expanding in the future if there are other steps we need to add. But WDYT?

docs/vendor/releases-creating-releases.mdx
paigecalvert: Yeah, the idea is that the release has Helm charts and standard manifests. For example, a Helm chart, a HelmChart custom resource, a Config custom resource, etc. The combo would be required to support KOTS installations for Helm charts. (Let me know if this makes sense or if you see a better way to word it)

docs/vendor/releases-creating-releases.mdx
paigecalvert: Good call, people always skip over an Overview

docs/vendor/releases-creating-releases.mdx
paigecalvert: Looking at it again, it needs to be clarified that the applies to KOTS only

docs/vendor/releases-creating-cli.mdx
paigecalvert: Clarified that it's update dependencies (it's shorthand for this to combine the step of updating dependencies and packaging into one), and then also added a link to the Helm package docs

docs/vendor/releases-creating-releases.mdx
```diff
@@ -10,9 +10,11 @@ This topic describes how to use the Replicated vendor portal to create and promo
 
 ## Create a Release
 
-You can use standard manifest files or Helm charts to create a release. 
+You can create a release with one or more Helm charts. Vendors that support Replicated KOTS installations can also create releases that contain standard Kubernetes manifests.
```
paigecalvert, line 13: ^ wdyt of the edit @jonquil2002 ? Does this help clarify at all that "or a combination" language?

jonquil2002 (dismissed):
lgtm

docs/vendor/releases-creating-releases.mdx
jonquil2002: This might just be the long hours, but it's still not clear to me in terms of "or a combination". I believe that "combination" means that a single KOTS release can support Helm chart installation with KOTS and installation with the Helm CLI. 

Or are we supporting Helm charts and standard manifests in a single KOTS release? If this is so, probably needs to be stated similarly here, and still clarified further under the numbered step.

docs/vendor/releases-creating-releases.mdx
```diff
@@ -24,11 +23,28 @@ To create and promote a release in the vendor portal:
 
   [View a larger image](/images/release-create-new.png)
 
-1. Drag and drop your application files to the file directory in the YAML editor.
+1. If a drop down is displayed, select one of the following options:
+   * **Create Helm-only release**: Create a release with one or more Helm charts. Supports installations with the Helm CLI only.
+   * **Create KOTS release**: Create a release with one or more Helm charts, standard Kubernetes manifests, or a combination. Supports installations with Replicated KOTS. KOTS releases that contain Helm charts also support installation with Helm.
```
jonquil2002, line 28: Maybe reword this to something like: "Create a release with one or more Helm charts or standard Kubernetes manifests. Supports installations with Replicated KOTS. For Helm charts, KOTS releases support installing with KOTS and installing with Helm in the same release."

jonquil2002 (approved):
lgtm. Still confused about "combination" and I've added some suggestions. But overall it looks great!

docs/vendor/releases-creating-releases.mdx
paigecalvert: I like it, thank you

docs/vendor/releases-creating-releases.mdx
paigecalvert: I'm just trying to edit this sentence:

>You can use standard manifest files or Helm charts to create a release.

Because it is _not_ true that Builders users can create a release with standard manifests. So I want to say: Everyone can use Helm charts in a release. KOTS users can also add some standard manifests in there, or not use Helm charts at all

PR #1331, opened 2023-07-31
RBAC for instance notifications

divolgin:


jonquil2002 (approved):
lgtm. Added "the" in two places and reordered the `team/` list alphabetically.

PR #1339, opened 2023-08-01
83756: Clarify affix field settings

jonquil2002:
SC: https://app.shortcut.com/replicated/story/83756/affix-parameter-in-kots-config-needs-to-be-set-on-all-objects-in-a-group-otherwise-it-doesn-t-get-applied

Previews:
- https://deploy-preview-1339--replicated-docs.netlify.app/reference/custom-resource-config#affix
- https://deploy-preview-1339--replicated-docs.netlify.app/vendor/admin-console-customize-config-screen#add-fields-to-the-configuration-screen

### DIFF
--- a/docs/reference/custom-resource-config.md
+++ b/docs/reference/custom-resource-config.md
@@ -184,7 +184,7 @@ Items have a `name`, `title`, `type`, and other optional properties.
   <tr>
     <th>Description</th>
     <td>
-      <p>Items can be affixed <code>left</code> or <code>right</code>. Affixing items allows them to appear in the admin console on the same line.<br></br><br></br>Specify the <code>affix</code> field to all of the items in a particular group to prevent the appearance of crowded text. </p>
+      <p>Items can be affixed <code>left</code> or <code>right</code>. Affixing items allows them to appear in the admin console on the same line.<br></br><br></br>Specify the <code>affix</code> field to all of the items in a particular group to preserve the line spacing and prevent the appearance of crowded text. </p>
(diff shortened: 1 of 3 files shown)


docs/reference/custom-resource-config.md
```diff
@@ -184,7 +184,7 @@ Items have a `name`, `title`, `type`, and other optional properties.
   <tr>
     <th>Description</th>
     <td>
-      <p>Items can be affixed <code>left</code> or <code>right</code>. Affixing items allows them to appear in the admin console on the same line.</p>
+      <p>Items can be affixed <code>left</code> or <code>right</code>. Affixing items allows them to appear in the admin console on the same line.<br></br><br></br>Specify the <code>affix</code> field to all of the items in a particular group to preserve the line spacing and prevent the appearance of crowded text. </p>
```
paigecalvert, line 187: Should use `<p>` tags here instead of `<br>` to create line breaks ^

The benefit of using a `<p>` tag is that it can be more easily controlled with CSS. For example, we could target all p tags to edit the amount of spacing that goes between each p. It's also better semantic HTML for screen readers to use `<p>` rather than `<br>`

So:
```
<p>Items can be affixed <code>left</code> or <code>right</code>. Affixing items allows them to appear in the admin console on the same line.</p><p>Specify the <code>affix</code> field to all of the items in a particular group to preserve the line spacing and prevent the appearance of crowded text. </p>
```

paigecalvert (approved):

PR #1350, opened 2023-08-02
Update replicated-sdk-overview.md

dexhorthy:
needs review from an SME, probably @marccampbell or @AmberAlston to ensure this is indeed the right direction we want to take things

### DIFF
--- a/docs/vendor/distributing-workflow.md
+++ b/docs/vendor/distributing-workflow.md
@@ -9,12 +9,10 @@ If you are new to Replicated, complete the following prerequisites before you ge
 * Complete a basic Replicated onboarding workflow to create an application and then promote and install initial releases in a development environment: 
   * (Recommended) For Helm chart-based applications, see [Replicated Quick Start](/vendor/replicated-onboarding).
   
-    :::note
     Distributing your application as a Helm chart is recommended because you can support both installations with the Helm CLI and with KOTS from the same release, without having to maintain separate sets of Helm charts or application manifests. For more information, see [Helm Charts with KOTS](/vendor/distributing-overview#helm-charts-with-kots) in _About Distributing Applications with Replicated_.
-    :::
 
   * Alternatively, if you do _not_ intend to distribute a Helm chart-based application with Replicated, see [KOTS Tutorial (UI)](tutorial-ui-setup) for a workflow that uses a sample application with standard Kubernetes manifests.
-* Review the [Features Checklist](/vendor/replicated-onboarding#features-checklist) in _Replicated Quick Start_ for a list of features to integrate with your application to fully onboard onto the Replicated platform. You can integrate these platform features before, during, or after onboarding with KOTS.
+* Review the [Features Checklist](/vendor/replicated-onboarding#features-checklist) in _Replicated Quick Start_ for a list of features to integrate with your application to fully onboard onto the Replicated platform. You can integrate these platform features before, during, or after onboarding with KOTS.   
 
 ## Custom Resources Checklist
 
@@ -33,7 +31,12 @@ For more information about creating releases, see [Managing Releases with the Ve
   <tr>
     <td>HelmChart</td>
     <td><p>Provides instructions for KOTS about how to deploy your Helm chart.</p><p><strong>Note:</strong> Required for supporting KOTS installations of Helm charts.</p></td>
-    <td><a href="helm-native-v2-using">Configuring the HelmChart Custom Resource</a></td>
+    <td>
+      <ul>
+        <li><a href="helm-native-v2-using">Configuring the HelmChart Custom Resource</a></li>
+        <li><a href="helm-kots-using-sdk">Using an SDK-Enabled Helm Chart for KOTS Installations</a></li>
+      </ul>  
+    </td>
   </tr>
   <tr>
     <td>Preflight and SupportBundle</td>
--- a/docs/vendor/distributing-workflow.md
+++ b/docs/vendor/distributing-workflow.md
@@ -12,7 +12,7 @@ If you are new to Replicated, complete the following prerequisites before you ge
     Distributing your application as a Helm chart is recommended because you can support both installations with the Helm CLI and with KOTS from the same release, without having to maintain separate sets of Helm charts or application manifests. For more information, see [Helm Charts with KOTS](/vendor/distributing-overview#helm-charts-with-kots) in _About Distributing Applications with Replicated_.
 
   * Alternatively, if you do _not_ intend to distribute a Helm chart-based application with Replicated, see [KOTS Tutorial (UI)](tutorial-ui-setup) for a workflow that uses a sample application with standard Kubernetes manifests.
-* Review the [Features Checklist](/vendor/replicated-onboarding#features-checklist) in _Replicated Quick Start_ for a list of features to integrate with your application to fully onboard onto the Replicated platform. You can integrate these platform features before, during, or after onboarding with KOTS.   
+* Review the [Features Checklist](/vendor/replicated-onboarding#features-checklist) in _Replicated Quick Start_ for a list of features to integrate with your application to fully onboard onto the Replicated platform. You can integrate these platform features before, during, or after onboarding with KOTS.
 
 ## Custom Resources Checklist
 
--- a/docs/vendor/helm-kots-using-sdk.md
+++ b/docs/vendor/helm-kots-using-sdk.md
@@ -1,0 +1,76 @@
+# Using an SDK-Enabled Chart for KOTS Installations
+
+This topic provides information about how to use Helm charts that include the Replicated SDK to support installations with Replicated KOTS. For more information about the SDK, see [About the Replicated SDK](replicated-sdk-overview).
+
+## Overview
+
+Distributing your application as one or more Helm charts is recommended because you can use the same Helm charts to support installations with both KOTS and with the Helm CLI.
+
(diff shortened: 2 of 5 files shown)


paigecalvert:
Thanks @dexhorthy . I added this story to our backlog for documenting this: https://app.shortcut.com/replicated/story/84712

paigecalvert:
Also, pinged Amber and Marc here to give them a heads up that their input is requested: https://replicated.slack.com/archives/C02BPJQQP28/p1691072503053779 🙂

docs/vendor/replicated-sdk-overview.md
```diff
@@ -123,4 +123,15 @@ For more information about viewing customer and instance insights in the vendor
 
 You can run the Replicated SDK in integration mode to more quickly test new functionality for your application. Integration mode allows you to use mock data to test changes locally without having to create a release in the vendor portal and then install your Helm chart from the Replicated registry.
 
-For more information, see [Developing Against the SDK API (Beta)](replicated-sdk-development).
\ No newline at end of file
+For more information, see [Developing Against the SDK API (Beta)](replicated-sdk-development).
+
+## Using an SDK-enabled Chart in a KOTS Release
+
+Many vendors will find it convenient to use the same Helm Chart for KOTS releases that they use for Helm CLI releases, in some cases even exposing the same Replicated release to both sets of customers. However, running both KOTS and the SDK can cause duplication of instance data in the Vendor Portal and related APIs. To avoid this, you will need to create a single Helm Chart that contains the Replicated SDK, but with a conditional value to exclude the SDK when the chart is being deployed by KOTS. To do this, you will need to make three changes to your application.
```
marccampbell, line 130: Why "single helm chart". I think this could be more accurately described as:

To avoid this, any of your Helm Charts that contain the Replicated SDK as a depedency or subchart will need to contain a conditional value to exclude the SDK...

dexhorthy, line 130:
```suggestion
Many vendors will find it convenient to use the same Helm Chart for KOTS releases that they use for Helm CLI releases, in some cases even exposing the same Replicated release to both sets of customers. However, running both KOTS and the SDK can cause duplication of instance data in the Vendor Portal and related APIs. To avoid this, any of your Helm Charts that contain the Replicated SDK will need a conditional value to exclude the SDK when the chart is being deployed by KOTS. To do this, you will need to make three changes to your application.
```

docs/vendor/distributing-workflow.md
```diff
@@ -33,7 +31,12 @@ For more information about creating releases, see [Managing Releases with the Ve
   <tr>
     <td>HelmChart</td>
     <td><p>Provides instructions for KOTS about how to deploy your Helm chart.</p><p><strong>Note:</strong> Required for supporting KOTS installations of Helm charts.</p></td>
-    <td><a href="helm-native-v2-using">Configuring the HelmChart Custom Resource</a></td>
+    <td>
+      <ul>
+        <li><a href="helm-native-v2-using">Configuring the HelmChart Custom Resource</a></li>
+        <li><a href="helm-kots-using-sdk">Using an SDK-Enabled Helm Chart for KOTS Installations</a></li>
```
paigecalvert, line 37: Add new topic ^

docs/vendor/helm-kots-using-sdk.md
```diff
@@ -0,0 +1,76 @@
+# Using an SDK-Enabled Chart for KOTS Installations
```
paigecalvert, line 1: It felt like there was enough here to expand on this in it's own topic. I put it over in the "Helm Charts with KOTS" section of the KOTS sidebar, which is where we have other info about configuring the HelmChart custom resource.

I added links to this topic in a few places

docs/vendor/replicated-sdk-overview.md
```diff
@@ -123,4 +123,10 @@ For more information about viewing customer and instance insights in the vendor
 
 You can run the Replicated SDK in integration mode to more quickly test new functionality for your application. Integration mode allows you to use mock data to test changes locally without having to create a release in the vendor portal and then install your Helm chart from the Replicated registry.
 
-For more information, see [Developing Against the SDK API (Beta)](replicated-sdk-development).
\ No newline at end of file
+For more information, see [Developing Against the SDK API (Beta)](replicated-sdk-development).
+
+## SDK-Enabled Helm Charts with KOTS
```
paigecalvert, line 128: I could also see this being called "Limitations", but as we only have one limitation right now, I think this heading does make the info easier to find. I could see people wondering "how's this thing work with kots?"

docs/vendor/helm-kots-using-sdk.md
```diff
@@ -0,0 +1,76 @@
...
+
+   **Example:**
+
+    ```yaml
+    # Helm chart values.yaml
+    
+    replicated:
+      # Set `enabled: true` to include the SDK by default
+      enabled: true
+    ```
```
paigecalvert, line 47: I thought it would be more descriptive to call the field `replictaed-sdk` rather than just `replicated`, but since the registry automatically injects a `replicated-sdk` field into the values file during install, I wasn't sure if that would cause issues. So, left as `replicated`

docs/vendor/distributing-workflow.md
```diff
@@ -9,9 +9,7 @@ If you are new to Replicated, complete the following prerequisites before you ge
 * Complete a basic Replicated onboarding workflow to create an application and then promote and install initial releases in a development environment: 
   * (Recommended) For Helm chart-based applications, see [Replicated Quick Start](/vendor/replicated-onboarding).
   
-    :::note
     Distributing your application as a Helm chart is recommended because you can support both installations with the Helm CLI and with KOTS from the same release, without having to maintain separate sets of Helm charts or application manifests. For more information, see [Helm Charts with KOTS](/vendor/distributing-overview#helm-charts-with-kots) in _About Distributing Applications with Replicated_.
-    :::
```
paigecalvert, line 13: ^ unrelated change, just found the note styling to be superfluous

docs/vendor/helm-kots-using-sdk.md
```diff
@@ -0,0 +1,76 @@
...
+
+If you include the Replicated SDK as a dependency in a Helm chart that you want to use for both KOTS and Helm CLI installations, then you must add a conditional value to exclude the SDK when the chart is installed by KOTS. Excluding the SDK for KOTS installations prevents duplication of instance data in the vendor portal and related APIs, which can be caused when both KOTS and the SDK are reporting data for an instance.
+
+## Limitation
+
+When the SDK is excluded from KOTS installations, the SDK API endpoints are not available. For example, you cannot use the `app` endpoints to allow your users to check for updates when the SDK is excluded. And, you cannot use the `license` endpoints to check customer entitlements during runtime.
+
+However, KOTS also provides access to this functionality that is available through the SDK API. For example, users can check for updates through the admin console, and vendors can check license entitlements at runtime using the admin console API. If you use any SDK API endpoints in your application, Replicated recommends that you verify that the comparable functionality available through KOTS provides the expected user experience.
```
dexhorthy, line 15: > vendors can check license entitlements at runtime using the admin console API

 this might not be as true as I thought it was

docs/vendor/helm-kots-using-sdk.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Overview
+
+Distributing your application as one or more Helm charts is recommended because you can use the same Helm charts to support installations with both KOTS and with the Helm CLI.
+
+If you include the Replicated SDK as a dependency in a Helm chart that you want to use for both KOTS and Helm CLI installations, then you must add a conditional value to exclude the SDK when the chart is installed by KOTS. Excluding the SDK for KOTS installations prevents duplication of instance data in the vendor portal and related APIs, which can be caused when both KOTS and the SDK are reporting data for an instance.
+
+## Limitation
+
+When the SDK is excluded from KOTS installations, the SDK API endpoints are not available. For example, you cannot use the `app` endpoints to allow your users to check for updates when the SDK is excluded. And, you cannot use the `license` endpoints to check customer entitlements during runtime.
```
dexhorthy, line 13: we could add a note that we have on our roadmap (no guarantee if we'll do it or when) that we'd like to consolidate these APIs so that the same endpoints are available in KOTS as well.

@marccampbell @AmberAlston another ping on this question because IDK how publicly we want to discuss that, or if it's still the intended direction

jonquil2002 (dismissed):
lgtm

docs/vendor/helm-kots-using-sdk.md
AmberAlston: We don't mention "admin console api" anywhere in our docs today because the kots api is only for kots internal use. We don't currently document or support that api for external use cases.

docs/vendor/helm-kots-using-sdk.md
AmberAlston: will look at updated content again, but we shouldn't mention anything that is not true today. We'd previously purposely stripped out any "future promises" from the docs. It's not a best practice, and there were several examples of poorly aged unfulfilled promises in the old kots docs. If and when we start to make this change, then we will update the docs accordingly.

dexhorthy:
lgtm

docs/vendor/helm-kots-using-sdk.md
AmberAlston: That documented ability to query the entitlements is really our one exception to that. I wouldn't remove the content, but we may want to slightly tweak the language on that page a bit so it doesn't as strongly imply a supported kots api. More like "send an HTTP request to the admin console {license} endpoint at..."

paigecalvert: Okay, that makes sense. I'll rewrite this limitation accordingly as well

paigecalvert: Updated the language in this section to link to the related KOTS docs and remove mention of the admin console API

paigecalvert:
Going to merge this in. Here's the follow up story for updating the language in the Checking Entitlements for KOTS topic per the discussion above about the admin console API: https://app.shortcut.com/replicated/story/85005

PR #1352, opened 2023-08-03
HOLD FOR RELEASE - Document replicated instance types for VMs

marccampbell:
Moving VM based Compatibility Matrix instances from VCPUs and Memory to instance types.


jonquil2002 (approved):
lgtm - added an xref to the CLI table linking to the instance types topic.

PR #1362, opened 2023-08-07
add kots version for regex validation

cbodonnell:
I think the release notes is the only place that mentions when support for regex validation was added to KOTS.  If a user is starting from this page, they wouldn't likely come across that information and will think it just doesn't work.

### DIFF
--- a/docs/reference/custom-resource-config.md
+++ b/docs/reference/custom-resource-config.md
@@ -638,7 +638,7 @@ The validation rules are as follows:
 - If a group `when` condition is set to `false`, the items in the group are not validated.
 
 ### `regex`
-A `regex` can be used to validate whether an item's value matches the provided regular expression `pattern`. The regex pattern should be of the [RE2 regular expression](https://github.com/google/re2/wiki/Syntax) type and can validate the `text`, `textarea`, `password`, and `file` field types.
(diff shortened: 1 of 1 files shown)


docs/reference/custom-resource-config.md
```diff
@@ -638,7 +638,7 @@ The validation rules are as follows:
 - If a group `when` condition is set to `false`, the items in the group are not validated.
 
 ### `regex`
-A `regex` can be used to validate whether an item's value matches the provided regular expression `pattern`. The regex pattern should be of the [RE2 regular expression](https://github.com/google/re2/wiki/Syntax) type and can validate the `text`, `textarea`, `password`, and `file` field types.
+For applications installed with KOTS v1.98.0 or later, a `regex` can be used to validate whether an item's value matches the provided regular expression `pattern`. The regex pattern should be of the [RE2 regular expression](https://github.com/google/re2/wiki/Syntax) type and can validate the `text`, `textarea`, `password`, and `file` field types.
```
cbodonnell, line 641: This might be better on line 629, but it's possible that we add other validation types in future KOTS versions, so I added it to the `regex` section

paigecalvert: I think nesting it under the regex makes good sense 👍

paigecalvert (approved):

PR #1364, opened 2023-08-08
Update docs for injected global.replicated Helm values

sgalsaleh:


paigecalvert (dismissed):
thanks @sgalsaleh ! just edited the example name for the customer because I don't think we actually mention the current plan names anywhere in the docs

paigecalvert (approved):

PR #1372, opened 2023-08-09
Update KOTS ref in sidebar

paigecalvert:
https://deploy-preview-1372--replicated-docs.netlify.app/


jonquil2002 (approved):
lgtm. Can the icon be made static rather than show just on hover? I think that would be helpfuly

paigecalvert:
@jonquil2002 the icon does not just show on hover, it's always there. Was it looking like that for you in the preview?

jonquil2002:
@paigecalvert No, it's ok now. I was having a lot of strange issues with my computer yesterday afternoon. Rebooting didn't help. The hard shutdown fixed it. The icon looks good.

PR #1399, opened 2023-08-25
add new section for managing apps

paigecalvert:
New topic that I added to a new "Applications" section in the sidebar: https://deploy-preview-1399--replicated-docs-upgrade.netlify.app/vendor/vendor-portal-manage-app. Previously, I'm not sure where I would have thought to look for info about where to find my app slug. 

and also lots of xrefs back to the section in this topic about how to get the app slug


xavpaice (approved):
LGTM, thanks!

PR #1404, opened 2023-08-29
Clarify lack of migration path for replication helm

paigecalvert:


AmberAlston:
Nitpick - Should probably be an uppercase N (from About Distributing Helm Chart) 
image 

Changes on About Distributing Helm Chart look good. 

I still have some clarity concerns on vendor/helm-native-v2-using#migrating. Because both Replicated Helm and Native Helm use the `kots.io/v1beta1` naming, it can easily read to me that I can actually upgrade from either one given we are only saying kots.io/v1beta1 in the migration content and don't explicitly call out `Migration from useHelmInstall: false Not Supported` until the end.

PR #1427, opened 2023-09-07
Add conditional config option examples

paigecalvert:
https://deploy-preview-1427--replicated-docs.netlify.app/vendor/config-screen-conditional


docs/vendor/config-screen-conditional.md
```diff
@@ -0,0 +1,307 @@
...
+kind: Config
+metadata:
+  name: config-sample
+spec:
+  groups:  
+  - name: ingress_settings
+    title: Ingress Settings
+    description: Configure Ingress for Determined
+    items:
+    - name: determined_ingress_type
```
diamonwiggins, line 236: Should we make this generic as opposed to keeping the `determined` naming here? If we have other places in the docs were we make references to open source apps in examples, then its probably fine? I'm good either way, just figured worth asking.

Regardless of the decision we make, do you think it would make sense to provide a link to our full examples on this page somewhere? Or maybe a different page for that?

https://github.com/replicatedhq/platform-examples

diamonwiggins (approved):
lgtm

docs/vendor/config-screen-conditional.md
paigecalvert: Good catch, thank you! I intended to delete the word "determined".

I also think it's a great idea to add a link to the platform-examples repo. There will probably end up being a few places in the docs where I should link to that, but this will be a good start

PR #1431, opened 2023-09-11
Update testing-how-to.md

xavpaice:
Adds some known limitations to CMX


xavpaice:
https://app.shortcut.com/replicated/story/88655/update-known-limitations-for-cmx

paigecalvert (approved):
thank you @xavpaice ! 

I moved the openshift-specific limitations to the table in the Supported Clusters topic, because it looks like that's where the distribution-specific limitations live. I also added a link from the main Limitations section to this topic to indicate to people that there are more distribution-specific limitations that they might want to know about

Issue #1461, opened 2023-09-22
Docs feedback on Installing


crdant:
URL: https://docs.replicated.com/reference/replicated-cli-installing
Feedback details:

Happy to provide a PR later, but working with a customer right now and realized it wasn't clear to them from the docs that they could use `replicated login` vs. `export REPLICATED_API_TOKEN=$TOKEN`

paigecalvert:
Thanks @crdant , I've ever seen the `replicated login` command before. Looks like we are missing both `login` and `logout` in the replicated CLI docs, as well as a couple others. I just sent a request to the vendor portal team to ask that they open a PR to add these into the replicated CLI docs: https://replicated.slack.com/archives/CEEFDQURM/p1695394971679939

And I can update this topic accordingly!

paigecalvert:
Closing as the PR is merged. thanks!

PR #1463, opened 2023-09-22
Add info about replicated login

paigecalvert:
Preview: https://deploy-preview-1463--replicated-docs.netlify.app/reference/replicated-cli-installing

For issue https://github.com/replicatedhq/replicated-docs/issues/1461


docs/partials/replicated-cli/_sudo-install.mdx
```diff
@@ -0,0 +1,3 @@
+:::note
+If you do not have root access to the `/usr/local/bin` directory, you can install with sudo by running `sudo mv replicated /usr/local/bin/replicated` instead of `mv replicated /usr/local/bin/replicated`.
+:::
```
paigecalvert, line 3: I noticed that I wasn't able to install unless I used sudo, so figured this would be helpful

docs/reference/replicated-cli-installing.md
```diff
@@ -98,6 +94,8 @@ To run the latest replicated CLI:
         replicated/vendor-cli --help
       ```
 
+      <!-- CHECK ACCURACY OF THESE COMMANDS -->
```
paigecalvert, line 97: I am suspicious of these docker/windows commands. The docs currently tell people to run one of these:

- Through a docker container:

      ```
      docker run \
        -e REPLICATED_APP=$APP_NAME \
        -e REPLICATED_API_TOKEN=$TOKEN \
        replicated/vendor-cli --help
      ```

 - On Windows:

      ```
      docker.exe run \
        -e REPLICATED_APP=%APP_NAME% \
        -e REPLICATED_API_TOKEN=%TOKEN% \
        replicated/vendor-cli --help
        ```

But, it seems like instead it should be something like:

```
docker run \
   replicated login --app SLUG --token TOKEN
```

docs/reference/replicated-cli-installing.md
```diff
@@ -107,4 +105,39 @@ To run the latest replicated CLI:
         replicated/vendor-cli --help
       ```
 
-  <EnvVariablesReplace/>
\ No newline at end of file
+## Authenticate with an Application Slug and API Token {#auth}
```
paigecalvert, line 108: Moved these steps to a new subheading since they don't really have to do with installing the CLI. It's usage

docs/reference/replicated-cli-installing.md
```diff
@@ -107,4 +105,39 @@ To run the latest replicated CLI:
...
+1. In the [vendor portal](https://vendor.replicated.com), go to the **Application Settings** page and copy the slug for the target application. See [Get the Application Slug](/vendor/vendor-portal-manage-app#slug) in _Managing Application_.
+
+1. Do _one_ of the following to authenticate with the replicated CLI:
+
+   * Authenticate with the `replicated login` command:
+
+     1. Run the following command:
+
+      ```
+      replicated login --app APP_SLUG --token TOKEN
```
grantmiller, line 125: @marccampbell you don't need a toke or an app slug for this command, right? (it creates a token for you and copies it to the right location or something)

@paigecalvert we should probably also make sure we document `replicated logout`

crdant: Not @marccampbel but I know the answer. You don't need a token when you use `replicated login`. It grabs an authentication token and creates a config file containing it. If that file is there and the variable isn't set, then it uses that token.

docs/reference/replicated-cli-installing.md
crdant: Either of those would work. The version in the docs is telling docker to set up the environment with the variables set. The option you suggest is telling the command to use a specific app and token without looking at its environment for them.

crdant (changes requested):
Update to show that the API token and associated setup isn't required when using `replicated login`.

docs/reference/replicated-cli-installing.md
crdant: In the rendered version these commands seem to be cut off

image

docs/reference/replicated-cli-installing.md
```diff
@@ -75,36 +74,123 @@ To install latest replicated CLI on Linux:
...
+
+    ```dos
+    docker.exe run \
+      -e REPLICATED_API_TOKEN=%TOKEN% \
+      replicated/vendor-cli --help
+    ```
+
+    Replace `TOKEN` with your API token.
+
+  For more information about the `docker run` command, see [docker run](https://docs.docker.com/engine/reference/commandline/run/) in the Docker documentation.
```
paigecalvert, line 120: Added link to docker docs for more info

docs/reference/replicated-cli-installing.md
```diff
@@ -75,36 +74,123 @@ To install latest replicated CLI on Linux:
...
-    - Through a docker container:
+  Download and install the files. For simplicity, the usage in the next step is represented assuming that the CLI is downloaded and installed to the desktop.
 
-      ```shell
-      docker run \
-        -e REPLICATED_APP=$APP_NAME \
-        -e REPLICATED_API_TOKEN=$TOKEN \
-        replicated/vendor-cli --help
-      ```
+1. Authorize the replicated CLI:
```
paigecalvert, line 99: Updated these commands so that they use the token env var, and do not include the app env var

PR #1493, opened 2023-10-05
[HOLD FOR SDK RELEASE] add section for customizing environment variables

cbodonnell:
This PR adds supporting documentation for the feature added in https://github.com/replicatedhq/replicated-sdk/pull/96.

### DIFF
--- a/docs/vendor/install-with-helm.md
+++ b/docs/vendor/install-with-helm.md
@@ -42,7 +42,7 @@ To install a Helm chart:
     :::
 
     :::note
-    To install the SDK with custom RBAC permissions, include the `--set` flag with the `helm install` command to override the value of the `replicated.serviceAccountName` field with a custom service account. For more information, see [Customizing RBAC for the SDK](/vendor/replicated-sdk-rbac).
+    To install the SDK with custom RBAC permissions, include the `--set` flag with the `helm install` command to override the value of the `replicated.serviceAccountName` field with a custom service account. For more information, see [Customizing RBAC for the SDK](/vendor/replicated-sdk-customizing#customizing-rbac-for-the-sdk).
     :::
 
 1. (Optional) In the vendor portal, click **Customers**. You can see that the customer you used to install is marked as **Active** and the details about the application instance are listed under the customer name. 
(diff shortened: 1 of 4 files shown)


paigecalvert (approved):

docs/vendor/install-with-helm.md
```diff
@@ -42,7 +42,7 @@ To install a Helm chart:
     :::
 
     :::note
-    To install the SDK with custom RBAC permissions, include the `--set` flag with the `helm install` command to override the value of the `replicated.serviceAccountName` field with a custom service account. For more information, see [Customizing RBAC for the SDK](/vendor/replicated-sdk-rbac).
+    To install the SDK with custom RBAC permissions, include the `--set` flag with the `helm install` command to override the value of the `replicated.serviceAccountName` field with a custom service account. For more information, see [Customizing RBAC for the SDK](/vendor/replicated-sdk-customizing#customizing-rbac-for-the-sdk).
```
ajp-io, line 45:
```suggestion
    To install the SDK with custom RBAC permissions, include the `--set` flag with the `helm install` command to override the value of the `replicated.serviceAccountName` field with a custom service account. For more information, see [Customizing RBAC for the SDK](/vendor/replicated-sdk-customizing#customize-rbac-for-the-sdk).
```

docs/vendor/replicated-sdk-customizing.md
```diff
@@ -65,4 +69,32 @@ To install the SDK with custom RBAC:
   helm install wordpress oci://registry.replicated.com/my-app/beta/wordpress --set replicated.serviceAccountName=mycustomserviceaccount
   ```
 
- For more information about installing with Helm, see [Installing with Helm](/vendor/install-with-helm).  
\ No newline at end of file
+ For more information about installing with Helm, see [Installing with Helm](/vendor/install-with-helm).  
+
+## Add Environment Variables
+
+You can add additional environment variables to the Replicated SDK deployment. For example, if the SDK is running behind an HTTP proxy server, then the user could set `HTTP_PROXY` or `HTTPS_PROXY` environment variables to provide the hostname or IP address of the proxy server.
```
ajp-io, line 76: I wonder if it would be helpful to explain that this is helpful if an environment variable isn't exposed via a Helm value. So you can set it directly this way in that case.

paigecalvert (approved):

PR #1515, opened 2023-10-10
Add customer-instance CSV data dictionary

FourSigma:
Data table for CSV / JSON export

PR #1517, opened 2023-10-11
Download -> Export

dexhorthy:
What do you think of this title change? I'm not sure what needs to happen to update the sidebar title


paigecalvert:
Made this change as part of Siva's PR to add the data dictionary https://github.com/replicatedhq/replicated-docs/pull/1515

PR #1553, opened 2023-10-25
Update ports field

paigecalvert:
https://deploy-preview-1553--replicated-docs.netlify.app/reference/custom-resource-application#ports


docs/partials/custom-resource-application/_ports-serviceName.mdx
```diff
@@ -0,0 +1 @@
+<li><code>ports.serviceName</code>: The name of the service that receives the traffic. KOTS can create a port forward to ClusterIP, NodePort, or LoadBalancer services. For embedded clusters provisioned by Replicated kURL, NodePort type services are recommended because the service is exposed on the host where it is accessible to users. For more information about Kubernetes service types, see <a href="https://kubernetes.io/docs/concepts/services-networking/service/">Service</a> in the Kubernetes documentation.</li>
```
paigecalvert, line 1: This part is updated:
>KOTS can create a port forward to ClusterIP, NodePort, or LoadBalancer services. For embedded clusters provisioned by Replicated kURL, NodePort type services are recommended because the service is exposed on the host where it is accessible to users.

Want to make sure this is accurate and/or if we have any additional recommendations. I think this will need to be better expanded on (perhaps providing examples of each type of service) in a separate topic, but at least want to make sure this section correctly lists what's supported.

For context, I have a thread here where I was asking Andrew about what's supported for embedded clusters: https://replicated.slack.com/archives/C036BTS7JCE/p1698166699186899

docs/partials/custom-resource-application/_ports-servicePort.mdx
```diff
@@ -0,0 +1 @@
+<li><p><code>ports.servicePort</code>: The <code>containerPort</code> of the Pod where the service is running. This is the port where KOTS forwards traffic.</p></li>
```
paigecalvert, line 1: Updated this to say "The <code>containerPort</code> of the Pod where the service is running." Is that the right way to word it?

docs/reference/custom-resource-application.md
```diff
@@ -304,20 +309,19 @@ spec:
...
-      <p>You can use the <code>ports</code> field to create a port-forward to a service that has a <code>ClusterIP</code> type. For embedded clusters provisioned by Replicated kURL, you can also create a custom link to a service that has a <code>NodePort</code> type.</p>
-      <p>For more information about configuring a custom link in embedded clusters to a <code>NodePort</code> type service, see <a href="/vendor/admin-console-adding-buttons-links">Adding Buttons and Links</a>.</p>
-      <p><code>ports</code> has the following fields:</p>
+      <p>Extra ports (additional to the <code>:8800</code> admin console port) that are port-forwarded when running the <code>kots admin-console</code> command. With ports specified, KOTS can establish port forwarding to simplify connections to the deployed application. When the application starts and the service is ready, the kots CLI will print a message in the terminal with the URL where the port-forwarded service can be accessed.</p>
```
paigecalvert, line 315: Added this description for how the user accesses the port-forwarded svc:

>When the application starts and the service is ready, the kots CLI will print a message in the terminal with the URL where the port-forwarded service can be accessed.

docs/partials/custom-resource-application/_ports-applicationURL.mdx
```diff
@@ -0,0 +1 @@
+<li><p><code>ports.applicationUrl</code>: Must match a service found in the Kubernetes Application manifest.</p><p> For example, <code>ports.applicationUrl: "http://my-application-url"</code> in the KOTS Application custom resource must match <code>descriptor.links.url: https://my-application-url</code> in the Kubernetes Application manifest.</p></li>
```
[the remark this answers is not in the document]
cbodonnell: It's just used for dashboard links based on what I see in the code

docs/partials/custom-resource-application/_ports-serviceName.mdx
cbodonnell: My understanding of this is that we don't do any automatic port-forwarding in embedded clusters for the reasons Andrew mentioned in the community post. In general, these port-forwards are more relevant for existing cluster installs. In embedded clusters we create a NodePort service for the admin console on 8800, but if the vendor wants more ports exposed on the node, they would have to create additional NodePort services themselves as Andrew mentions in the follow-up comment.

I think a note like this about embedded clusters might be more relevant around this area where we talk these port-forwards in general https://github.com/replicatedhq/replicated-docs/pull/1553/files#diff-c5e9e1291693135346d30f42ff5574afc15e68d727986db8d9f46e7340fb9309R315

docs/partials/custom-resource-application/_ports-servicePort.mdx
cbodonnell: I think so given the callout below

cbodonnell (approved):

PR #1582, opened 2023-11-03
[HOLD FOR 1.104.0 RELEASE] note about KOTS support bundle spec discovery

cbodonnell:
Clarifying note based on questions in Slack https://replicated.slack.com/archives/C02TQTV9G2V/p1698956150308879.

Support was added to KOTS for the `troubleshoot.sh/kind` label in https://github.com/replicatedhq/kots/pull/4113


docs/partials/support-bundles/_generate-bundle.mdx
```diff
@@ -1,36 +1,45 @@
-Support bundles are generated from one or more support bundle specifications. Specification locations include YAML manifest files in the cluster, URLs, and Secrets or ConfigMap objects embedded in a Kubernetes cluster. For more information, see [Collecting a Support Bundle](https://troubleshoot.sh/docs/support-bundle/collecting/) and [Discover Cluster Specs](https://troubleshoot.sh/docs/support-bundle/discover-cluster-specs/) in the Troubleshoot documentation.
+Support bundles are generated by providing one or more support bundle specifications with the `kubectl support-bundle` command.
 
-The following examples show some possible CLI commands to generate a support bundle for one or more specifications and locations:
+For installations with the Helm CLI or with KOTS v1.104.0 and later, specifications can be automatically discovered in Secrets or ConfigMap objects in the cluster using the `--load-cluster-specs` flag with the `kubectl support-bundle`. Alternatively, for installations with the Helm CLI or with KOTS versions earlier than 1.104.0, specifications can be provided from YAML files in the local directory or from a URL.
```
paigecalvert, line 3: Clarify that `--load-cluster-specs` flag applies to Helm CLI installs and to KOTS v1.104.0 or later.

Also made some edits to this description overall because I found it a little difficult to follow before

docs/vendor/preflight-kots-defining.md
```diff
@@ -15,13 +15,9 @@ For information about defining host preflight checks for Replicated kURL, see [C
 
 ## Define the Preflight Checks Specification
```
paigecalvert, line 16: Just some minor copyediting for simplicity. Couldn't resist

docs/vendor/support-bundle-helm-customizing.md
```diff
@@ -9,9 +9,9 @@ import RedactorsAbout from "../partials/redactors/_redactors-about.mdx"
 import PreflightSbHelmTemplates from "../partials/preflights/_preflight-sb-helm-templates.mdx"
 
 
-# Customize Support Bundles for Helm Installations
+# Customize Support Bundles for Helm Charts
 
-This topic provides a basic understanding and some key considerations about support bundle specifications for Helm installations to help guide you in defining them for your application.
+This topic provides a basic understanding and some key considerations about support bundle specifications for Helm charts to help guide you in defining them for your application. The content in this topic applies to Helm chart-based applications installed with the Helm CLI or with Replicated KOTS v1.104.0 and later.
```
paigecalvert, line 14: > The content in this topic applies to Helm chart-based applications installed with the Helm CLI or with Replicated KOTS v1.104.0 and later.

Want to confirm this. I think this topic applies to everyone distributing helm chart-based apps as I'm pretty sure that we want to recommend people distributing Helm charts create their preflight specs within their Helm chart. 

The part I'm not as sure about is if we want to specify KOTS 1.104.0 or later. They can't use `--load-cluster-specs` with earlier versions, but are there other reasons why they would include the spec in their helm chart for kots installations?

docs/vendor/support-bundle-helm-customizing.md
```diff
@@ -23,14 +23,16 @@ Customizing a support bundle is unique to your application and depends on what k
...
+The support-bundle plugin requires a support bundle specification as input. You can define support bundle specifications in using the following kinds of resources:
+
+- Kubernetes Secret or ConfigMap with `label.troubleshoot.sh/kind: support-bundle`
+- SupportBundle custom resource (`apiVersion: troubleshoot.sh/v1beta2` and `kind: SupportBundle`)
 
 ### Create a Secret (Recommended)
 
-Replicated recommends using Secrets to contain a support bundle specifications in your Helm chart. This method allows customers to automatically discover and generate a support bundle without specifying a long URL. Using Secrets also allows specifications to be templated using information in the `values.yaml` file.
+Replicated recommends defining support bundle specifications in Secrets in your Helm chart templates. For installations with the Helm CLI or with KOTS v1.104.0 and later, defining specifications in Secrets allows your users to automatically discover and generate a support bundle without needing to provide multiple files or specify a long URL with the `kubectl support-bundle` command. Using Secrets also allows specifications to be templated using information in the `values.yaml` file.
```
paigecalvert, line 35: Calling out again the 1.104.0 and later requirement for loading cluster specs

docs/vendor/support-bundle-helm-customizing.md
```diff
@@ -65,13 +67,13 @@ To create a Secret for the support bundle specification:
...
     ```
 
-1. Add the Secret to your Helm chart `templates/` directory.  
+1. Add the Secret to your Helm chart `templates/` directory.
 
 Next, define the support bundle specification by adding collectors and analyzers. For more information, see [Define the Support Bundle Specification](#spec).
 
 ### Create a SupportBundle Custom Resource
 
-If you do not want to use Secrets, you can create a SupportBundle custom resource instead. Helm templates are supported when the specification is distributed using an OCI registry.
```
paigecalvert, line 74: >  Helm templates are supported when the specification is distributed using an OCI registry.

I didn't see anything in the output of kubectl support-bundle --help or in the troubleshoot docs that indicated you could provide a link to the OCI registry. I think this might have been accidentally conflated with the kubectl preflight option to point to an OCI registry

docs/vendor/support-bundle-helm-customizing.md
```diff
@@ -65,13 +67,13 @@ To create a Secret for the support bundle specification:
...
+If you do not want to use Secrets or ConfigMaps, you can create a SupportBundle custom resource instead.
 
 Create a SupportBundle custom resource (`kind: SupportBundle`) using the following basic support bundle template. For more information about this custom resource, see [Preflight and Support Bundle](/reference/custom-resource-preflight).
```
paigecalvert, line 78: ^ Is there any requirement that this spec is wrapped in a template so that it's deployed to the cluster? or does that only matter for the preflight specs? It's not really clear to me, if I'm trying to distribute with the Helm CLI, what I would do with this SupportBundle custom resource. Add it to my templates? Just save it in a YAML file and email it to my customer as needed?

docs/vendor/support-bundle-helm-customizing.md
ajp-io: The only reasons to include the spec this way is 1) you're also shipping with Helm CLI, or 2) it works, so you feel like doing it this way. But I think it's fine to mention the KOTS version so vendors know in which versions of KOTS their customers can get the support bundle from the Helm chart.

docs/vendor/support-bundle-helm-customizing.md
ajp-io: Looks good.

docs/vendor/support-bundle-helm-customizing.md
ajp-io: An engineer would know better than I do. You could be right. I know you can point to an HTTP endpoint, but maybe we modified this to include an OCI registry too (like back when we were first working on Helm CLI support).

docs/vendor/support-bundle-helm-customizing.md
ajp-io: I think we should clarify that `renderpreflights` thing. The name itself is ambiguous—I assume it's a name we chose and not a hardcoded string, but the name isn't obvious to me. Maybe this lets you skip preflights. My guess is that it lets you skip preflights somehow, so we don't need it for support bundles, but we should confirm and clarify in that doc too.

paigecalvert: The `renderpreflights` thing is indeed just an example string. The purpose of it is: "If you use this input kind, you must use conditional statements because the custom resource definition is not installed in the cluster." As in, if you want to create a preflight definition in a Preflight custom resource manifest and add it to your templates (instead of creating a preflight spec in a Secret or ConfigMap), then you need to wrap it in a conditional statement that evaluates to true so that the preflight custom resource is installed in the cluster. (I think you can always skip preflights for Helm CLI installs since it requires running a separate command anyway, right?)

What I wasn't understanding was why that same thing doesn't seem to apply to this support bundle spec. Like, is the spec supposed to get installed to the cluster?

ajp-io: Still trying to get my head fully around that. If the CRD isn't installed, what does the value do? I don't quite see how it's any different.

But I get your point about why would that not also be needed for support bundles.

docs/partials/support-bundles/_generate-bundle.mdx
ajp-io: This could be read to say that 1.104.0 and later can't use YAML files or a URL, which isn't true.

paigecalvert: Ah good call

docs/vendor/support-bundle-helm-customizing.md
paigecalvert: Diamon's looking into it: https://replicated.slack.com/archives/C016DU6CXNE/p1699304942774859

docs/vendor/support-bundle-helm-customizing.md
paigecalvert: Yeah good point. Maybe this option can just be removed from the docs all together if it's not recommended and added to a community article instead...I'll add this to a new story to investigate more fully.

docs/vendor/support-bundle-helm-customizing.md
paigecalvert: Sounds good, will leave as is

docs/vendor/support-bundle-helm-customizing.md
paigecalvert: Diamon says "Looks like the answer is yes". so I'll add this statement back in and then take the same approach to investigate further as part of a follow up story

paigecalvert: Okay, added this info back in with an example of what this is talking about: 

```
:::note
For specifications created in a SupportBundle custom resource, Helm templates are supported only when the support bundle is generated from a link to the specification at an OCI registry. For example, `kubectl support-bundle oci://my.oci.registry/image`.
:::
```

Figured this could be a placeholder, but still needs to be revisited to fully document generating a SB from a spec at an OCI registry URL

ajp-io (approved):
I fixed one small typo, but otherwise I approve. I think we can revisit these docs as part of our review of KOTS vs. Helm docs anyway.

PR #1593, opened 2023-11-06
Clarify purpose of firewalls table

paigecalvert:


docs/enterprise/installing-general-requirements.md
```diff
@@ -286,23 +286,29 @@ KOTS has been tested for compatibility with the following registries:
...
+| `proxy-auth.replicated.com` | Required&#42;  | Required&#42;   | To pull private images through `proxy.replicated.com`, the on-prem Docker client must authenticate with this service using a license ID. This domain is owned by Replicated, Inc which is headquartered in Los Angeles, CA. For the list of IP addresses for `proxy-auth.replicated.com`, see [replicatedhq/ips](https://github.com/replicatedhq/ips/blob/main/ip_addresses.json#L52-L57) in GitHub.   |
+| `registry.replicated.com` | Required&#42;&#42;         | Required&#42;&#42;            | Some applications host private images in the Replicated registry at this domain. The on-prem docker client uses a license ID to authenticate to `registry.replicated.com`. This domain is owned by Replicated, Inc which is headquartered in Los Angeles, CA. For the list of IP addresses for `registry.replicated.com`, see [replicatedhq/ips](https://github.com/replicatedhq/ips/blob/main/ip_addresses.json#L20-L25) in GitHub.
+| kots.io              | Required     | Not Required | Requests are made to this domain when installing the Replicated kots CLI. This domain is owned by Replicated, Inc., which is headquartered in Los Angeles, CA.|
```
paigecalvert, line 307: I didn't see kots.io listed in the IP addresses repo, so I guess that means that the kots.io domain just needs to be added to an allowlist?

docs/enterprise/installing-general-requirements.md
```diff
@@ -286,23 +286,29 @@ KOTS has been tested for compatibility with the following registries:
...
+| github.com           | Required     | Not Required | Requests are made to this domain when installing the Replicated kots CLI. |
```
paigecalvert, line 308: Do they just need to add github.com to their allowlist? Is there upstream docs that we'd need to link to here as well?

docs/enterprise/installing-general-requirements.md
```diff
@@ -286,23 +286,29 @@ KOTS has been tested for compatibility with the following registries:
...
+| k8s.kurl.sh<br />s3.kurl.sh    | Not Required | Required     | Embedded cluster installation scripts and artifacts are served from [kurl.sh](https://kurl.sh). An application identifier is sent in a URL path, and bash scripts and binary executables are served from kurl.sh. This domain is owned by Replicated, Inc., which is headquartered in Los Angeles, CA. For the list of IP addresses for the `k8s.kurl.sh` service, see [replicatedhq/ips](https://github.com/replicatedhq/ips/blob/main/ip_addresses.json#L34-L39) in GitHub. |
```
paigecalvert, line 309: I saw k8s.kurl.sh listed in the IP addresses repo, but not s3.kurl.sh. Do you know what the guidance is for that domain?

docs/enterprise/installing-general-requirements.md
diamonwiggins: I don't believe kots.io is a requirement for on-prem installs.

docs/enterprise/installing-general-requirements.md
diamonwiggins: I think each individual org will decide how far they want to go here in allowing for the entire domain or specifically an IP address range. I don't think the honest is on us to link to documentation for every third party if we are providing at least one domain here contrary to Dockerhub, but if you wanted to link to some docs, i don't think it would hurt - https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-githubs-ip-addresses

docs/enterprise/installing-general-requirements.md
diamonwiggins: Ah interesting. The IPs for kurl.sh are the same for s3.kurl.sh, so maybe we don't need to update the IPs repo, but that's a good call out. s3.kurl.sh is used for Airgap installs I believe.

diamonwiggins (dismissed):

docs/enterprise/installing-general-requirements.md
paigecalvert: That looks correct based on this table (looks like it's listed as Required for existing cluster and not required for embedded cluster). To make sure I'm interpreting this correctly: is knowing the IP addresses only something that's required for existing cluster installs?

diamonwiggins (approved):
lgtm

PR #1596, opened 2023-11-06
Add Google Artifact Registry to supported registries

divolgin:


paigecalvert (approved):

docs/vendor/packaging-private-images.md
```diff
@@ -10,6 +10,7 @@ Replicated recommends that application vendors use one the following external pr
 * DockerHub
 * GitHub Container Registry
 * Google Container Registry
+* Google Artifact Registry
```
obfuscurity, line 13: small nit but if we're following alphabetical convention here (and it appears we are), we should place GAR one line up.

paigecalvert (approved):

PR #1600, opened 2023-11-08
Remove air gap limitation and mentions of KOTS from custom metrics

sgalsaleh:


docs/vendor/custom-metrics.md
```diff
@@ -18,9 +18,9 @@ Custom metrics can be used to generate insights on customer usage and adoption o
 
 ## How the Vendor Portal Collects Custom Metrics
 
-The vendor portal collects custom metrics through Replicated KOTS or through the Replicated SDK, depending on which is installed in the cluster alongside the application.
+The vendor portal collects custom metrics through the Replicated SDK that is installed in the cluster alongside the application.
 
-KOTS and the SDK both expose an in-cluster API where you can configure your application to POST metric payloads. When an application instance sends data to the API, KOTS or the SDK sends the data (including any custom and built-in metrics) to the Replicated app service. The app service is located at `replicated.app` or at your custom domain.
+The Replicated SDK exposes an in-cluster API where you can configure your application to POST metric payloads. When an application instance sends data to the API, the SDK sends the data (including any custom and built-in metrics) to the Replicated app service. The app service is located at `replicated.app` or at your custom domain.
```
ajp-io, line 23: I'll let Paige comment, but she might refer to it as "the Replicated SDK" once and then "the SDK" thereafter.

docs/vendor/custom-metrics.md
```diff
@@ -32,20 +32,14 @@ The following diagram demonstrates how a custom `activeUsers` metric is sent to
 
 ## Requirements
 
-To support the collection of custom metrics, one of the following must be running in the cluster alongside the application instance:
-* Replicated KOTS version 1.102.0 or later
-* Replicated SDK version 1.0.0-beta.5 or later
+To support the collection of custom metrics, Replicated SDK version 1.0.0-beta.5 or later must be running in the cluster alongside the application instance.
```
ajp-io, line 35: Air gap support was only added in 1.0.0-beta.12. So we either need to just say 1.0.0-beta.12 here, or we could say that online requires 1.0.0-beta.5 and later and air gap requires 1.0.0-beta.12 or later. The next line talks about gracefully handling 404s. I assume air gap gives you a 404 before 1.0.0-beta.12, so maybe we should just list 1.0.0-beta.12 as the required version.

docs/vendor/custom-metrics.md
```diff
@@ -32,20 +32,14 @@ The following diagram demonstrates how a custom `activeUsers` metric is sent to
...
 
-If you have any customers that are running earlier versions of KOTS or the SDK, Replicated recommends that you add logic to your application to gracefully handle a 404 from the in-cluster APIs.
+If you have any customers that are running earlier versions of the SDK, Replicated recommends that you add logic to your application to gracefully handle a 404 from the in-cluster APIs.
```
ajp-io, line 37: Just removing extraneous words.
```suggestion
If you have any customers running earlier versions of the SDK, Replicated recommends that you add logic to your application to gracefully handle a 404 from the in-cluster APIs.
```

docs/vendor/custom-metrics.md
```diff
@@ -76,45 +70,22 @@ Custom metrics have the following limitations:
 
 You can configure your application to POST a set of metrics as key value pairs to the API that is running in the cluster alongside the application instance.
 
-Both KOTS and the SDK expose an in-cluster API:
+The Replicated SDK provides an in-cluster API custom metrics endpoint at `http://replicated:3000/api/v1/app/custom-metrics`.
```
ajp-io, line 73: I'll leave it to Paige whether we need to add the endpoint here or whether the example below suffices.

docs/vendor/custom-metrics.md
paigecalvert: Yes Replicated SDK would generally just be the first mention

docs/vendor/custom-metrics.md
paigecalvert: I think it makes sense to just list `beta.12`, especially since it's pretty trivial to just update the number in the dependency. Should we explicitly state somewhere that both online and air gap is supported?

ajp-io: That would probably good since someone might expect it to not work in air gap

docs/vendor/custom-metrics.md
paigecalvert: I like the inclusion of the endpoint. Helps to complete the first sentence I think

paigecalvert (approved):

PR #1627, opened 2023-11-15
Managing Helm chart resources

paigecalvert:
- [Orchestrating Resource Deployment](https://deploy-preview-1627--replicated-docs.netlify.app/vendor/orchestrating-resource-deployment): 
  - Adds info about HelmChart `weight`
  - Some reorg accordingly
- [Conditionally Including or Excluding Resources](https://deploy-preview-1627--replicated-docs.netlify.app/vendor/packaging-include-resources):  
  - Adds info about optional dependencies for Helm charts
  - Adds info about HelmChart `exclude` field
  - Some reorg accordingly


sidebars.js
```diff
@@ -31,14 +31,11 @@ const sidebars = {
     'vendor/distributing-workflow',
     {
       type: 'category',
-      label: 'Helm Charts with KOTS',
+      label: 'Distributing Helm Chart-Based Applications with KOTS',
```
paigecalvert, line 34: Hoping to clarify that this section specifically has the topics about distributing helm charts with KOTS, not _all_ of the topics that even mention Helm charts

sidebars.js
```diff
@@ -31,14 +31,11 @@ const sidebars = {
...
       items: [
         'vendor/helm-native-about',
         'vendor/helm-native-v2-using',
         'vendor/helm-v2-migrate',
-        'vendor/helm-native-helm-install-order',
-        'vendor/helm-optional-charts',
-        'vendor/helm-optional-value-keys',
```
paigecalvert, line 41: - `vendor/helm-native-helm-install-order `is deleted and the content is merged with `vendor/orchestrating-resource-deployment`
- `vendor/helm-optional-charts` and `vendor/helm-optional-value-keys` were relocated

sidebars.js
```diff
@@ -98,13 +96,20 @@ const sidebars = {
...
-        'vendor/packaging-include-resources',
+        {
+          type: 'category',
+            label: 'Conditionally Deploying Resources',
+            items: [
+              'vendor/packaging-include-resources',
+              'vendor/helm-optional-charts',
+              'vendor/tutorial-adding-db-config',
+            ],
+        },
```
paigecalvert, line 107: Created this new Conditionally Deploying Resources subcategory under Managing Resources and Objects to group together the main topic plus the two "example-style" topics that existed on this subject

docs/vendor/packaging-include-resources.md
```diff
@@ -1,28 +1,81 @@
 # Conditionally Including or Excluding Resources
 
-This topic describes how to use Replicated KOTS annotations and Replicated template functions to include or exclude optional resources based on a conditional statement.
+This topic describes how to include or exclude optional application resources based on one or more conditional statements. The information in this topic applies to Helm chart- and standard manifest-based applications.
```
paigecalvert, line 3: Updated this topic to include:
- Optional dependencies for helm charts
- HelmChart exclude field
- Updated Overview with info about Replicated template functions

Did some reorg/section renaming accordingly

docs/vendor/helm-optional-value-keys.md
```diff
@@ -1,4 +1,4 @@
-# Configuring Optional Value Keys
+# Configuring Optional Helm Chart Value Keys
```
paigecalvert, line 1: ^ making it more clear that this is a Helm chart thing

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,98 @@
+import HooksLimitation from "../partials/helm/_hooks-limitation.mdx"
+import WeightLimitation from "../partials/helm/_helm-cr-weight-limitation.mdx"
+import HookWeightsLimitation from "../partials/helm/_hook-weights-limitation.mdx"
+
 # Orchestrating Resource Deployment
 
-Vendors often need to control the how Kubernetes resources are deployed. For example, some applications require that certain resources are deployed in a Ready state before other resources can be deployed.
+This topic describes how to orchestrate the deployment order of resources deployed by your application. The information in this topic applies to Helm chart- and standard manifest-based applications deployed with Replicated KOTS.
```
paigecalvert, line 7: Updated this topic to:
- Add HelmChart weight field 
- Update overview content
- Simplify headings/some reorg

sidebars.js
```diff
@@ -59,6 +56,7 @@ const sidebars = {
           'vendor/admin-console-customize-config-screen',
           'vendor/config-screen-map-inputs',
           'vendor/config-screen-conditional',
+          'vendor/helm-optional-value-keys',
```
paigecalvert, line 59: `'vendor/helm-optional-value-keys'` (https://deploy-preview-1627--replicated-docs.netlify.app/vendor/helm-optional-value-keys) discusses using the values and optionalValues fields in the HelmChart custom resource to create a mapping from the Config screen.

This topic could use improvement at some point, but for the sake of this PR, I figured relocating it to the Configuration Screen section might make sense as it's talking about mapping values. Open to other thoughts on that

docs/vendor/packaging-include-resources.md
```diff
@@ -51,11 +104,11 @@ spec:
 ...
 ```
 
-## Conditionally Include a Resource
+### `kots.io/when`
 
 When the `kots.io/when: '<bool>'` annotation is present on a resource and evaluates to false, the resource is excluded from the deployment.
 
-The following example excludes the postgres `StatefulSet` resource when the `install_postgres` checkbox is enabled:
+The following example uses the `kots.io/exclude` annotation and the ConfigOptionEquals template function to exclude the postgresql `StatefulSet` resource when the `install_postgres` checkbox on the admin console **Config** page is enabled:
```
ajp-io, line 111: I think this suggestion is correct… I think this and the above are the exact same example just inverted.
```suggestion
The following example uses the `kots.io/when` annotation and the ConfigOptionEquals template function to exclude the postgresql `StatefulSet` resource when the `install_postgres` checkbox on the admin console **Config** page is disabled:
```

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,98 @@
...
+
+## Helm Chart Deployment Order with `weight` {#weight}
+
+You can configure the [`weight`](/reference/custom-resource-helmchart-v2#weight) property of the Replicated HelmChart custom resource to define the order in which the Helm charts in your release are installed.
+
+KOTS directs Helm to install the Helm charts based on the value of `weight` in ascending order, deploying the chart with the lowest weight first. For example, a chart with a `weight` of `-1` deploys before a chart with a `weight` of `0`.
+
+The value for the `weight` property can be any negative or positive integer or `0`. By default, when you do not provide a `weight` for a Helm chart, the `weight` is `0`.
+
+Assigning a `weight` helps you avoid relying on Helm dependencies and subcharts to define a chart installation order with KOTS. For example, when you include hooks in your Helm charts, Helm waits for certain hooks to complete before continuing.
```
ajp-io, line 25: I'm not totally sure I know what this means. Does this imply that using subcharts lets you define an installation order but you can do it with KOTS instead?

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,98 @@
...
+
+### `weight` with Subcharts and Dependencies
+
+When you add a `weight` property to HelmChart custom resources in your application, KOTS instructs Helm to install any dependencies, including subcharts, along with the parent chart.
+
+For example, if you have two Helm charts in your application, one with a `weight` of `-1` and one with a `weight` of `0`, then Helm installs the chart with a `weight` of `-1` first, including any subcharts and dependencies listed in the `dependencies` field for that chart.
+
+If you do not add a `weight` to Helm charts in your application, you can still use dependencies and subcharts to define installation order constraints during application deployment.
+
+For more information about using Helm dependencies, see [Chart Dependencies](https://helm.sh/docs/topics/charts/#chart-dependencies) in the Helm documentation.
```
ajp-io, line 51: I don't know that subcharts/dependencies let you define an installation order. I wonder if this section can just be removed, assuming I'm right.

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,98 @@
...
+
+### `weight` with Hooks
```
ajp-io, line 53: I think this section conflates the `weight` that we offer in HelmChart and the `weight` that Helm supports for hooks ([docs](https://helm.sh/docs/topics/charts_hooks/)). At minimum, that would have to be clarified.

But I actually wonder if we need most of this section anyway. Anything about hook weights seems like a Helm feature we don't need to document. We do need to mention which hooks we support, but I don't think the page on orchestrating resource deployment is the right place.

docs/vendor/orchestrating-resource-deployment.md
paigecalvert: Yeah, it does look like this can be removed. I think the goal of this section is just to explain how the weight property works if you have subcharts (as in, "FYI: weight applies to any subcharts too"). But that can just be a sentence added to the main section above

docs/vendor/orchestrating-resource-deployment.md
paigecalvert: > We do need to mention which hooks we support, but I don't think the page on orchestrating resource deployment is the right place.

That's a good point. We do have some limitations listed over here on the overview page: docs.replicated.com/vendor/helm-native-about#kotsiov1beta2-limitations. Could add something over on that page too about which hooks are supported

docs/vendor/packaging-include-resources.md
```diff
@@ -1,28 +1,81 @@
...
-The `kots.io/exclude` and `kots.io/when` annotations have the following requirements:
+For example, the Replicated ConfigOptionEquals template functions returns true if the specified configuration option value is equal to a supplied value. This is useful for creating conditional statements that include or exclude a resource based on a user's application configuration choices.
+
+For more information about the available Replicated template functions, see [About Template Functions](/reference/template-functions-about).
+
+## Include or Exclude Helm Charts
+
+This section describes methods for including or excluding Helm charts from your application deployment.
+
+### Optional Dependencies
```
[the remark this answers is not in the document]
paigecalvert: Yeah good call. I removed the example and the description in favor of a sentence and a link

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,90 @@
...
+
+## Overview
+
+Many applications require that certain resources are deployed and in a ready state before other resources can be deployed. For applications deployed with KOTS, you can manage the order in which resources are deployed using the following methods:
+
+* For Helm charts, set the `weight` property in the corresponding HelmChart custom resource. See [Helm Chart Deployment Order with `weight`](#weight).
+
+* For standard manifests, add KOTS annotations to the resources. See [Standard Manifest Deployment Order with KOTS Annotations](#manifests).
+
+## Helm Chart Deployment Order
```
paigecalvert, line 17: Note: Got rid of this section in favor of just explaining that subcharts are also installed with the parent chart, including when a weight is set

Screenshot 2023-11-27 at 11 06 11 AM

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,90 @@
...
+
+This section describes the supported methods for controlling the deployment order of Helm charts in your release.
+
+### HelmChart `weight` {#weight}
+
+You can configure the [`weight`](/reference/custom-resource-helmchart-v2#weight) property of the Replicated HelmChart custom resource to define the order in which the Helm charts in your release are installed.
+
+KOTS directs Helm to install the Helm charts based on the value of `weight` in ascending order, deploying the chart with the lowest weight first. Any dependencies are installed along with the parent chart. For example, a chart with a `weight` of `-1` deploys before a chart with a `weight` of `0`.
```
paigecalvert, line 25: Note: Got rid of this section in favor of just explaining that subcharts are also installed with the parent chart, including when a weight is set

Screenshot 2023-11-27 at 11 06 11 AM

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,90 @@
...
+#### Limitations
+
+The `weight` field in the HelmChart custom resource has the following limitations:
+
+* <WeightLimitation/>
+
+* When installing a Helm chart-based application, KOTS always deploys standard Kubernetes manifests to the cluster _before_ deploying Helm charts. For example, if your release contains a Helm chart, a CRD, and a ConfigMap, then the CRD and ConfigMap resources are deployed before the Helm chart. The `weight` property does not allow Helm charts to be deployed before standard manifests.
 
-To give you more control over the deployment of Kubernetes resources, Replicated KOTS uses annotations to allow for more fine-grained control over the order in which resources are deployed.
+### Support for Helm Hooks
```
paigecalvert, line 53: Previously, this was nested under the section about `weight` even though it's discussing a completely separate topic. Renamed it and bumped it up a level to better differentiate

docs/vendor/orchestrating-resource-deployment.md
paigecalvert: @ajp-io  Okay so I went back and forth on it and ended up creating a separate section on this page called "Support for Helm Hooks", since hooks are relevant to controlling resource deployment for helm charts. That said, I could also see this "Support for Helm Hooks" being moved to [About Distributing Helm Charts with KOTS](https://docs.replicated.com/vendor/helm-native-about#overview), since it sounds like people can use hooks for more than just resources deployment (for example, running a job before/after something happens). Curious what you think

docs/vendor/orchestrating-resource-deployment.md
cbodonnell: makes sense to me since the HelmChart custom resource is only used for parent charts and I would expect subcharts to be installed along with the parent chart since that's standard Helm behavior

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,90 @@
...
-#### `kots.io/creation-phase: '<integer>'`
+* <HooksLimitation/>
 
-When this annotation is present on a resource, KOTS groups the resource into the specified creation phase. KOTS deploys each phase in order from lowest to highest. Resources within the same phase are deployed in the same order that Helm installs resources. For more information, see [Using Helm](https://helm.sh/docs/intro/using_helm/#:~:text=Helm%20installs%20resources%20in%20the,order) in the Helm documentation.
+* <HookWeightsLimitation/>
+  
+## Standard Manifest Deployment Order with KOTS Annotations {#manifests}
 
-#### `kots.io/deletion-phase: '<integer>'`
+You can use the KOTS annotations descirbed in this section to control the order in which standard manifests are deployed.
```
cbodonnell, line 75:
```suggestion
You can use the KOTS annotations described in this section to control the order in which standard manifests are deployed.
```

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -35,23 +93,30 @@ metadata:
...
+  name: myresources.example.com
+  annotations:
+    kots.io/creation-phase: "-1"
+    kots.io/deletion-phase: "1"
+...
+```
+### `kots.io/wait-for-ready`
 
-#### Example
+When the `kots.io/wait-for-ready: '<bool>'` annotation is present on a resource and evaluates to `'true'`, KOTS waits for the resource to be in a Ready state before deploying any other resources. To determine if a resource is ready, KOTS uses any status informers that were created for the resource type. If there is no existing status informer for the given resource type, then KOTS waits until the resource exists and is queryable from the Kubernetes API server. For more information about status informers, see [Adding Resource Status Informers](/vendor/admin-console-display-app-status).
```
cbodonnell, line 117: > To determine if a resource is ready, KOTS uses any status informers that were created for the resource type.

Just clarifying that this doesn't have to do with any of the status informers that a vendor would specify in their app spec, but is more related to whether KOTS has existing status informer logic for a particular resource type to know what "ready" means (which is [this list](https://github.com/replicatedhq/kots/blob/main/pkg/appstate/wait.go#L25-L30)).

paigecalvert: Ahh, thank you. I did interpret this to mean the status informers that the vendor configures.

paigecalvert: Updated:

>For most resource types, KOTS has existing logic to determine if a resource is ready. If there is no existing logic for the given resource type, then KOTS waits until the resource exists and is queryable from the Kubernetes API server.

docs/vendor/packaging-include-resources.md
```diff
@@ -51,11 +104,11 @@ spec:
 ...
 ```
 
-## Conditionally Include a Resource
+### `kots.io/when`
 
 When the `kots.io/when: '<bool>'` annotation is present on a resource and evaluates to false, the resource is excluded from the deployment.
```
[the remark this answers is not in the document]
ajp-io: Okay, I'll bring up a possible deprecation with our team then.

docs/vendor/orchestrating-resource-deployment.md
ajp-io: The hooks let you do something while a particular chart is installed. But they do nothing for orchestrating the deployment of multiple charts. If you're installing with Helm yourself, you could just run the install commands in a particular order. But since KOTS is doing it for you, you need to tell it the order with `weight`.

docs/vendor/orchestrating-resource-deployment.md
ajp-io: Let's double check with @cbodonnell, but I don't think hooks and their weights have to do with deployment order. They let you intervene when a particular chart is installed, but they don't really do ordering. The only ordering I see is that hooks have a weight so you can put _the hooks_ in a particular order. But that's just a Helm thing anyway. Chart ordering is just a KOTS `weight` thing.

Let me know if you need to chat about this to clarify anything.

docs/vendor/orchestrating-resource-deployment.md
paigecalvert: I updated this as part of moving it to the other topic and out of Orchestrating Resource Deployment. I figured we didn't really even need to try to define hooks

sidebars.js
ajp-io: At a glance, this looks like the right section to me.

docs/vendor/helm-optional-value-keys.md
```diff
@@ -1,19 +1,88 @@
...
 
-Some advanced use cases involve writing values to a values file only if there is a value to apply. For example, if a customer has the option to configure a different value than the default value in your file, such as setting the database option they want to use. You can include an optional value key so that your application can dynamically include it based on a `when` condition. 
+## Override Existing Value Keys
 
-For more information about the `optionalValues` property, see [`optionalValues`](/reference/custom-resource-helmchart-v2#optionalvalues) in _HelmChart v2 (Beta)_.
+During installation or upgrade, KOTS uses the values in the HelmChart custom resource `values` key to override the values of any matching keys in the corresponding Helm chart `values.yaml` file.
+
+You can use the HelmChart custom resource `values` key to override existing values in the Helm chart `values.yaml`, including updating a value or setting a key to `null` to delete it.
+
+### Update a Value
```
ajp-io, line 19: I don't want to block this PR, so let's make a note and come back to it separately, but I have some thoughts about the wording here and how to make it more natural to Helm users. I think "key" has its place but "value" is generally more useful, and "setting" a value is more natural than "updating."

docs/vendor/helm-optional-value-keys.md
```diff
@@ -44,11 +113,7 @@ externalDatabase:
   ##
   database: bitnami_wordpress
 ```
-If a user wants to configure an external database, you can enable this dynamically through your deployment by adding an `optionalValues` section to the HelmChart custom resource, instead of attempting to modify the render logic in the Helm chart.
-
-The `optionalValues` section includes a `when` condition that instructs KOTS how to determine if these keys should be merged. It also includes a `recursiveMerge` field that defines how to merge the dataset.
-
-For example, the HelmChart custom resource can be configured as follows:
+If a user wants to configure an external database, you can enable this dynamically through your deployment by adding an `optionalValues` section to the HelmChart custom resource, instead of attempting to modify the render logic in the Helm chart:
```
ajp-io, line 116: Not sure I understand the "through your deployment" part, or at least it doesn't feel needed. And I don't totally follow the "modify the render logic" part. I feel like it makes just as much sense like this.
```suggestion
If a user wants to configure an external database, you can enable this dynamically by adding an `optionalValues` section to the HelmChart custom resource:
```

docs/vendor/orchestrating-resource-deployment.md
```diff
@@ -1,32 +1,70 @@
...
 # Orchestrating Resource Deployment
 
-Vendors often need to control the how Kubernetes resources are deployed. For example, some applications require that certain resources are deployed in a Ready state before other resources can be deployed.
+This topic describes how to orchestrate the deployment order of resources deployed as part of your application. The information in this topic applies to Helm chart- and standard manifest-based applications deployed with Replicated KOTS.
+
+## Overview
+
+Many applications require that certain resources are deployed and in a ready state before other resources can be deployed. For applications deployed with KOTS, you can manage the order in which resources are deployed using the following methods:
+
+* For Helm charts, set the `weight` property in the corresponding HelmChart custom resource. KOTS also supports some Helm hooks. See [HelmChart `weight`](#weight) or [Support for Helm Hooks](#hooks).
```
ajp-io, line 13:
```suggestion
* For Helm charts, set the `weight` property in the corresponding HelmChart custom resource. KOTS also supports some Helm hooks. See [HelmChart `weight`](#weight) or [Support for Helm Hooks](/vendor/helm-native-about#hooks).
```

ajp-io: Is it weird now that the other links are on this page but this links to another page? If we're going to mention it, should it be on this page? Sorry if I'm back and forth here… I was coming to some realizations about hooks too. Feel free to chat about this if you want.

paigecalvert: Oops good catch on that. I'll actually just remove that link all together since we no longer want to have the info about hooks in the orchestrating resource deployment topic

PR #1610, opened 2023-11-10
Clarify options for defining preflights

paigecalvert:
Defining Preflight Checks: https://deploy-preview-1610--replicated-docs-upgrade.netlify.app/vendor/preflight-defining

Running Preflight Checks for Helm Installations: https://deploy-preview-1610--replicated-docs-upgrade.netlify.app/vendor/preflight-running

PR #1645, opened 2023-11-20
Simplify support bundle options

paigecalvert:
This is the follow-up PR to https://github.com/replicatedhq/replicated-docs/pull/1610, which simplified the Preflight Checks documentation.

New topic that condenses the information about how to add/customize support bundles for Helm chart- and standard-manifest-based apps into one topic (previously was split across two): https://deploy-preview-1645--replicated-docs.netlify.app/vendor/support-bundle-customizing


docs/vendor/preflight-defining.md
```diff
@@ -23,7 +23,9 @@ import TabItem from '@theme/TabItem';
 
 # Defining Preflight Checks
```
paigecalvert, line 24: Just a few changes here to match some updates I made to the corresponding Support Bundle topic

docs/vendor/support-bundle-customizing.md
```diff
@@ -0,0 +1,350 @@
...
+import LogsSelectorsSecret from "../partials/support-bundles/_logs-selectors-secret.mdx"
+import LogsSelectorsCr from "../partials/support-bundles/_logs-selectors-cr.mdx"
+import LogsLimitsSecret from "../partials/support-bundles/_logs-limits-secret.mdx"
+import LogsLimitsCr from "../partials/support-bundles/_logs-limits-cr.mdx"
+import RedisMysqlSecret from "../partials/support-bundles/_redis-mysql-secret.mdx"
+import RedisMysqlCr from "../partials/support-bundles/_redis-mysql-cr.mdx"
+import RunPodsSecret from "../partials/support-bundles/_run-pods-secret.mdx"
+import RunPodsCr from "../partials/support-bundles/_run-pods-cr.mdx"
+
+# Adding and Customizing Support Bundles
```
paigecalvert, line 22: Called this "Adding and Customizing Support Bundles" rather than using the same "Defining Preflight Checks" verbiage because there are actually two different tasks when it comes to support bundles:
- You can add the default (empty) spec and call it a day
- You can also then optionally customize the spec for your app

docs/vendor/support-bundle-customizing.md
```diff
@@ -0,0 +1,350 @@
...
+
+This topic describes how to add a default support bundle specification to a release for your application. It also describes how to optionally customize the default support bundle specification based on your application's needs. For more information about support bundles, see [About Preflight Checks and Support Bundles](/vendor/preflight-support-bundle-about).
+
+The information in this topic applies to Helm chart- and standard manifest-based application installed with the Helm CLI or with Replicated KOTS.
+
+## Add the Specification to a Manifest File
+
+This section describes how to add an empty support bundle specification to a manifest file. An empty support bundle specification includes the following collectors by default:
```
paigecalvert, line 30: Not sure if we like the word "empty" here, but it was the simplest terminology that came to mind. Essentially, trying to convey that you can add a blank spec and the support bundle works by default

docs/vendor/support-bundle-customizing.md
```diff
@@ -0,0 +1,350 @@
...
+* [clusterInfo](https://troubleshoot.sh/docs/collect/cluster-info/)
+* [clusterResources](https://troubleshoot.sh/docs/collect/cluster-resources/)
+
+You do not need manually include the `clusterInfo` or `clusterResources` collectors in the specification.
```
paigecalvert, line 34: Q: Are the cluster info and cluster resources collectors that only things included when you add an empty support bundle spec? Or is it all of this? https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/defaultspec/spec.yaml

I was having a hard time figuring out what was included by default based on the existing content in the docs

docs/vendor/support-bundle-customizing.md
```diff
@@ -0,0 +1,350 @@
...
+  collectors: []
+  analyzers: []
+```
+For more information about the SupportBundle custom resource, see [Preflight and Support Bundle](/reference/custom-resource-preflight).
+
+## (Optional) Customize the Specification
+
+You can optionally customize the support bundles for your application by:
+* Adding collectors and analyzers
+* Editing or excluding the default `clusterInfo` and `clusterResources` collectors
```
paigecalvert, line 117: Q: Similar to above, do we need to talk about editing the whole default spec using the content here https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/defaultspec/spec.yaml? 

If so, I could reframe this to talk about "editing the default spec" rather than just editing the two default collectors

docs/vendor/support-bundle-customizing.md
```diff
@@ -0,0 +1,350 @@
...
+        ignoreRBAC: true
+```
+
+The following example shows how to use Helm template functions to set the namespace:
+
+```yaml
+spec:
+  collectors:
+    - clusterResources:
+        namespace: {{ .Release.Namespace }}
```
paigecalvert, line 179: Q: I couldn't figure out if this key was `namespaces` or `namespace`. In the troubleshoot docs, it says `namespaces`. However, I saw some examples (such as [slackernews](https://github.com/slackernews/slackernews/blob/main/chart/slackernews/templates/support-bundle.yaml)) that used `namespace`

docs/vendor/support-bundle-customizing.md
```diff
@@ -0,0 +1,350 @@
...
+- [clusterPodStatuses](https://troubleshoot.sh/docs/analyze/cluster-pod-statuses/)
+- [replicasetStatus](https://troubleshoot.sh/docs/analyze/replicaset-status/)
+- [statefulsetStatus](https://troubleshoot.sh/docs/analyze/statefulset-status/)
+- [postgresql](https://troubleshoot.sh/docs/analyze/postgresql/), [mysql](https://troubleshoot.sh/docs/analyze/mysql/), and [redis](https://troubleshoot.sh/docs/analyze/redis/)
+
+### Customize the Default `clusterResources` Collector
+
+You can edit the default `clusterResources` using the following properties:
+
+* `namespaces`: The list of namespaces where the resources and information is collected. If the `namespaces` key is not specified, then the `clusterResources` collector defaults to collecting information from all namespaces. The `default` namespace cannot be removed, but you can specify additional namespaces.
```
paigecalvert, line 155: >The `default` namespace cannot be removed, but you can specify additional namespaces.

This was in the docs previously, and I'm not sure I follow. It sounds like it's saying: if you want to scope this collector to only a few namespaces, then you always have to manually list the `default` namespace as well. Is that true? Or is it just saying that even if I don't list `default`, `default` is always included?

docs/vendor/support-bundle-customizing.md
```diff
@@ -0,0 +1,350 @@
...
+```yaml
+spec:
+  collectors:
+    - clusterInfo:
+        exclude: true
+    - clusterResources:
+        exclude: true
+```
+
+## Example Specifications
```
paigecalvert, line 209: Used the same format as the examples in the corresponding [Defining Preflight Checks](https://docs.replicated.com/vendor/preflight-defining#example-specifications) topic for this section

docs/partials/support-bundles/_deploy-status-cr.mdx
```diff
@@ -0,0 +1,21 @@
+```yaml
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: example
+spec:
+  collectors: []
```
paigecalvert, line 7: In some examples that I saw (eg [slackernews](https://github.com/slackernews/slackernews/blob/main/chart/slackernews/templates/support-bundle.yaml)), people used `collectors: {}` instead of `collectors: []`. Is there a significant difference between these formats?

docs/partials/support-bundles/_http-requests-cr.mdx
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: example
+spec:
+  collectors:
+    - http:
+        collectorName: slack
+        get:
+          url: https://api.slack.com/methods/api.test
```
paigecalvert, line 11: As I understood it, adding a collector like this just _adds_ to the default collectors and doesn't overwrite anything. Is that correct?

docs/partials/support-bundles/_deploy-status-cr.mdx
banjoh: `collectors: []` is the correct syntax. Using `{}` is not the correct syntax

docs/partials/support-bundles/_http-requests-cr.mdx
banjoh: That is correct

banjoh (dismissed):

docs/vendor/support-bundle-customizing.md
banjoh: Its `namespaces`

https://troubleshoot.sh/docs/collect/cluster-resources/#namespaces-optional

docs/vendor/support-bundle-customizing.md
banjoh: ≥ Are the cluster info and cluster resources collectors that only things included when you add an empty support bundle spec

Yes they are. The default spec (https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/defaultspec/spec.yaml) is treated like all other specs.

PR #1738, opened 2023-12-14
Tutorial: Map config values to Helm chart

paigecalvert:
Adds a new tutorial that shows how to override values in the Grafana Helm chart using the config page:
* This tutorial was designed to replace this existing "tutorial" that uses the grafana helm chart https://docs.replicated.com/vendor/helm-mapping-example, which is out of date and impossible to follow
* Goal was to create a tutorial that 1) does not require any prior knowledge, and 2) that also emphasizes the theme of mapping the config page to helm values as opposed to focusing on the most basic install concepts

Preview: https://deploy-preview-1738--replicated-docs.netlify.app/vendor/tutorial-config-setup


docs/vendor/helm-mapping-example.md
paigecalvert, line 1: The tutorial replaces this topic, which was never really updated from the kots.io days: https://docs.replicated.com/vendor/helm-mapping-example

docs/vendor/tutorial-config-create-customer.md
```diff
@@ -0,0 +1,38 @@
+# Step 5: Create a KOTS-Enabled Customer
```
paigecalvert, line 1: Mostly copied from the new KOTS/Helm tutorial

docs/vendor/tutorial-config-create-release.md
```diff
@@ -0,0 +1,134 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import HelmChart from "../partials/getting-started/_grafana-helmchart.mdx"
+import KotsApp from "../partials/getting-started/_grafana-kots-app.mdx"
+import K8sApp from "../partials/getting-started/_grafana-k8s-app.mdx"
+import Config from "../partials/getting-started/_grafana-config.mdx"
+
+# Step 4: Add the Chart Archive to a Release
```
paigecalvert, line 8: Mostly copied from the new KOTS/Helm tutorial, with of course the addition of a Config custom resource in the release files

sidebars.js
```diff
@@ -433,6 +432,19 @@ const sidebars = {
...
+          items: [
+            'vendor/tutorial-config-setup',
+            'vendor/tutorial-config-get-chart',
+            'vendor/tutorial-config-create-app',
+            'vendor/tutorial-config-package-chart',
+            'vendor/tutorial-config-create-release',
+            'vendor/tutorial-config-create-customer',
+            'vendor/tutorial-config-install-kots',
+          ],
+        },
```
paigecalvert, line 447: Right now, I put this in the Tutorials section of the Getting Started category in the main docs sidebar. An alternative might be to put this in the KOTS sidebar in the Configuration Screen section (since that's where the example that this tutorial replaces used to be). Not sure if you have any thoughts/preference on this

docs/vendor/tutorial-config-setup.md
```diff
@@ -0,0 +1,40 @@
...
+This topic provides a summary of the goals and outcomes for the tutorial and also lists the prerequisites to set up your environment before you begin.
+
+## Summary
+
+This tutorial introduces you to mapping user-supplied values from the Replicated KOTS admin console configuration page to a Helm chart `values.yaml` file.
+
+In this tutorial, you use a sample Helm chart to learn how to:
+
+* Define a user-facing application configuration page in the KOTS admin console
+* Override Helm chart values with the user-supplied values from the admin console configuration page
```
ajp-io, line 12: I'm wondering if override is the right word versus set. I hear override and think that we might be overriding Helm in some way. Whereas I think this just shows you how to set particular Helm values from KOTS.

So I lean toward using set, but let me know what you think.

If we do change it, it looks like we say override in a lot of places, and maybe it's right in some places. I'm just responding to this instance.

docs/vendor/tutorial-config-create-release.md
```diff
@@ -0,0 +1,134 @@
...
+   </TabItem>
+   <TabItem value="k8s-app" label="k8s-app.yaml">
+   <h5>Description</h5>
+    <p>The Kubernetes Application custom resource supports functionality such as including buttons and links on the admin console dashboard. The YAML below adds an <strong>Open App</strong> button to the admin console dashboard that opens the application using the port forward configured in the KOTS Application custom resource.</p>
+    <h5>YAML</h5>
+    <K8sApp/>
+   </TabItem>
+   <TabItem value="config" label="kots-config.yaml">
+    <h5>Description</h5>
+    <p>The Config custom resource specifies a user-facing configuration page in the admin console designed for collecting application configuration preference from users. The YAML below creates "Admin User" and "Admin Password" fields that will be shown to the user on the configuration page during installation. These fields will be used to set the login credentials for Grafana.</p>
```
ajp-io, line 56:
```suggestion
    <p>The Config custom resource specifies a user-facing configuration page in the admin console designed for collecting application configuration from users. The YAML below creates "Admin User" and "Admin Password" fields that will be shown to the user on the configuration page during installation. These fields will be used to set the login credentials for Grafana.</p>
```

docs/vendor/tutorial-config-create-release.md
```diff
@@ -0,0 +1,134 @@
...
+    <h5>YAML</h5>
+    <Config/>
+   </TabItem>
+   <TabItem value="helmchart" label="grafana.yaml">
+    <h5>Description</h5>
+    <p>The KOTS HelmChart custom resource provides instructions to KOTS about how to deploy the Helm chart.</p>
+    <p>The HelmChart custom resource below contains a <code>values</code> key, which creates a mapping to the Grafana <code>values.yaml</code> file. In this case, the <code>values.admin.user</code> and <code>values.admin.password</code> fields map to <code>admin.user</code> and <code>admin.password</code> in the Grafana <code>values.yaml</code> file.</p>
+    <p>During installation, KOTS renders the ConfigOption template functions in the <code>values.admin.user</code> and <code>values.admin.password</code> fields and then overrides the corresponding Grafana values accordingly.</p>
```
ajp-io, line 64: Looking at overrides here too, and I definitely lean toward saying "set" instead.

docs/vendor/tutorial-config-install-kots.md
```diff
@@ -0,0 +1,152 @@
...
+
+   [View a larger version of this image](/images/grafana-login.png)
+
+1. On the Grafana login page, enter the username and password that you created on the **Configure Grafana** page. Confirm that you can log in to the application to access the Grafana dashboard:
+
+   <img alt="Grafana dashboard" src="/images/grafana-dashboard.png" width="500px"/>
+
+   [View a larger version of this image](/images/grafana-dashboard.png)
+
+1. Uninstall the Grafana application from your cluster:
```
ajp-io, line 114: Technically need to ctrl+c the port forward first

sidebars.js
ajp-io: I think it's too specific for its current spot. Seems lower level than the other tutorials there now.

The KOTS config screen is one option. Somewhere about Helm is another option. I could see someone wanting to know how to set Helm chart values, and this is a good example. But I guess this isn't the only way to do that (license fields, for example), so if this is more specific to the config, than maybe the config screen is the right spot.

ajp-io: Makes me wonder if we have any general guidance on mapping to Helm chart values, or if we always act like config is what you're going to use.

ajp-io: I guess this section does, though it leads you to believe the config is the only way https://deploy-preview-1738--replicated-docs.netlify.app/vendor/helm-optional-value-keys#overview

ajp-io (commented):
There's a note about requiring a certain KOTS version to deploy the SDK, which isn't needed here.

docs/vendor/tutorial-config-create-release.md
paigecalvert: Works for me 👍 I'll keep that in mind for the future

docs/vendor/tutorial-config-install-kots.md
paigecalvert: Good call thanks

sidebars.js
paigecalvert: Yeah, that page purports to give a generic overview of mapping Helm values with KOTS, though it's not quite there yet. I do have a story in my backlog for improving it (which includes adding mention of how this work for license fields too)

paigecalvert: And okay I was starting to lean towards moving it too. Especially since it's so KOTS specific, it felt weird to potentially be trying to learn about the config page, then get linked over to a totally different part of the docs to follow a tutorial on it. I can tailor the Intro to account for the fact that we are no longer assuming it's someone's first time using replicated

docs/vendor/tutorial-config-install-kots.md
```diff
@@ -0,0 +1,150 @@
...
+    • Creating namespace ✓
+    • Waiting for datastore to be ready ✓
+  Enter a new password for the admin console (6+ characters): ••••••••
+  • Waiting for Admin Console to be ready ✓
+
+  • Press Ctrl+C to exit
+  • Go to http://localhost:8800 to access the Admin Console
+  ```
+
+1. With the port forward running, in a browser, go to `http://localhost:8800` to access the admin console.
```
ajp-io, line 48: Might just be clearer wording.
```suggestion
1. With the port forward running, go to `http://localhost:8800` in a browser to access the admin console.
```

ajp-io (commented):
In my last review, I said we could remove the note about the KOTS version required to install the SDK. I totally didn't realize the SDK is added in step 3. (I think I did that step on autopilot.)

You can add that back if you want to. But do you think it's necessary to include the SDK here? It's not needed for the tutorial, but it could be good to reinforce the importance. Or we could mention you should but we'll leave it out for this tutorial.

WDYT?

PR #1744, opened 2023-12-15
Add cluster update ttl docs

paigecalvert:
New topic: https://deploy-preview-1744--replicated-docs.netlify.app/reference/replicated-cli-cluster-update-ttl

### DIFF
--- a/docs/reference/replicated-cli-cluster-create.md
+++ b/docs/reference/replicated-cli-cluster-create.md
@@ -1,5 +1,5 @@
 import Help from "../partials/replicated-cli/_help.mdx"
-
+import Output from "../partials/replicated-cli/_output.mdx"
 
 # cluster create
 
@@ -47,20 +47,16 @@ replicated cluster create [flags]
     <td>integer</td>
     <td>The node count. <strong>Default:</strong> 1</td>
   </tr>
-  <tr>
-    <td>--output</td>
-    <td>string</td>
-    <td>The output format to use. <strong>Valid values:</strong> json or table. <strong>Default:</strong> table</td>
-  </tr>
+  <Output/>
   <tr>
     <td>--ttl</td>
     <td>string</td>
-    <td>The cluster Time to Live (TTL) duration, in hours, before the cluster is automatically deleted by the service. TTL starts when the cluster is in a Ready state. <strong>Valid values:</strong> 1 - 48. <strong>Default:</strong> 1</td>
+    <td>The cluster Time to Live (TTL) duration, in hours. When the TTL expires, the cluster is automatically deleted. TTL starts when the cluster is in a Ready state. <strong>Valid values:</strong> 1h through 48h. <strong>Default:</strong> One hour</td>
   </tr>
   <tr>
     <td>--version</td>
     <td>string</td>
-    <td>(Required) The Kubernetes version to provision. For OpenShift clusters, provide the supported OpenShift version. The format is distribution dependent. For supported versions, see <a href="/vendor/testing-supported-clusters">Supported Compatibility Matrix Cluster Types (Beta)</a>.</td>
+    <td>The Kubernetes version to provision. For OpenShift clusters, provide the supported OpenShift version. The format is distribution dependent. For supported versions, see <a href="/vendor/testing-supported-clusters">Supported Compatibility Matrix Cluster Types (Beta)</a>.</td>
   </tr>
   <tr>
     <td>--wait</td>
--- a/docs/reference/replicated-cli-cluster-create.md
+++ b/docs/reference/replicated-cli-cluster-create.md
@@ -51,12 +51,12 @@ replicated cluster create [flags]
   <tr>
     <td>--ttl</td>
     <td>string</td>
-    <td>The cluster Time to Live (TTL) duration, in hours. When the TTL expires, the cluster is automatically deleted. TTL starts when the cluster is in a Ready state. <strong>Valid values:</strong> 1h through 48h. <strong>Default:</strong> One hour</td>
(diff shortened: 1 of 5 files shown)


docs/reference/replicated-cli-cluster-create.md
```diff
@@ -47,20 +47,16 @@ replicated cluster create [flags]
...
-    <td>--output</td>
-    <td>string</td>
-    <td>The output format to use. <strong>Valid values:</strong> json or table. <strong>Default:</strong> table</td>
-  </tr>
+  <Output/>
   <tr>
     <td>--ttl</td>
     <td>string</td>
-    <td>The cluster Time to Live (TTL) duration, in hours, before the cluster is automatically deleted by the service. TTL starts when the cluster is in a Ready state. <strong>Valid values:</strong> 1 - 48. <strong>Default:</strong> 1</td>
+    <td>The cluster Time to Live (TTL) duration, in hours. When the TTL expires, the cluster is automatically deleted. TTL starts when the cluster is in a Ready state. <strong>Valid values:</strong> 1h through 48h. <strong>Default:</strong> One hour</td>
```
paigecalvert, line 54: Updated the "valid values" because it looks like you are required to specify `h`, as in `2h`

docs/reference/replicated-cli-cluster-update-ttl.md
```diff
@@ -0,0 +1,49 @@
...
+    <th width="50%">Description</th>
+  </tr>
+  <Help/>
+  <Output/>
+  <tr>
+    <td><code>--ttl</code></td>
+    <td>string</td>
+    <td>
+    <p>The updated TTL for the cluster, in hours. <strong>Valid values:</strong> 1h through 48h. For example, to update the TTL to ten hours, use <code>--ttl 10h</code>.</p>
+    <p>The updated TTL must be higher than the current TTL. You cannot reduce the TTL of a cluster.</p>
```
paigecalvert, line 29: >The updated TTL must be higher than the current TTL. You cannot reduce the TTL of a cluster.

When I attempted to reduce the TTL, I got an error. So I'm assuming this is true

docs/reference/replicated-cli-cluster-update-ttl.md
```diff
@@ -0,0 +1,49 @@
+import Help from "../partials/replicated-cli/_help.mdx"
+import Output from "../partials/replicated-cli/_output.mdx"
+
+# cluster update ttl
+
+Increase the Time to Live (TTL) for a cluster provisioned with the Replicated compatibility matrix. The `cluster update ttl` command updates the expiration timestamp for a cluster and does _not_ change the creation timestamp.
```
paigecalvert, line 6: >Increase

Looks like the usage for this command is specifically to increase the TTL, not reduce

docs/vendor/testing-about.md
```diff
@@ -62,7 +62,6 @@ The compatibility matrix has the following limitations:
 - Multi-node support is available only for GKE, EKS, and AKS.
 - There is no support for IPv6.
 - The `cluster upgrade` feature is available only for kURL distributions. See [cluster upgrade](/reference/replicated-cli-cluster-upgrade).
-- Clusters have a maximum Time To Live (TTL) of 48 hours. See [Setting TTL](#setting-ttl) below.
```
paigecalvert, line 65: This isn't really a limitation, I don't think. Just a fact of the feature. And it was also pointing to a section that was deleted

docs/reference/replicated-cli-cluster-create.md
```diff
@@ -47,20 +47,16 @@ replicated cluster create [flags]
...
   </tr>
   <tr>
     <td>--version</td>
     <td>string</td>
-    <td>(Required) The Kubernetes version to provision. For OpenShift clusters, provide the supported OpenShift version. The format is distribution dependent. For supported versions, see <a href="/vendor/testing-supported-clusters">Supported Compatibility Matrix Cluster Types (Beta)</a>.</td>
```
paigecalvert, line 63: This was the only item here that was marked as required, but I don't think that's true, because I was able to create a cluster from the command line without it. Are there any flags that are required?

docs/reference/replicated-cli-cluster-create.md
jdewinne: This should be `10m` to `48h`. You can also specify for example `120m`, which would be 2 hours.

docs/reference/replicated-cli-cluster-create.md
```diff
@@ -47,20 +47,16 @@ replicated cluster create [flags]
...
+    <td>The Kubernetes version to provision. For OpenShift clusters, provide the supported OpenShift version. The format is distribution dependent. For supported versions, see <a href="/vendor/testing-supported-clusters">Supported Compatibility Matrix Cluster Types (Beta)</a>.</td>
```
jdewinne, line 59: Remove Beta?

docs/reference/replicated-cli-cluster-update-ttl.md
jdewinne: It is possible to reduce also. For example: If the original TTL is 24hours. And the cluster is maybe running for an hour already, and the end user wants to reduce TTL to 12hours. That will work.
What will not work is that if for example the cluster is already running for 6 hours, to reduce the TTL to less than 6.

docs/reference/replicated-cli-cluster-update-ttl.md
```diff
@@ -0,0 +1,49 @@
...
+    <th width="20%">Type (if applicable)</th>
+    <th width="50%">Description</th>
+  </tr>
+  <Help/>
+  <Output/>
+  <tr>
+    <td><code>--ttl</code></td>
+    <td>string</td>
+    <td>
+    <p>The updated TTL for the cluster, in hours. <strong>Valid values:</strong> 1h through 48h. For example, to update the TTL to ten hours, use <code>--ttl 10h</code>.</p>
```
jdewinne, line 28: Min ttl is `10m`

docs/reference/replicated-cli-cluster-update-ttl.md
jdewinne: Your cluster was probably already running longer vs the updated ttl

paigecalvert: ahh makes sense

PR #1788, opened 2023-12-29
Update instance information

FourSigma:
Update instance information bullet points


AmberAlston (approved):
Looks good. Just need to update the screenshots when we get the chance

Issue #1839, opened 2024-01-10
Docs feedback on releases-creating-customer


AmberAlston:
Not urgent, and really this feedback is applicable to more than just this page, but it was this page that triggered the thought. 

In this section we walk you through how you can create and edit a customer within the UI, and we never mention that you can also take these same actions within the API. A customer didn't realize this was possible because they would have had to go digging within the v3 api docs to discover it. 

For any actions you can also do with the Replicated API (https://replicated-vendor-api.readme.io/reference/createapp) or the Replicated CLI (https://docs.replicated.com/reference/replicated-cli-installing) for that matter, we should consider at least minor mentions in the relevant sections that tend to be more "how to do it via the UI" heavy. 

Given we push automating many of these processes with the customer's CI/CD or other systems, this could help make the necessary info more discoverable especially to newer vendors/prospects

paigecalvert:
Thanks Amber, makes sense. Don't want to recreate the API docs throughout the product docs, but like you say, I think at least a description at the top of the page that says you can do this with UI, CLI, or API (with a link to the API & CLI topics) would be helpful and a pretty quick fix.

Looks like there's a handful of topics that this update would apply to so I can make a story 👍

paigecalvert:
https://app.shortcut.com/replicated/story/96248/add-links-to-relevant-cli-and-api-docs-for-vendor-tasks

PR #1846, opened 2024-01-11
Add links to cli and api docs

paigecalvert:
From https://github.com/replicatedhq/replicated-docs/issues/1839


AmberAlston (approved):
Looks great, thank you for such a fast turnaround on the feedback!

PR #1884, opened 2024-01-24
Add new diagrams

paigecalvert:


paigecalvert:
These are getting adding mostly as part of https://github.com/replicatedhq/replicated-docs/pull/1900.

Going to add the other diagrams as part of other stories

PR #1904, opened 2024-01-30
Add kots OS requirements

paigecalvert:


docs/enterprise/installing-general-requirements.md
```diff
@@ -55,6 +55,10 @@ To install the admin console on an existing cluster, the cluster must meet the f
 
     * **1GB for rqlite PersistentVolume**: The admin console requires 1GB for a rqlite StatefulSet to store version history, application metadata, and other small amounts of data needed to manage the application(s). During deployment, the rqlite component is secured with a randomly generated password, and only exposed as a ClusterIP on the overlay network.  
 
+* **Supported operating systems**: The following are the supported operating systems for nodes:
+  * Linux AMD64
+  * Linux ARM64
```
ajp-io, line 60: @sgalsaleh Do you think we should say "Linux (any distribution)" for operating system and then have a separate section for architectures and list AMD64 and ARM64?

sgalsaleh: I think this is fine for now.

paigecalvert: @ajp-io LMK if you're OK with leaving this as is for now ^. If so I can merge

PR #1997, opened 2024-02-13
chore: Update hyperlink

banjoh:
https://docs.replicated.com/vendor/packaging-embedded-kubernetes#include-a-kubernetes-installer-in-an-application-release-beta does not direct us to the correct place


paigecalvert:
thanks @banjoh !

PR #1999, opened 2024-02-13
[HOLD FOR MERGE] update docs with custom id information on create customer page

alicenstar:
Related story: https://app.shortcut.com/replicated/story/98722/docs-update-create-edit-customer-in-context-of-custom-id-option


docs/vendor/releases-creating-customer.md
```diff
@@ -31,6 +31,8 @@ To create a customer:
    <ChangeChannel/>
    :::
 
+1. For **Custom ID**, you can enter a custom ID for the customer. The custom ID is a unique identifier for the customer that you can use to track the customer in your own systems (ex: Salesforce). The custom ID is not used by Replicated for any purpose.
```
AmberAlston, line 34: We should be clear that while the idea is that it's unique on their side, we do not enforce uniqueness on the Replicated side. This currently reads to me like we do. 

maybe something like "Setting a custom ID allows you to easily tie this Replicated customer record to your own internal customer data systems during data exports. We suggest setting an alphanumeric value such as your Salesforce ID, Hubspot ID, etc for that customer. Replicated does not require the custom ID to be unique. The custom ID is for Vendor data reconciliation purposes, and is not used by Replicated for any functionality purposes.

alicenstar: @AmberAlston I updated the text here, let me know what you think (I just copied what you had here)!

AmberAlston (approved):

AmberAlston:
LGTM. Approved. Ship it when code is merged.

paigecalvert (approved):
thanks @alicenstar and @AmberAlston !

PR #2013, opened 2024-02-18
Update replicated-cli-cluster-create.md

xavpaice:


docs/reference/replicated-cli-cluster-create.md
```diff
@@ -83,6 +83,12 @@ replicated cluster create [flags]
   replicated cluster create --name eks-example --distribution eks --version 1.27 --node-count 3 --instance-type m6i.large
   ```
 
+- For an EKS distribution with an additional node group:
+
+  ```bash
+  replicated cluster create --name eks-nodegroup-example --distribution eks --instance-type m6i.large --nodes 1 --nodegroup name=arm,instance-type=m7g.large,nodes=1,disk=50
```
jdewinne, line 89: I would leave out the `name` as it is non required.

jdewinne (approved):

docs/reference/replicated-cli-cluster-create.md
xavpaice: agreed, though I could also leave --instance-type and --nodes if we're doing that.  I put that there for consistency with the other examples.

paigecalvert (approved):

PR #2075, opened 2024-03-05
Adding node autoscaling

jdewinne:


paigecalvert (approved):
@jdewinne I see this is still in Draft but approving in case it needs to be merged (I see this update went out in the release yesterday)

paigecalvert (approved):

PR #2107, opened 2024-03-11
Update ports in systems requirements 

diamonwiggins:
Since `1.17`, the ports for the `kube-controller-manager` and `kube-scheduler` have changed. See - https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.17.md#cluster-lifecycle-1

Upstream docs - https://kubernetes.io/docs/reference/networking/ports-and-protocols/

### DIFF
--- a/docs/enterprise/installing-general-requirements.md
+++ b/docs/enterprise/installing-general-requirements.md
@@ -258,7 +258,7 @@ To install KOTS in an embedded cluster created by kURL, your environment must me
 * 4 CPUs or equivalent per machine
 * 8GB of RAM per machine
 * 40GB of disk space per machine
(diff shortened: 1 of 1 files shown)


adamancini (commented):
lgtm

paigecalvert (dismissed):
thank you @diamonwiggins @adamancini !

diamonwiggins:
From working with a customer and recently doing an install of kURL, i've noticed that the ports for the controller-manager and scheduler are not opened on the primary interface of a node, but just on `localhost`, so I'm thinking that we can remove these ports altogether

laverya (approved):

paigecalvert (approved):

PR #2210, opened 2024-04-02
Update replicated-cli-installing.md

brianrhea:
The documentation says to replace `APP_SLUG` but the examples used `APP_NAME`


paigecalvert (approved):
Good catch thanks @brianrhea

PR #2213, opened 2024-04-03
Add repl template functions examples

paigecalvert:
https://deploy-preview-2213--replicated-docs-upgrade.netlify.app/reference/template-functions-examples


docs/reference/template-functions-about.mdx
```diff
@@ -1,25 +1,18 @@
...
+import SyntaxInteger from "../partials/template-functions/_syntax-integer.mdx"
+import UseCases from "../partials/template-functions/_use-cases.mdx"
 
 # About Template Functions
 
 This topic describes Replicated KOTS template functions, including information about use cases, template function contexts, syntax.
 
 ## Overview
 
-For standard Kubernetes manifests deployed with Replicated KOTS, Replicated provides a set of custom template functions based on the Go text/template library.
```
paigecalvert, line 12: ^ they also work in kots custom resources, which are not deployed

docs/reference/template-functions-about.mdx
```diff
@@ -48,52 +41,29 @@ Within valid YAML that is templated at the same time, you can use Go to assign a
...
-    </td>
-  </tr>
-  <tr>
-    <th>Example</th>
-    <td><SyntaxStringExample/>
-    </td>
-  </tr>
-</table>
+* [`repl{{ ... }}`](#syntax-integer)
+* [`{{repl ... }}` or `'{{repl ... }}'`](#syntax-string)
```
paigecalvert, line 46: ^ Showing both quoted and non-quoted for this syntax

docs/reference/template-functions-about.mdx
```diff
@@ -48,52 +41,29 @@ Within valid YAML that is templated at the same time, you can use Go to assign a
...
+
+### `{{repl ... }}` or `'{{repl ... }}'` {#syntax-string}
+
+This syntax renders string, boolean, or integer values and can be used at the middle or end of a value in YAML. This syntax supports being wrapped in quotes, if you need to the rendered result of the template function to be quoted. When wrapped in quotes, this syntax returns string values only. 
+
+To use this syntax at the beginning of a value in YAML, it _must_ be wrapped in quotes because you cannot start a YAML value with the `{` character. For example, `'{{repl ... }}'`.
+
+For example:
+
+<SyntaxStringExample/>
```
paigecalvert, line 66: ^ got rid of the the tables and just put all the content in plain text under the subheadings.

Updated the description to clarify that they can both render boolean, integer, and string values and added some use cases/limitations. Didn't put anything in there yet about recommending one over the other. It was hard to avoid updating these at all as part of this PR since it's helpful to link back and forth

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,385 @@
+import IntegerComparison from "../partials/template-functions/_integer-comparison.mdx"
+import StringComparison from "../partials/template-functions/_string-comparison.mdx"
+import GoSprig from "../partials/template-functions/_go-sprig.mdx"
+import UseCases from "../partials/template-functions/_use-cases.mdx"
+
+# KOTS Template Function Examples
```
paigecalvert, line 6: Added a variety of comparison, conditionals, and formatting examples. Tested all of these with different apps but could use a second (or third) pair of eyes to make sure I didn't miss something. 

For each example, I included:
* Which KOTS template functions are used
* Links to third party docs if available
* Some comments in the YAML to help clarify as needed
* Just the YAML file that shows the template functions syntax (ie, not including the HelmChart CR, the Helm values file, and the Config custom resource, for example). There might be an exception or two to this, but in general, it felt like a lot to include all the connected YAML files when this page is more "show me how the syntax works"
* Sometimes showed the result of the template functions, like in the Comparison section. Otherwise, sometimes explained what the result would be or just let the description of the example do the talking

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,385 @@
...
+        ports:
+          http:
+            enabled: true
+            # Render the HTTP port for the service depending on the user's selection
+            port: repl{{ if ConfigOptionEquals "ingress_type" "load_balancer" }} repl{{ ConfigOption "load_balancer_port" }} repl{{ else }} 8081 repl{{ end }}
+            protocol: HTTP
+            targetPort: 8081
+```
+
+#### Multi-Line in KOTS HelmChart Values
```
paigecalvert, line 172: Diamon and I couldn't sort out the formatting for the nested if else statement here: https://github.com/replicated-collab/git-guardian-kots/issues/198, so left it out. I just cut out the nested if else statement from the example and it worked fine. 

I did include a note in the If Else section above that says if you have more complex if else statements/nested statements, then you can also try doing the if else from your Helm templates, rather than with the replicated template functions (Diamon suggested that this might be a preferable method to control flow in a case like this anyway)

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,385 @@
...
+    {{repl else -}}
+      {{repl ConfigOption "external_postgres_user" | Base64Encode }}
+    {{repl end}}
+  DB_PASSWORD: >-
+    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}
+      {{repl ConfigOption "embedded_postgres_password" | Base64Encode }}
+    {{repl else -}}
+      {{repl ConfigOption "external_postgres_password" | Base64Encode }}
+    {{repl end}}
+```
```
paigecalvert, line 256: ^ pulled from an existing example in the docs. Just needed to add the `-` characters to the end of the if and else templates to trim space

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,385 @@
...
+## Formatting Examples
+
+This section includes examples of how to format the rendered output of KOTS template functions.
+
+In addition to the examples in this section, KOTS template functions in the Static context include several options for formatting values, such as converting strings to upper or lower case and trimming leading and trailing space characters. For more information, see [Static Context](/reference/template-functions-static-context). 
+
+### Indentation
+
+When using template functions within nested YAML, it is important that the rendered template functions are indented correctly so that the YAML renders. A common use case for adding indentation to KOTS template functions is when templating annotations for resources or objects deployed by your application based on user-supplied values, as shown in the examples below. 
+
```
paigecalvert, line 295: Pulled these indentation examples from Templating Annotations: https://docs.replicated.com/vendor/resources-annotations-templating

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,387 @@
...
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: example
+  annotations:
+    kots.io/placeholder: |-
+      repl{{ printf "my.custom/annotation.ingress.hostname: %s" (ConfigOption "ingress_hostname") | nindent 4 }}
+```
```
paigecalvert, line 347: ^ got this example from https://docs.replicated.com/vendor/resources-annotations-templating#map-annotations-from-multiple-configuration-fields

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,387 @@
...
+
+```yaml
+password: '{{repl ConfigOption “password” }}'
+```
+
+Result:
+
+```yaml
+password: “abc123”
+```
```
paigecalvert, line 377: ^ Simpler examples that don't include the surrounding YAML files. Alex added these here https://docs.google.com/document/d/1105K6LFGxpBOXu0DqEuhgTfYGLq4dtPwwbim1Ak0294/edit. Seemed fine since the full YAML context isn't really important in this case

docs/reference/template-functions-about.mdx
```diff
@@ -48,52 +41,29 @@ Within valid YAML that is templated at the same time, you can use Go to assign a
 
 ## Syntax {#syntax}
```
paigecalvert, line 42: Updated the descriptions in this section to clarify that both syntaxes can render boolean, integer, and string values and added some use cases/limitations. Didn't put anything in there yet about recommending one over the other.

It was hard to avoid updating these at all as part of this PR since it's helpful to link back and forth between the examples and this syntax description. I can also revert these changes for now if we'd rather wait for Salah

docs/vendor/tutorial-adding-db-config.md
```diff
@@ -686,33 +686,33 @@ metadata:
   name: postgres
 data:
   DB_HOST: >-
-    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" }}
+    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}
```
paigecalvert, line 689: Adding the -'s because it didn't render properly without them in my testing

docs/vendor/tutorial-adding-db-config.md
```diff
@@ -951,20 +951,3 @@ spec:
 ### Integrating a Real Database
 
 If you'd like at this point, you can integrate a real database in your environment, just fill out your configuration fields. You'll know you did it right if your pg-consumer pod can connect.
-
-
-<!-- Coming Soon!
```
paigecalvert, line 956: Exhibit A why "coming soon" content doesn't belong in the docs site ...

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,388 @@
...
+# KOTS HelmChart custom resource
+apiVersion: kots.io/v1beta2
+kind: HelmChart
+metadata:
+  name: samplechart
+spec:
+  values:
+    replicated-sdk: >-
+      {{repl if HasLocalRegistry -}}
+        {{repl LocalRegistryHost }}/{{repl LocalRegistryNamespace }}/replicated-sdk:v1.0.0-beta.14
```
ajp-io, line 195: Might as well use the latest!
```suggestion
        {{repl LocalRegistryHost }}/{{repl LocalRegistryNamespace }}/replicated-sdk:v1.0.0-beta.17
```

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,388 @@
...
+      {{repl else -}}
+        docker.io/replicated/replicated-sdk:v1.0.0-beta.14
```
ajp-io, line 197:
```suggestion
        docker.io/replicated/replicated-sdk:v1.0.0-beta.17
```

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,388 @@
...
+      {{repl end}}
+```
+
+Given the example above, if the user is _not_ using a local registry, then the `replicated-sdk` value in the Helm chart is set to the location of the image on the default docker registry, as shown below:
+
+```yaml
+# Helm chart values file
+
+replicated-sdk: 'docker.io/replicated/replicated-sdk:v1.0.0-beta.14'
```
ajp-io, line 206:
```suggestion
replicated-sdk: 'docker.io/replicated/replicated-sdk:v1.0.0-beta.17'
```

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,388 @@
...
+    {{repl else -}}
+      {{repl ConfigOption "external_postgres_password" | Base64Encode }}
+    {{repl end}}
+```
+
+### Ternary Operators
+
+Ternary operators are useful for templating strings where certain values within the string must be rendered differently depending on a given condition. Compared to if else statements, ternary operators are useful when a small portion of a string needs to be conditionally rendered, as opposed to rendering different values based on a conditional statement. For example, a common use case for ternary operators is to template the path to an image repository based on user-supplied values.
+
+The following example uses ternary operators to render the registry and repository for an nginx image depending on if a local image regsitry is used. This example uses the following KOTS template functions:
```
ajp-io, line 262:
```suggestion
The following example uses ternary operators to render the registry and repository for a private nginx image depending on if a local image regsitry is used. This example uses the following KOTS template functions:
```

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,387 @@
...
+
+The following example shows templating annotations for an Ingress object. This example uses the KOTS [ConfigOption](/reference/template-functions-config-context#configoption) template function to return user-supplied annotations from the admin console **Config** page. It also uses [nindent](https://masterminds.github.io/sprig/strings.html) to indent the rendered value four spaces.
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: example-ingress
+  annotations:
+    kots.io/placeholder: |-
```
ajp-io, line 325: Unrelated to this, but I think we should improve the `kots.io/placeholder` docs. I assume that `kots.io/placeholder` goes away (hence, placeholder), but that isn't explicitly said, and I'm not even positive. I'd have to test it. So I think saying something about that over there would be helpful.

docs/reference/template-functions-examples.mdx
ajp-io: Might need an engineer to confirm, but using the placeholder here doesn't feel necessary. I feel like this would work, but `kots.io/placeholder` isn't something I'm that familiar with.
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: example
  annotations:
    my.custom/annotation.ingress.hostname: repl{{ ConfigOption "ingress_hostname" }}
```

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,387 @@
...
+### Render Quoted Values
+
+To wrap a rendered value in quotes, you can pipe the result from KOTS template functions with the `repl{{ ... }}` syntax into quotes using `| quote`. Or, you can use the `'{{repl ... }}'` syntax instead.
+
+One use case for quoted values in YAML is when indicator characters are included in values. In YAML, indicator characters (`-`, `?`, `:`) have special semantics and must be escaped if used in values. For more information, see [Indicator Charactors](https://yaml.org/spec/1.2.2/#53-indicator-characters) in the YAML documentation.
+
+#### Example with `| quote`
+
+```yaml
+password: repl{{ ConfigOption “password” | quote }}
```
ajp-io, line 357: Would be good to use an example with an indicator character, since this would work without quotation marks. Like try something that fails without `quote`.

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,385 @@
...
+
+### If Else Statements
+
+A common use case for if else statements with KOTS template functions is to set values for resources or objects deployed by your application, such as custom annotations or service types, based on user-specific data.
+
+This section includes examples of both single line and multi-line if else statements. Using multi-line formatting can be useful to improve the readability of YAML files when longer or more complex if else statements are needed.
+
+Multi-line if else statements can be constructed using YAML block scalars and block chomping characters to ensure the rendered result is valid YAML. A _folded_ block scalar style is denoted using the greater than (`>`) character. With the folded style, single line breaks in the string are treated as a space. Additionally, the block chomping minus (`-`) character is used to remove all the line breaks at the end of a string. For more information about working with these characters, see [Chapter 8: Block Style Productions](https://yaml.org/spec/1.2.2/#chapter-8-block-style-productions) in the YAML documentation.
```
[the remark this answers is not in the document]
ajp-io: I don't know either, so someone else can confirm

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,385 @@
...
+    {{repl end}}
+  DB_PASSWORD: >-
+    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}
+      {{repl ConfigOption "embedded_postgres_password" | Base64Encode }}
+    {{repl else -}}
+      {{repl ConfigOption "external_postgres_password" | Base64Encode }}
+    {{repl end}}
+```
+
+### Ternary Operators
```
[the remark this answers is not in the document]
ajp-io: I might be able to if I mess around with it, but going to submit my review so you can look stuff over.

ajp-io (commented):
Looks good! I wrote a few comments about examples, and I think the syntaxes part needs more examples, but maybe that's for another PR.

I'm interested what an engineer says too. I've seen several customers create certificates with template functions. I wonder if that's common enough to warrant an example. Or what other examples are common enough that an engineer would think of them and say, document this too.

Overall this seems like a great improvement, and I learned a few things.

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,388 @@
...
+      title: Conditional Item
+      type: text
+      # Display this item only when the customer enables the option_one config field *and*
+      # has the feature-1 entitlement in their license
+      when: repl{{ and (LicenseFieldValue "feature-1" | ParseBool) (ConfigOptionEquals "select_one_example" "option_one")}}  
+```
+
+### Integer Comparison
+
+Integer values can be compared using operators such as greater than, less than, equal to, and so on. Because many KOTS template functions return string values, working with integer values often requires using another function, such as the KOTS [ParseInt](/reference/template-functions-static-context#parseint) template function or the Sprig [atoi](https://masterminds.github.io/sprig/conversion.html) function, to return the integer represented by the string.
```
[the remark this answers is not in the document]
cbodonnell: Our `ParseInt` function uses the [strconv.ParseInt](https://pkg.go.dev/strconv#ParseInt) function, so you _could_ specify that you want to parse the integer with a different `base` value. If you don't pass a second argument to it, we will default to base 10 ([code](https://github.com/replicatedhq/kots/blob/v1.108.5/pkg/template/static_context.go#L224-L231)).

And according to the docs for [strconv.Atoi](https://pkg.go.dev/strconv#Atoi):
> Atoi is equivalent to ParseInt(s, 10, 0), converted to type int.

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,387 @@
...
+
+```yaml
+# KOTS HelmChart custom resource
+apiVersion: kots.io/v1beta2
+kind: HelmChart
+metadata:
+  name: samplechart
+spec:
+  values:
+    replicated-sdk: >-
```
cbodonnell, line 193: I know this is just a minimal example, but the actual value to set in this use-case would be `images.replicated-sdk`. Think it's worth doing that here so that if someone copy-pastes this snippet, it's valid?

reference: https://github.com/replicatedhq/replicated-sdk/blob/v1.0.0-beta.17/chart/values.yaml.tmpl#L5-L6

ajp-io: Definitely should match. Good catch.

docs/reference/template-functions-examples.mdx
cbodonnell: Tested this out and it looks like it remains, but is empty. I don't see any place in our code where we do anything with it, so I think this is just a trick to make this kind of templating work with yaml.

docs/reference/template-functions-examples.mdx
cbodonnell: right, placeholder is more for that use-case above where the annotation key and value would come from the result of a template function since just a template function by itself makes the yaml invalid

docs/vendor/tutorial-adding-db-config.md
cbodonnell: When used with go templates, it will strip out whitespace on the side that it's on https://pkg.go.dev/text/template#hdr-Text_and_spaces

I'm guessing that without it may be rendering the newlines and making it invalid?

docs/reference/template-functions-examples.mdx
ajp-io: Do you mean the result is like this?
```yaml
  annotations:
    kots.io/placeholder: ""
    from: the-config
```

ajp-io: From elsewhere in the docs, maybe it's this
```yaml
kots.io/placeholder: |-
```
But is `|-` a string? I'm confused by that.

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,388 @@
...
+  chart:
+    name: my-app
+    chartVersion: 0.23.0
+  values:
+    services:
+      my-service:
+        enabled: true
+        appName: ["my-app"]
+        # Render the service type based on the user's selection
+        type: repl{{ if ConfigOptionEquals "ingress_type" "load_balancer" }} LoadBalancer repl{{ else }}ClusterIP repl{{ end }}
```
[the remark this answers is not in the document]
cbodonnell: This use-case is what makes me think of the other `{{repl` syntax since it could be more readable without needing to add extra spaces since you may not be able to do that in all situations. Might not be the case here, but this is when I find myself using it.
```suggestion
        type: '{{repl if ConfigOptionEquals "ingress_type" "load_balancer" }}LoadBalancer{{repl else }}ClusterIP{{repl end }}'
```

docs/reference/template-functions-examples.mdx
cbodonnell: mine rendered like this, with it as an empty multiline string:
```yaml
  annotations:
    kots.io/placeholder: |-
      
    example.com/annotation: some-value
```

ajp-io: I see. So the blank newline is just part of the value for `kots.io/placeholder`, and then the annotation is completely outside that string.

docs/reference/template-functions-examples.mdx
ajp-io: I'm having trouble thinking of a better example, but something where we concatenate a config value onto some existing string.

docs/vendor/tutorial-adding-db-config.md
paigecalvert: Yeah exactly, it rendered with extra newlines

paigecalvert: I can repeat the tests with a variety of `-`s and share the results

docs/reference/template-functions-examples.mdx
paigecalvert: Good call, I can use a `key: "key: value"` example like the one you added to the google doc, unless you can think of a better example

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,388 @@
...
+      - name: option_one
+        title: Option One
+      - name: option_two
+        title: Option Two  
+    - name: conditional_item
+      title: Conditional Item
+      type: text
+      # Display this item only when the customer enables the option_one config field *and*
+      # has the feature-1 entitlement in their license
+      when: repl{{ and (LicenseFieldValue "feature-1" | ParseBool) (ConfigOptionEquals "select_one_example" "option_one")}}
```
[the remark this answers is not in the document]
paigecalvert: Yeah it was a bool, and to be fair I didn't test it _without_ ParseBool to see if that worked too, so I can give that a shot.

I was sort of just assuming this would only apply for boolean type fields, though I guess there could be a use case for wanting to check that a string/integer license field is present without necessarily caring about what the value is...I'm just not sure how ParseBool could take a value like "potato" and return a boolean from that, but I'll test this too!

docs/reference/template-functions-examples.mdx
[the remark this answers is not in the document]
paigecalvert: Thanks Craig; I'll add in this extra detail

docs/reference/template-functions-examples.mdx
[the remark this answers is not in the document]
paigecalvert: @cbodonnell , this question is more YAML-specific, but do you know if this is an accurate description of these characters and how they're used with multi-line YAML?

docs/reference/template-functions-examples.mdx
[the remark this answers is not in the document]
paigecalvert: Makes sense, it does look easier to read. I'll use this syntax instead (good to show a mix of syntax use cases anyways)

docs/reference/template-functions-examples.mdx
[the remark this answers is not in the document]
ajp-io: Not sure I'm reading your response correctly, so sorry if this is off base or is stuff you already know. `ParseBool` turns a string value like `"true"` or `"false"` into a boolean `true` or `false`. So if the value is already an unquoted boolean, you wouldn't need to use `ParseBool`. That's really what I was getting at. If the license returns `true` for a boolean license field value, then `ParseBool` is unnecessary. But I don't know if the license returns it as a string or a bool.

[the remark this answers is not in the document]
paigecalvert: Okay yeah I'm with ya. 

Looks like without ParseBool, it always returns true (which I think would just be evaluating "is there a licensefield with this name, regardless of what it's value is").

To test, I made bool, string, and int license fields. I set the bool field to False, the string field to "false" and the int field to `0`.

The logical `and` statement is expected to be TRUE when the license field is TRUE AND the boolean config option is FALSE.

Since I set all the license fields to FALSE, all the logical AND conditionals should also evaluate to FALSE.

Upstream configmap:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  config-field-bool: '{{repl ConfigOptionEquals "bool_item" "0"}}'
  license-field-bool: '{{repl LicenseFieldValue "bool" }}'
  license-field-string: '{{repl LicenseFieldValue "string" }}'
  license-field-integer: '{{repl LicenseFieldValue "Integer" }}'
  bool-logical-and-parsebool: '{{repl and (LicenseFieldValue "bool" | ParseBool) (ConfigOptionEquals "bool_item" "0")}}'
  bool-logical-and-no-parsebool: '{{repl and (LicenseFieldValue "bool") (ConfigOptionEquals "bool_item" "0")}}'
  string-logical-and-parsebool: '{{repl and (LicenseFieldValue "string" | ParseBool) (ConfigOptionEquals "bool_item" "0")}}'
  string-logical-and-no-parsebool: '{{repl and (LicenseFieldValue "string") (ConfigOptionEquals "bool_item" "0")}}'
  integer-logical-and-parsebool: '{{repl and (LicenseFieldValue "Integer" | ParseBool) (ConfigOptionEquals "bool_item" "0")}}'
  integer-logical-and-no-parsebool: '{{repl and (LicenseFieldValue "Integer") (ConfigOptionEquals "bool_item" "0")}}'   
```

Rendered configmap:

```
kind: ConfigMap
apiVersion: v1
metadata:
  annotations:
    kots.io/app-slug: repl-template-functions
  labels:
    kots.io/app-slug: repl-template-functions
    kots.io/backup: velero
  name: test
data:
  config-field-bool: "true"
  license-field-bool: "false"
  license-field-integer: "0"
  license-field-string: "false"
  bool-logical-and-parsebool: "false"
  bool-logical-and-no-parsebool: "true"
  integer-logical-and-parsebool: "false"
  integer-logical-and-no-parsebool: "true"
  string-logical-and-parsebool: "false"
  string-logical-and-no-parsebool: "true"
```

In the rendered configmap, the comparisons that exclude the ParseBool function always return true, even though the license field values are all false. The same result applies regardless of the type of license field (bool, string, or int)

docs/vendor/tutorial-adding-db-config.md
paigecalvert: Okay reporting back. 

Here's the upstream configmap, which includes a `-` at the end of the if and else template functions:
```
data:
  DB_USER: >-
    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}
      {{repl Base64Encode "postgres" }}
    {{repl else -}}
      {{repl ConfigOption "external_postgres_user" | Base64Encode }}
    {{repl end}}
  DB_PASSWORD: >-
    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}
      {{repl ConfigOption "embedded_postgres_password" | Base64Encode }}
    {{repl else -}}
      {{repl ConfigOption "external_postgres_password" | Base64Encode }}
    {{repl end}}
```
And here's the rendered result:
```
data:
  DB_PASSWORD: cGFpZ2U=
  DB_USER: cG9zdGdyZXM=
```
Here's an upstream configmap without the `-`s at the end of if and else:
```
data:
  DB_USER: >-
    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" }}
      {{repl Base64Encode "postgres" }}
    {{repl else }}
      {{repl ConfigOption "external_postgres_user" | Base64Encode }}
    {{repl end}}
  DB_PASSWORD: >-
    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" }}
      {{repl ConfigOption "embedded_postgres_password" | Base64Encode }}
    {{repl else }}
      {{repl ConfigOption "external_postgres_password" | Base64Encode }}
    {{repl end}}
```

and here's the rendered result of that:

```
data:
  DB_PASSWORD: |2-

    cGFpZ2U=
  DB_USER: |2-

    cG9zdGdyZXM=
```

Not entirely sure what `|2-` is coming from, but it appears that the `-` at the end of ifs and elses trims this

paigecalvert: A couple other examples:

```yaml
data:
  # include -'s before and after if, else, and end
  DB_USER: >-
    {{- repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}
      {{repl Base64Encode "postgres" }}
    {{- repl else -}}
      {{repl ConfigOption "external_postgres_user" | Base64Encode }}
    {{- repl end -}}
  # include -'s after if, else, and end
  DB_PASSWORD: >-
    {{repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}
      {{repl ConfigOption "embedded_postgres_password" | Base64Encode }}
    {{repl else -}}
      {{repl ConfigOption "external_postgres_password" | Base64Encode }}
    {{repl end -}}
  # include this field to show how it's going to get sucked onto the same line as DB_PASSWORD when rendered
  field-below-DB_PASSWORD: 'string'   
```

```yaml
data:
  # it thinks that the field below is part of the value of DB_PASSWORD when there's a - at the end of the end statement 
  DB_PASSWORD: 'cGFpZ2U= field-below-DB_PASSWORD: ''string''    '
  # nothing good happened with DB_USER
  DB_USER: |-
    {{- repl if ConfigOptionEquals "postgres_type" "embedded_postgres" -}}

      cG9zdGdyZXM=
    {{- repl else -}}

      cGFpZ2U=
    {{- repl end -}}
```

docs/reference/template-functions-examples.mdx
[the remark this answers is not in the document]
paigecalvert: Also, just created a PR to clarify that `when` property stuff. Will ping you separately for a review https://github.com/replicatedhq/replicated-docs/pull/2257

docs/reference/template-functions-examples.mdx
paigecalvert: Yeah I see what you mean; I'm having a hard time thinking of any examples either that couldn't just be achieved by putting some strings and repl template functions next to each other

docs/reference/template-functions-about.mdx
paigecalvert: True, I like that suggestion

docs/reference/template-functions-examples.mdx
paigecalvert: I ended up just removing this example for now since I couldn't think of a good use case. We can always add something like this back in as it comes up (plus I expect we'll keep adding to this topic in general)

ajp-io: That's fair. I don't think there's anything `printf` can do that you can't do by concatenating stuff together yourself. So this is a convenience more than anything else.

docs/reference/template-functions-examples.mdx
paigecalvert: Made a story for looking into improvements for the placeholder docs https://app.shortcut.com/replicated/story/102880/improve-placeholder-docs

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,430 @@
...
+customTag: repl{{ ConfigOption "tag" | quote }}
+```
+
+The result for both examples is:
+
+```yaml
+customTag: 'key: value'
+```
+
+## Variables Example
```
paigecalvert, line 363: I add a new Variables section since the previous review.

This is because I came across this example that's currently sitting at the bottom of the Config Context page in the docs: https://docs.replicated.com/reference/template-functions-config-context#tls-cert-example

I always forget about it because it's so hard to find. I figured it could fit well here instead. As part of this, I also rewrote the description, because I found it very hard to understand previously

I went with "variables" as the category since I think that's the most broadly applicable part of this example (ie, anyone wondering how to use vars in the config custom resource could reference this)

docs/reference/template-functions-about.mdx
```diff
@@ -38,63 +27,100 @@ You can use KOTS template functions in any Kubernetes manifest files for applica
...
 
-During application installation and upgrade, KOTS templates all Kubernetes manifest files in a release (except for the Config custom resource) at the same time during a single process. For the Config custom resource, KOTS templates each array of items separately. For information about the syntax of items in the Config custom resource, see [items](/reference/custom-resource-config#items) in _Config_.
+During application installation and upgrade, KOTS templates all Kubernetes manifest files in a release (except for the Config custom resource) at the same time during a single process.
 
-Within valid YAML that is templated at the same time, you can use Go to assign a template function to a variable for the purpose of accessing the rendered value from another template function. For an example of using Go variables with KOTS template functions to generate and then access values, see [Example: Using Variables to Generate TLS Certificates and Keys](/reference/template-functions-config-context#tls-cert-example).
+For the Config custom resource, KOTS templates each item separately. For information about the Config custom resource, see [Config](/reference/custom-resource-config).
```
paigecalvert, line 36: ^ I noticed that the statement "For the Config custom resource, KOTS templates each array of items separately" conflicted with the statement "Replicated evaluates and templates each item in the Config custom resource separately" [here](https://docs.replicated.com/reference/template-functions-config-context#tls-cert-example)

Dmitriy confirmed that it's every item separate. See https://replicated.slack.com/archives/C036BTS7JCE/p1713210501472269 

I also pulled out a lot of the language about variables here because it was just too confusing. Maybe we can continue to add more content and example about working with variables with template functions, but probably doesn't fit in this section

docs/reference/template-functions-about.mdx
```diff
@@ -38,63 +27,100 @@ You can use KOTS template functions in any Kubernetes manifest files for applica
...
-    <td><p>This syntax uses the delimiter after <code>repl</code>, and does not require quotation marks.
-    Use this syntax to render integer values.</p></td>
-  </tr>
-  <tr>
-    <th>Example</th>
-    <td><SyntaxIntegerExample/>
-    </td>
-  </tr>
-</table>
+This syntax is recommended for most use cases.
```
paigecalvert, line 46: Updated this section with more examples and simplified descriptions explaining the use cases

docs/reference/template-functions-about.mdx
```diff
@@ -38,63 +27,100 @@ You can use KOTS template functions in any Kubernetes manifest files for applica
...
+```yaml
+links:
+  - description: "Click Me"
+    url: 'https://{{repl ConfigOption "hostname"}}:{{repl ConfigOption "port"}}'
+```
+```yaml
+links:
+  - description: "Click Me"
+    url: https://example.com:8888
+```
```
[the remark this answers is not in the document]
paigecalvert: Also, not sure if the heading "Example Without Quotes" makes the most sense here. It's true that the template functions aren't wrapped in quotes, even though the value is, but not sure if that's confusing

docs/reference/template-functions-examples.mdx
ajp-io: This is a great addition. Doesn't have to block this, but it would be helpful to explain some of what's going on here. Like at least link to the template functions we use here, like we do for other examples. I didn't know what `dig` did off the top of my head, and without reading this very closely, I don't know how a few steps work.

paigecalvert: Can do, looks like I did forget to talk about the dictionary sprig functions

docs/reference/template-functions-about.mdx
paigecalvert: Sorry, not sure why this old comment of mine is still here. Looks like it was from last week

docs/reference/template-functions-about.mdx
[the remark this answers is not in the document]
paigecalvert: Yeah agreed...maybe this example can just go away for this iteration. I'd rather not introduce more confusion and maybe it's something that needs more investigation

docs/vendor/config-screen-conditional.mdx
```diff
@@ -22,40 +24,9 @@ For additional examples of using conditional statements in the Config custom res
 
 It can be useful to show or hide configuration fields depending on the distribution of the cluster because different distributions often have unique requirements.
 
-In the following example, the `when` properties use the Distribution template function to return the Kubernetes distribution of the cluster where Replicated KOTS is running. If the distribution of the cluster matches the specified distribution, then the `when` property evaluates to true. For more information about the Distribution template function, see [Distribution](/reference/template-functions-static-context#distribution) in _Static Context_.
+In the following example, the when properties use the Distribution template function to return the Kubernetes distribution of the cluster where Replicated KOTS is running. If the distribution of the cluster matches the specified distribution, then the when property evaluates to true. For more information about the Distribution template function, see [Distribution](/reference/template-functions-static-context#distribution) in _Static Context_.
```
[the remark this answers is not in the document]
paigecalvert: Sure didn't

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,438 @@
...
+
+### If-Else Statements
+
+A common use case for if-else statements with KOTS template functions is to set values for resources or objects deployed by your application, such as custom annotations or service types, based on user-specific data.
+
+This section includes examples of both single line and multi-line if-else statements. Using multi-line formatting can be useful to improve the readability of YAML files when longer or more complex if-else statements are needed.
+
+Multi-line if-else statements can be constructed using YAML block scalars and block chomping characters to ensure the rendered result is valid YAML. A _folded_ block scalar style is denoted using the greater than (`>`) character. With the folded style, single line breaks in the string are treated as a space. Additionally, the block chomping minus (`-`) character is used to remove all the line breaks at the end of a string. For more information about working with these characters, see [Block Style Productions](https://yaml.org/spec/1.2.2/#chapter-8-block-style-productions) in the YAML documentation.
```
paigecalvert, line 143: Asked in eng AMA channel to make sure this is accurate: https://replicated.slack.com/archives/C036BTS7JCE/p1713370246927409

docs/reference/template-functions-about.mdx
paigecalvert: But yeah good point, that one probably is strings only

paigecalvert: Okay, maybe a weird counter example, but:
```
spec:
  ports:
    - port: 808{{repl ConfigOption "someinteger"}}
```
rendered:
```
spec:
  ports:
  - port: 8081
```

ajp-io: Then I think you're right :)

docs/reference/template-functions-about.mdx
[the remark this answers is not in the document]
ajp-io: Sounds good. It would be good to follow up on, because this does somewhat confuse my understanding of when to use which. But for now, I think our guidance is still valid.

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,438 @@
...
+* KOTS [ConfigOption](/reference/template-functions-config-context#configoption) template function to render the user-supplied value for the ingress hostname. This is passed as a parameter to the [genCA](https://masterminds.github.io/sprig/crypto.html) and [genSignedCert](https://masterminds.github.io/sprig/crypto.html) functions 
+* Sprig [genCA](https://masterminds.github.io/sprig/crypto.html) and [genSignedCert](https://masterminds.github.io/sprig/crypto.html) functions to generate a CA and a certificate signed by the CA
+* Sprig [dict](https://masterminds.github.io/sprig/dicts.html), [set](https://masterminds.github.io/sprig/dicts.html), and [dig](https://masterminds.github.io/sprig/dicts.html) dictionary functions to create a dictionary with entries for both the CA and the certificate, then traverse the dictionary to return the values of the CA, certificate, and key.
+* [toJson](https://masterminds.github.io/sprig/defaults.html) and [fromJson](https://masterminds.github.io/sprig/defaults.html) Sprig functions to encode the CA and certificate into a JSON string, then decode the JSON for the purpose of displaying the values on the **Config** page as defaults
+
```
paigecalvert, line 380: ^ Added this list of which functions are used in the example and for what purpose @ajp-io 

LMK if this addresses what you were thinking

docs/reference/template-functions-examples.mdx
```diff
@@ -0,0 +1,438 @@
...
+          default: |-
+            repl{{ $ca := genCA (ConfigOption "ingress_hostname") 365 }}
+            repl{{ $tls := dict "ca" $ca }}
+            repl{{ $cert := genSignedCert (ConfigOption "ingress_hostname") (list ) (list (ConfigOption "ingress_hostname")) 365 $ca }}
+            repl{{ $_ := set $tls "cert" $cert }}
+            repl{{ toJson $tls }}
+        - name: tls_ca
+          title: Signing Authority
+          type: textarea
+          default: repl{{ fromJson (ConfigOption "tls_json") | dig "ca" "Cert" "" }}
```
ajp-io, line 412: Should this be "ca" instead of "Cert"?

paigecalvert: Huh not sure, good question.

Looking at the [dig](https://masterminds.github.io/sprig/dicts.html) docs, it seems that dig is looking through nested keys in the dictionary (in this case, ca > Cert, cert > Cert, and cert > Key). So I wonder if theoretically this might look like:
```
{
  ca: {
    Cert: {
         [the CA certificate]
    }
  }
  cert: {
    Cert: {
      [the signed cert]
    }
    Key: {
      [the key for the signed cert]
    }
  }
}
``` 
Obviously not actually looking at the JSON output, but could explain why both are named Cert if the Cert fields are nested

ajp-io: That might be true. This is why in the future it could even be nice to have line by line comments to show what this does. I can see someone not fully understanding this example without playing with it themselves (I fall into that camp).

But it's fine for now.

paigecalvert: Yeah I'm with you, it's not always obvious how this stuff works. There might be an opportunity to create community posts out of some of these examples too if they step-by-step examples become too long to fit

cbodonnell (approved):

PR #2226, opened 2024-04-07
Replicated SDK instance-tags API

FourSigma:
Add descriptions of the new `/app/instance-tags` endpoint on Replicated SDK

### DIFF
--- a/docs/reference/replicated-sdk-apis.md
+++ b/docs/reference/replicated-sdk-apis.md
@@ -92,7 +92,7 @@ Response:
 
 ### POST /app/custom-metrics
 
-Send custom application metrics. For more information and examples see [Configuring Custom Metrics](/vendor/custom-metrics)
+Send custom application metrics. For more information and examples see [Configuring Custom Metrics](/vendor/custom-metrics).
 
 
 ### POST /app/instance-tags
--- a/docs/reference/replicated-sdk-apis.md
+++ b/docs/reference/replicated-sdk-apis.md
@@ -100,7 +100,7 @@ Programmatically set new instance tags or overwrite existing tags. Instance tags
 
 Setting a tag with the `name` key will set the instance's name in the vendor portal.
 
-The `force` parameter defaults to `false`. If `force` is `false`, pre-existing tags will not be overwritten and the existing tags take precedence. If the `force` parameter is set to `true`, any pre-existing tags will be overwritten.
+The `force` parameter defaults to `false`. If `force` is `false`, conflicting pre-existing tags will not be overwritten and the existing tags take precedence. If the `force` parameter is set to `true`, any conflicting pre-existing tags will be overwritten.
 
 To delete a particular tag, set the key's value to an empty string `""`. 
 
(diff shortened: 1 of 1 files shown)


docs/reference/replicated-sdk-apis.md
```diff
@@ -92,7 +92,39 @@ Response:
 
 ### POST /app/custom-metrics
 
-Send custom application metrics. For more information and examples see [Configuring Custom Metrics](/vendor/custom-metrics)
+Send custom application metrics. For more information and examples see [Configuring Custom Metrics](/vendor/custom-metrics).
```
ajp-io, line 95: Unrelated to this PR, but we should have a request and response for this too, even if we have more detailed info elsewhere.

docs/reference/replicated-sdk-apis.md
```diff
@@ -92,7 +92,38 @@ Response:
...
+
+### POST /app/instance-tags
+
+Programmatically set new instance tags or overwrite existing tags. Instance tags are key-value pairs, where the key and the value are strings.
+
+Setting a tag with the `name` key will set the instance's name in the vendor portal.
+
+The `force` parameter defaults to `false`. If `force` is `false`, pre-existing tags will not be overwritten and the existing tags take precedence. If the `force` parameter is set to `true`, any pre-existing tags will be overwritten.
```
ajp-io, line 103: I simplified this a lot, but I think it still gets the point across. I found the old wording a bit confusing (e.g., how often is "periodically"?).

docs/reference/replicated-sdk-apis.md
```diff
@@ -92,7 +92,38 @@ Response:
...
 ### POST /app/custom-metrics
 
-Send custom application metrics. For more information and examples see [Configuring Custom Metrics](/vendor/custom-metrics)
+Send custom application metrics. For more information and examples see [Configuring Custom Metrics](/vendor/custom-metrics).
+
+### POST /app/instance-tags
+
+Programmatically set new instance tags or overwrite existing tags. Instance tags are key-value pairs, where the key and the value are strings.
+
+Setting a tag with the `name` key will set the instance's name in the vendor portal.
```
ajp-io, line 101: Added a note about this so people know how to set instance name

docs/reference/replicated-sdk-apis.md
```diff
@@ -92,7 +92,38 @@ Response:
...
+
+```json
+{
+  "data": {
+    "force": "false",
+    "tags": {
+      "name": "my-instance-name",
+      "preExistingKey": "will-not-be-overwritten",
+      "cpuCores": "10",
+      "supportTier": "basic",
```
ajp-io, line 120: Removed a few tags for simplicity and made `"preExistingKey": "will-not-be-overwritten"` to show it won't be overwritten.

cbodonnell (dismissed):

AmberAlston:
Minor update to better match modal wording

docs/reference/replicated-sdk-apis.md
AmberAlston: I'm leaning towards removing this `preExistingKey` example. I'm concerned readers will get confused they have to pass this to not have things overwritten. Seems silly that someone might read it that way, but we've had folks get confused in the past that "{my app}" doesn't work in our example cli/api commands.

AmberAlston (approved):

paigecalvert (approved):

PR #2254, opened 2024-04-12
Updating GitOps messaging

AmberAlston:
This PR clarifies our messaging around use of GitOps with KOTS, which is no longer recommended. This PR resulted from a Product Team discussion on April 12 2024.


paigecalvert (approved):
Thanks @AmberAlston !

Minor edits:
* Moved the statement to an "important" admonition in a partial
* Removed another mention of GitOps from the Installation Overview topic

PR #2257, opened 2024-04-12
Update requirements and limitations of config when property

paigecalvert:
https://deploy-preview-2257--replicated-docs.netlify.app/reference/custom-resource-config#when

### DIFF
--- a/docs/partials/config/_property-when.mdx
+++ b/docs/partials/config/_property-when.mdx
@@ -1,6 +1,4 @@
-The `when` property in the Config custom resource denotes configuration groups or items that are displayed on the admin console **Config** page only when a condition evaluates to true. When the condition evaluates to false, the group or item is not displayed. It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application.
-
-Conditional statements in the `when` property can be used to evaluate the user's environment, license entitlements, and configuration choices. For example:
+It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application. Conditional statements in the `when` property can be used to evaluate the user's environment, license entitlements, and configuration choices. For example:
 * The Kubernetes distribution of the cluster
 * If the license includes a specific feature entitlement
 * The number of users that the license permits
--- a/docs/partials/config/_property-when.mdx
+++ b/docs/partials/config/_property-when.mdx
@@ -1,6 +1,6 @@
-The `when` property in the Config custom resource denotes configuration items that are displayed on the admin console **Config** page only when a condition evaluates to true. When the condition evaluates to false, the field is not displayed. It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application.
+The `when` property in the Config custom resource denotes configuration groups or items that are displayed on the admin console **Config** page only when a condition evaluates to true. When the condition evaluates to false, the group or item is not displayed. It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application.
 
-You can conditionally show or hide configuration fields based on the user's environment, license entitlements, and configuration choices. For example, conditional statements in the `when` property can be used to evaluate:
+Conditional statements in the `when` property can be used to evaluate the user's environment, license entitlements, and configuration choices. For example:
 * The Kubernetes distribution of the cluster
 * If the license includes a specific feature entitlement
 * The number of users that the license permits
--- a/docs/partials/config/_property-when.mdx
+++ b/docs/partials/config/_property-when.mdx
@@ -1,4 +1,4 @@
-It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application. Conditional statements in the `when` property can be used to evaluate the user's environment, license entitlements, and configuration choices. For example:
+It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application. Conditional statements in the `when` property can be used to evaluate things like the user's environment, license entitlements, and configuration choices. For example:
 * The Kubernetes distribution of the cluster
 * If the license includes a specific feature entitlement
 * The number of users that the license permits
--- a/docs/partials/config/_property-when.mdx
+++ b/docs/partials/config/_property-when.mdx
@@ -1,7 +1,9 @@
-The `when` property denotes conditional user inputs that are only displayed on the admin console Config page when a condition evaluates to true. The equality check in the conditional statement of the `when` property must match exactly, without quotes.
+The `when` property in the Config custom resource denotes configuration items that are displayed on the admin console **Config** page only when a condition evaluates to true. When the condition evaluates to false, the field is not displayed. It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application.
 
-The `when` property can be used as a group or item property. The `select_one` item type can be used as an option under the `when` property. However, `when` is not an item type, and so it cannot be used as an option under `select_one`.
+You can conditionally show or hide configuration fields based on the user's environment, license entitlements, and configuration choices. For example, conditional statements in the `when` property can be used to evaluate:
+* The Kubernetes distribution of the cluster
+* If the license includes a specific feature entitlement
+* The number of users that the license permits
+* If the user chooses to bring their own external database, rather than using an embedded database offered with the application
 
-When the `when` property for a group or item evaluates to `false`, the group or item is not displayed on the Config page.
-
-If the user provides input on the Config page that makes a `when` property evaluate to `true`, then the Config page updates immediately to display the previously hidden group or item.
+You can construct conditional statements in the `when` property using KOTS template functions. KOTS template functions are a set of custom template functions based on the Go text/template library. For more information, see [About Template Functions](/reference/template-functions-about).
--- a/docs/partials/config/_when-requirements.mdx
+++ b/docs/partials/config/_when-requirements.mdx
@@ -1,0 +1,8 @@
+`when` has the following requirements and limitations:
+* The `when` property accepts the following types of values:
+     * Booleans
+     * Strings that match "true", "True", "false", or "False"
+    
+    [KOTS template functions](/reference/template-functions-about) can be used to render these supported value types.
+* For the `when` property to evaluate to true, the values compared in the conditional statement must match exactly without quotes  
+* `when` cannot be applied to the items nested under a `select_one` item. To conditionally show or hide `select_one` items, apply the `when` property to the `select_one` item itself.
--- a/docs/partials/config/_when-requirements.mdx
+++ b/docs/partials/config/_when-requirements.mdx
@@ -1,8 +1,6 @@
-`when` has the following requirements and limitations:
 * The `when` property accepts the following types of values:
      * Booleans
      * Strings that match "true", "True", "false", or "False"
     
     [KOTS template functions](/reference/template-functions-about) can be used to render these supported value types.
-* For the `when` property to evaluate to true, the values compared in the conditional statement must match exactly without quotes  
-* `when` cannot be applied to the items nested under a `select_one` item. To conditionally show or hide `select_one` items, apply the `when` property to the `select_one` item itself.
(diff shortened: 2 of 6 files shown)


docs/reference/custom-resource-config.mdx
```diff
@@ -131,6 +131,23 @@ spec:
...
+For examples of configuring the `when` property, see [Using Conditional Statements in Configuration Fields](/vendor/config-screen-conditional).
+
+#### Requirements and Limitations
+
+`when` has the following requirements and limitations:
+* The `when` property accepts the following types of values:
+   * Booleans
+   * Strings that match "true", "True", "false", or "False"
+
+   [KOTS template functions](/reference/template-functions-about) can be used to render these supported value types.
```
paigecalvert, line 143: ^ it looks like both "true" as a string and true as a boolean worked when I tested it. Integers didn't work

This linter rule in the vendor portal it seems to confirm: https://github.com/replicatedhq/vandoor/blob/b6116908862b263f4260e96e3186ee0dd117ae30/graphql-api/src/policies/rego/kots-spec-default.rego#L669

docs/reference/custom-resource-config.mdx
```diff
@@ -131,6 +131,23 @@ spec:
...
+* Values compared in the conditional statement of the `when` property must match exactly without quotes for the `when` property to evaluate to true
+* `when` can be used only as a property of a group or item. `when` is _not_ an item type and so cannot be used as an option under `select_one`
```
paigecalvert, line 145: > so cannot be used as an option under `select_one`

this was here before so I imagine this is a common issue that people have run into

docs/reference/custom-resource-config.mdx
```diff
@@ -131,6 +131,24 @@ spec:
...
+
+`when` has the following requirements and limitations:
+* The `when` property accepts the following types of values:
+   * Booleans
+   * Strings that match "true", "True", "false", or "False"
+
+   [KOTS template functions](/reference/template-functions-about) can be used to render these supported value types.
+* Values compared in the conditional statement of the `when` property must match exactly without quotes for the `when` property to evaluate to true
+* `when` can be used only as a property of a group or item
+* `when` is _not_ an item type and cannot be used as a standalone item, including as an option under `select_one` items
```
paigecalvert, line 146: >including as an option under `select_one` items

this was called out before, so I'm assuming it's a common point of confusion. I updated the description around it to more generically explain that `when` can't be used as an item

docs/reference/custom-resource-config.mdx
```diff
@@ -131,6 +131,24 @@ spec:
...
+
+#### Example
+
+The following example shows a group of items that is displayed on the **Config** page only when the user enables the `http_enabled` configuration field. This example uses the KOTS [ConfigOptionEquals](/reference/template-functions-config-context#configoptionequals) template function to evaluate the value of the `http_enabled` configuration field.
+
```
paigecalvert, line 151: ^ added a description

docs/reference/custom-resource-config.mdx
```diff
@@ -145,7 +163,7 @@ spec:
     - name: example_group_2
       title: Second Group
       description: This is descriptive help text.
-      # The group is not displayed if when is false
```
paigecalvert, line 148: ^ strong double negative here

docs/reference/custom-resource-config.mdx
```diff
@@ -166,12 +184,14 @@ spec:
...
         type: password
-        value: "{{repl RandomString 10}}"     
+        value: '{{repl RandomString 10}}'     
 ```
 
 ![Only the first and third groups appear on the config screen](/images/config-screen-group-when-false.png)
 [View a larger version of this image](/images/config-screen-group-when-false.png)
 
+For additional examples of configuring the `when` property, see [Using Conditional Statements in Configuration Fields](/vendor/config-screen-conditional).
+
```
paigecalvert, line 194: ^ added a link to the large set of conditional examples

docs/vendor/config-screen-conditional.mdx
```diff
@@ -1,21 +1,15 @@
...
 ## Overview
 
-The `when` property in the Config custom resource denotes configuration items that are displayed on the admin console **Config** page only when a condition evaluates to true. When the condition evaluates to false, the field is not displayed. It can be useful to show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application.
-
-You can conditionally show or hide configuration fields based on the user's environment, license entitlements, and preferences. For example, conditional statements in the `when` property can be used to evaluate:
-* The Kubernetes distribution of the cluster
-* If the license includes a specific feature entitlement
-* The number of users that the license permits
-* If the user chooses to bring their own external database, rather than using an embedded database offered with the application
+<PropertyWhen/>
```
paigecalvert, line 9: ^ no changes except replacing the overview with the partial

docs/reference/custom-resource-config.mdx
cbodonnell: this aligns with what i've seen too

docs/reference/custom-resource-config.mdx
```diff
@@ -131,6 +131,24 @@ spec:
...
+#### Requirements and Limitations
+
+`when` has the following requirements and limitations:
+* The `when` property accepts the following types of values:
+   * Booleans
+   * Strings that match "true", "True", "false", or "False"
+
+   [KOTS template functions](/reference/template-functions-about) can be used to render these supported value types.
+* Values compared in the conditional statement of the `when` property must match exactly without quotes for the `when` property to evaluate to true
```
cbodonnell, line 144: i see this was moved from the partial [here](https://github.com/replicatedhq/replicated-docs/pull/2257/files#diff-639ebb6d1ebbbd99eead40d225e5bda26e5ea17ac273181b68a9e183db84096cL1), but i'm not sure i'm understanding what this call out is supposed to be communicating exactly

docs/reference/custom-resource-config.mdx
cbodonnell: found this section of the code for how we parse the values https://github.com/replicatedhq/kotskinds/blob/main/multitype/quotedbool.go#L21-L22. seems like there's a variety of things that would work. i probably need to test some of these first to confirm, but i wonder if a quoted integer like `"0"` works based on this?

paigecalvert: A interesting, I can test that too

docs/reference/custom-resource-config.mdx
paigecalvert: Yeah this was here before, and to me it sounded like it was trying to say that "even slight variations will render false so be sure that they match exactly". Not sure if this is maybe obvious enough that it could be removed?

ajp-io (commented):
Just from a quick read, I noticed that `when` is a property of a group and of an item, but it seems like there's some conflating of the two. Like the first mention of `when` on the page is about groups, but then there are mentions of items under there (e.g., "`when` can be used only as a property of a group or item").

It looks like we need a clearer separation of the two, though obviously the concepts are the same and the templating and whatnot.

docs/reference/custom-resource-config.mdx
paigecalvert: Looks like the linter is throwing an error for "0" or "1"

Screenshot 2024-04-15 at 9 11 37 AM

docs/reference/custom-resource-config.mdx
```diff
@@ -436,12 +456,12 @@ Items have a `name`, `title`, `type`, and other optional properties.
...
 </table>
 
-### `when`
+### `when` {#when-item}
 
 <table>
   <tr>
     <th>Description</th>
-    <td><p><PropertyWhen/></p></td>
+    <td><p>The <code>when</code> property denotes configuration groups or items that are displayed on the admin console <strong>Config</strong> page only when a condition evaluates to true. <code>when</code> is a property of both groups and items. For more information about the <code>when</code> property, including requirements and limitations, see <a href="/reference/custom-resource-config#when">Group Properties > `when`</a> above.</p></td>
```
paigecalvert, line 464: ^ changed this description to explain that it applies to groups and items, and points the user to the Group > when section for more info

docs/partials/config/_property-when.mdx
```diff
@@ -1,7 +1,7 @@
-The `when` property denotes conditional user inputs that are only displayed on the admin console Config page when a condition evaluates to true. The equality check in the conditional statement of the `when` property must match exactly, without quotes.
+It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application. Conditional statements in the `when` property can be used to evaluate the user's environment, license entitlements, and configuration choices. For example:
```
ajp-io, line 1: I don't care about the particular wording I'm suggesting here. But something that softens this sentence might be good since technically you can evaluate other stuff, and this list isn't exhaustive. This might not be realistic, but you could evaluate a random string so that some subset of your customers see config options for like an A/B test.
```suggestion
It can be useful to conditionally show or hide fields so that your users are only provided the configuration options that are relevant to them. This helps to reduce user error when configuring the application. Conditional statements in the `when` property can be used to evaluate things like the user's environment, license entitlements, and configuration choices. For example:
```

docs/partials/config/_when-requirements.mdx
```diff
@@ -0,0 +1,8 @@
+`when` has the following requirements and limitations:
+* The `when` property accepts the following types of values:
+     * Booleans
+     * Strings that match "true", "True", "false", or "False"
+    
+    [KOTS template functions](/reference/template-functions-about) can be used to render these supported value types.
+* For the `when` property to evaluate to true, the values compared in the conditional statement must match exactly without quotes  
+* `when` cannot be applied to the items nested under a `select_one` item. To conditionally show or hide `select_one` items, apply the `when` property to the `select_one` item itself.
```
ajp-io, line 8: Presumably this is only relevant to the `when` on an item, so maybe remove from the partial and just put it in the item property section.

docs/reference/custom-resource-config.mdx
```diff
@@ -145,7 +161,7 @@ spec:
     - name: example_group_2
       title: Second Group
       description: This is descriptive help text.
```
ajp-io, line 163: Nbd to fix here since it might involve a new screenshot, but it would help if these descriptions weren't here or were different for each group. I actually thought this sample was wrong because I went looking for that description and saw it in the screenshot, but it was from the third group and not the second group.

docs/reference/custom-resource-config.mdx
```diff
@@ -129,8 +131,22 @@ spec:
...
+`when` is a property of both groups and items. See [Item Properties > `when`](/reference/custom-resource-config#when-item) below.
+:::
+
+#### Requirements and Limitations
+
+<WhenRequirements/>
+
+#### Example
+
+The following example shows a group of items that is displayed on the **Config** page only when the user enables the `http_enabled` configuration field. This example uses the KOTS [ConfigOptionEquals](/reference/template-functions-config-context#configoptionequals) template function to evaluate the value of the `http_enabled` configuration field.
```
ajp-io, line 148: I like how the items `when` example says specifically what will show as a result of what condition. NBD if you use my wording.
```suggestion
In the following example, the `example_group_2` group of items will be displayed on the **Config** page only when the user enables the `http_enabled` configuration field. This example uses the KOTS [ConfigOptionEquals](/reference/template-functions-config-context#configoptionequals) template function to evaluate the value of the `http_enabled` configuration field.
```

docs/reference/custom-resource-config.mdx
```diff
@@ -166,12 +182,14 @@ spec:
       - name: password_text
         title: Password
         type: password
-        value: "{{repl RandomString 10}}"     
+        value: '{{repl RandomString 10}}'
```
ajp-io, line 185: NBD but this line is on this page a few times, but the others still have double quotes

docs/reference/custom-resource-config.mdx
```diff
@@ -145,7 +161,7 @@ spec:
     - name: example_group_2
       title: Second Group
       description: This is descriptive help text.
-      # The group is not displayed if when is false
+      # The group is displayed only when the `http_enabled` field is selected
```
ajp-io, line 164:
```suggestion
      # This group is displayed only when the `http_enabled` field is selected 
```

docs/partials/config/_property-when.mdx
paigecalvert: Works for me 👍

docs/reference/custom-resource-config.mdx
paigecalvert: Made a story for this https://app.shortcut.com/replicated/story/102871/update-descriptions-in-the-config-custom-resource-examples

PR #2320, opened 2024-05-02
Add air gap telemetry beta docs

paigecalvert:
https://deploy-preview-2320--replicated-docs.netlify.app/vendor/telemetry-air-gap


docs/vendor/telemetry-air-gap.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+Instead, to gather telemetry from air gap instances, the vendor portal analyzes archived event files contained in customer-generated support bundles. When support bundles from air gap environments are uploaded to the vendor portal, these event files are processed and associated with the correct customer and instance ID. The vendor portal then updates the insights and event data for the instance accordingly.
+
+The event files are named `sdk-events` and are created by the Replicated SDK, which continuously logs events in a Kubernetes Secret. The SDK uses the same state change logic used by the vendor portal to log events. For more information, see [How the Vendor Portal Generates Events and Insights](/vendor/instance-insights-event-data#about-events) in _About Instance and Event Data_.
```
paigecalvert, line 15: ^ Looks like the SDK is required to collect air gap telemetry. Should we add that to the requirements?

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+In addition to the `sdk-events` files, the vendor portal also receives data such as Kubernetes statuses and application Pod logs from support bundles generated in air gap environments.
+
+All support bundles that are uploaded to the vendor portal from any air gap customer contribute to a comprehensive dataset, providing parity in the telemetry for air gap and online instances. Replicated recommends that you collect support bundles from air gap customers regularly (monthly or quarterly) to improve the completeness of air gap telemetry. The vendor portal handles any overlapping event archives idempotently, ensuring data integrity.
+
+## Requirements
+
+Air gap telemetry has the following requirements:
+
+* KOTS v1.101.0 or later must be installed in the environment where the instance is running
```
paigecalvert, line 25: ^ We say KOTS is required, but KOTS is not mentioned in the Overview, so I'm not sure it's clear why it's required. What does KOTS contribute to this process?

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+Instead, to gather telemetry from air gap instances, the vendor portal analyzes archived event files contained in customer-generated support bundles. When support bundles from air gap environments are uploaded to the vendor portal, these event files are processed and associated with the correct customer and instance ID. The vendor portal then updates the insights and event data for the instance accordingly.
+
+The event files are named `sdk-events` and are created by the Replicated SDK, which continuously logs events in a Kubernetes Secret. The SDK uses the same state change logic used by the vendor portal to log events. For more information, see [How the Vendor Portal Generates Events and Insights](/vendor/instance-insights-event-data#about-events) in _About Instance and Event Data_.
+
+In addition to the `sdk-events` files, the vendor portal also receives data such as Kubernetes statuses and application Pod logs from support bundles generated in air gap environments.
```
paigecalvert, line 17: ^ How does the vendor portal use the pod logs/status data? Also, I image this is reliant upon the types of collectors that the vendor included in the support bundle spec?

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+All support bundles that are uploaded to the vendor portal from any air gap customer contribute to a comprehensive dataset, providing parity in the telemetry for air gap and online instances. Replicated recommends that you collect support bundles from air gap customers regularly (monthly or quarterly) to improve the completeness of air gap telemetry. The vendor portal handles any overlapping event archives idempotently, ensuring data integrity.
+
+## Requirements
+
+Air gap telemetry has the following requirements:
+
+* KOTS v1.101.0 or later must be installed in the environment where the instance is running
+* The support bundle must be generated by KOTS. Generic troubleshoot specs (which do not include instance data) do not contribute to telemetry for air gap instances.
```
paigecalvert, line 26: ^ Is this requirement accurate? I think I had pulled this from that google doc, but now that I'm reading through it, we say that the support bundle can be generated from the CLI or through the admin console, so I'm confused by what it means that it has to be generated by KOTS. I'm also not sure what the part about "generic support bundles not including instance data" means, as I don't think there's any such thing as a "KOTS support bundle"

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -0,0 +1,46 @@
...
+* KOTS must be running in air gapped mode to collect telemetry in the support bundle.
```
paigecalvert, line 27: ^ similarly, it looks like we're implying that KOTS collect instance data and add it to the support bundle, but above in the Overview, it sounds like it's the SDK that's doing the collecting

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Limitations
+
+Air gap telemetry has the following limitations:
+
+* Air gap telemetry is supported only for instances installed with KOTS. Air gap telemetry is not supported for instances installed with the Helm CLI.
+* Telemetry data is capped at 4,000 events or 1MB; whichever limit is reached first. For optimal use, consider collecting support bundles regularly (monthly or quarterly) from air gap customers.
```
paigecalvert, line 34: ^ what happens when the limit is reached?

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+The event files are named `sdk-events` and are created by the Replicated SDK, which continuously logs events in a Kubernetes Secret. The SDK uses the same state change logic used by the vendor portal to log events. For more information, see [How the Vendor Portal Generates Events and Insights](/vendor/instance-insights-event-data#about-events) in _About Instance and Event Data_.
+
+In addition to the `sdk-events` files, the vendor portal also receives data such as Kubernetes statuses and application Pod logs from support bundles generated in air gap environments.
+
+All support bundles that are uploaded to the vendor portal from any air gap customer contribute to a comprehensive dataset, providing parity in the telemetry for air gap and online instances. Replicated recommends that you collect support bundles from air gap customers regularly (monthly or quarterly) to improve the completeness of air gap telemetry. The vendor portal handles any overlapping event archives idempotently, ensuring data integrity.
```
paigecalvert, line 19: Is there anything to add about how custom metrics are collected for air gap instances? I guess the SDK is required for that, and are custom metrics also included in the sdk-events file?

docs/vendor/telemetry-air-gap.mdx
alicenstar: I think this refers to the fact that we can display a granular app status on the instance details page for when app status changes.

Example (except I don't have any resource states populated here unfortunately):
Screenshot 2024-05-07 at 1 44 06 PM

divolgin: That data is only used by analyzers. Vendor portal doesn't use it.  App status comes from telemetry data (instance events).

paigecalvert: Gotcha, so it has to do with what's collected by support bundle collectors and not really specific to telemetry. Sounds like this mention of logs/statuses could probably just be removed from this topic then to avoid confusion (?)

docs/vendor/telemetry-air-gap.mdx
divolgin: I'm pretty sure `sdk-events` is not a thing. All files are named `report.json` and they are stored in different folders. But I would also question whether this should be documented because this also makes it harder for us to change where this data is stored.

paigecalvert: True, probably unnecessarily detailed regardless of what it's called

docs/vendor/telemetry-air-gap.mdx
divolgin: Until recently KOTS was the only way to install in air gap environments. Maybe @ajp-io can shed some light here.

docs/vendor/telemetry-air-gap.mdx
divolgin: Admin Console provides UI and CLI options. KOTS is required to generate the spec that will live in the cluster that support bundle can use (even when running from command line). But the command shown [here](CustomAppMetricsReportFiles), for example, will not contain any telemetry data because it uses the generic spec.

docs/vendor/telemetry-air-gap.mdx
divolgin: Older events are purged until the payload is within the limit.

docs/vendor/telemetry-air-gap.mdx
divolgin: I agree.

docs/vendor/telemetry-air-gap.mdx
ajp-io: We don't have any explicit support for air gap Helm installations (like no air gap bundles), but there's also nothing preventing a vendor from doing an air gap Helm install on their own.

If you look at the docs [here](https://docs.replicated.com/vendor/install-with-helm#limitations), we just say that we don't have air gap bundles—we don't say air gap isn't supported altogether (because that was confusing and not entirely true).

So I don't think we need to limit air gap telemetry to KOTS installations, because we do already have some vendors like Howso installing with Helm in air gapped envs.

paigecalvert: I was also wondering if maybe there was something specific with KOTS 1.101.0 (which was called out here https://docs.google.com/document/d/1ixUbL5a7Cx21UZEcJreAD9H8BN4MpjA41K8SEjcklSg/edit#heading=h.nizsivsuh7hi), but it doesn't look like it: https://docs.replicated.com/release-notes/rn-app-manager#11010

paigecalvert: > We don't have any explicit support for air gap Helm installations (like no air gap bundles), but there's also nothing preventing a vendor from doing an air gap Helm install on their own.
> 
> If you look at the docs [here](https://docs.replicated.com/vendor/install-with-helm#limitations), we just say that we don't have air gap bundles—we don't say air gap isn't supported altogether (because that was confusing and not entirely true).
> 
> So I don't think we need to limit air gap telemetry to KOTS installations, because we do already have some vendors like Howso installing with Helm in air gapped envs.

Neat, that makes sense. So, most likely vendors using air gap telemetry will distribute their apps with KOTS. But KOTS isn't technically required to get air gap telemetry. Whatcha think @brianrhea ?

docs/vendor/telemetry-air-gap.mdx
paigecalvert: Hm, I think what's confusing me about the CLI part is that I thought you use the support-bundle CLI to generate bundles, not the kots CLI. And the support-bundle CLI can be used for KOTS or Helm installs

I am missing something here... I feel like maybe I am not understanding the relationship between KOTS and the support-bundle CLI

ajp-io: Not sure that I'm answering the question, but you're right that you use the support-bundle plugin, not the kots CLI, to collect a support bundle. I think what Dmitriy is saying is that you can collect the support bundle by pushing a button in the admin console or by running a support-bundle command. But the spec that defines the support bundle must be the spec that KOTS created and stored in the cluster.

That's probably lower level than a vendor needs to know. Something like "you have to generate the support bundle by clicking the button in the admin console or by running the `support-bundle` command running in the admin console" probably suffices.

PR #2331, opened 2024-05-08
SDK Air Gap Mode

paigecalvert:


docs/vendor/replicated-sdk-airgap.mdx
```diff
@@ -0,0 +1,50 @@
+# Installing the SDK in Air Gap Environments (Beta)
+
+This topic explains how to install the Replicated SDK in air gap environments by enabling air gap mode.
+
+## Overview
+
+The Replicated SDK has an _air gap mode_ that allows it to run in environments with no outbound internet access. When installed in air gap mode, the SDK does not attempt to connect to the internet. This avoids any failures that would occur when the SDK is unable to make outbound requests in air gap environments.
+
+Air gap mode is enabled when `isAirgap: true` is set in the values for the SDK Helm chart. For more information, see [Install the SDK in Air Gap Mode](#install) below. Allowing air gap mode to be controlled with the `isAirgap` value means that vendors and enterprise customers do not need to rely on air gap environments being automatically detected, which is unreliable and error-prone. The `isAirgap` value also allows the SDK to be installed in air gap mode even if the instance can access the internet.
```
paigecalvert, line 9: >Allowing air gap mode to be controlled with the `isAirgap` value means that vendors and enterprise customers do not need to rely on air gap environments being automatically detected, which is unreliable and error-prone. The `isAirgap` value also allows the SDK to be installed in air gap mode even if the instance can access the internet.

I included this statement because I saw that a vendor had asked Alex about it here: https://github.com/replicated-collab/git-guardian-kots/issues/147#issuecomment-1813937173

If this doesn't feel like it would be an FAQ, we can remove this part about why having air gap mode be a value is a good idea.

docs/vendor/replicated-sdk-airgap.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+## Differences in Air Gap Mode
+
+Air gap mode differs from non-air gap installations of the SDK in the following ways:
+* The SDK stores instance telemetry and custom metrics in a Kubernetes Secret in the customer environment, rather than attempting to send telemetry and custom metrics back to the Replicated vendor portal. The telemetry and custom metrics stored in the Secret are collected whenever a support bundle is generated in the environment, and are reported when the support bundle is uploaded to the vendor portal. For more information about telemetry for air gap instances, see **LINK TO AIR GAP TELEMETRY TOPIC**.
+* The SDK returns an empty array (`[]`) for any requests to check for updates using the [`/api/v1/app/updates`](/reference/replicated-sdk-apis#get-appupdates) SDK API endpoint. This is because the SDK is not able to receive updates from the vendor portal when running in air gap environments.
+* The SDK does not atttempt to update tags for the instance in the vendor portal using the [`/app/instance-tags`](/reference/replicated-sdk-apis#post-appinstance-tags) SDK API endpoint.
```
paigecalvert, line 16: ^ this one about instance-tags was a guess. I figured the other possibility is that this also get stored in the cluster and then passed to the vendor portal via support bundles?

docs/vendor/replicated-sdk-airgap.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+In air gap mode, the SDK can still make requests to SDK API endpoints that do not require outbound internet access, such as the [`license`](/reference/replicated-sdk-apis#license) endpoints and the [`/app/info`](/reference/replicated-sdk-apis#get-appinfo) endpoint. Note that any values returned by these endpoints might not include the latest information available in the vendor portal because the SDK cannot receive updates when running in air gap environments.
```
paigecalvert, line 18: ^ this one about being able to make requests to endpoints like the license endpoints was also a guess.

docs/vendor/replicated-sdk-airgap.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+Air gap mode is enabled when `isAirgap: true` is set in the values for the SDK Helm chart. For more information, see [Install the SDK in Air Gap Mode](#install) below. Allowing air gap mode to be controlled with the `isAirgap` value means that vendors and enterprise customers do not need to rely on air gap environments being automatically detected, which is unreliable and error-prone. The `isAirgap` value also allows the SDK to be installed in air gap mode even if the instance can access the internet.
+
+## Differences in Air Gap Mode
+
+Air gap mode differs from non-air gap installations of the SDK in the following ways:
+* The SDK stores instance telemetry and custom metrics in a Kubernetes Secret in the customer environment, rather than attempting to send telemetry and custom metrics back to the Replicated vendor portal. The telemetry and custom metrics stored in the Secret are collected whenever a support bundle is generated in the environment, and are reported when the support bundle is uploaded to the vendor portal. For more information about telemetry for air gap instances, see **LINK TO AIR GAP TELEMETRY TOPIC**.
```
paigecalvert, line 14: >For more information about telemetry for air gap instances, see **LINK TO AIR GAP TELEMETRY TOPIC**.

Hoping to merge this around the same time that the air gap telemetry PR goes out

docs/vendor/replicated-sdk-airgap.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+### KOTS
+
+When the SDK is installed by KOTS in an air gap environment, KOTS automatically sets `isAirGap: true` in the SDK Helm chart values to enable air gap mode. No additional configuration is required.
+
+KOTS automatically detects an air gap installation and sets `isAirGap: true` in the SDK values when any of the following are true:
+* The user passes the `--airgap=true` flag with the `kots install` command.
+* The user provides the path to the `.airgap` bundle for the release by passing the `--airgap-bundle` flag with the `kots install` command. Setting `--airgap-bundle` implies `--airgap=true`.
+* The user provides a local registry where images are pushed by passing the `--kotsadm-registry`, `--registry-username`, and `--registry-password` flags with the `kots install` command.
+* For Replicated kURL installations, the user passes `-s airgap` with the kURL installation command.
```
paigecalvert, line 48: ^ These were all the ways I could think of that trigger air gap mode for KOTS, though this list could certainly use confirmation.

ajp-io: Yeah, I don't know whether this is correct or exhaustive. @sgalsaleh could confirm, but we could also consider just leaving it at "if you do an air gap installation, it'll get set automatically," and assume people know how to do air gap in other places. This list confuses me a bit because I'm like, shouldn't there just be two ways to do an air gap installation, one for KOTS and one for kURL?

docs/vendor/replicated-sdk-airgap.mdx
ajp-io: You can make these requests, but the values returned will be what was injected into the SDK when the chart was pulled. Whereas for online environments, we reach out to the vendor portal for up-to-date information for these.

docs/vendor/replicated-sdk-airgap.mdx
ajp-io: I don't expect it to be an FAQ, but it doesn't hurt to leave it either.

docs/vendor/replicated-sdk-airgap.mdx
ajp-io: You might have to ask the vendor portal team about this since they added that endpoint. I don't know how it handles air gap (or if it does, which would be unfortunate).

paigecalvert: Yeah fair enough, probably easier to just say if you're following these steps to do an air gap install, it'll work. And then we'll see if anyone has questions

docs/vendor/replicated-sdk-airgap.mdx
paigecalvert: Yeah fair enough, probably easier to just say if you're following these steps to do an air gap install, it'll work. And then we'll see if anyone has questions

docs/vendor/replicated-sdk-airgap.mdx
paigecalvert: The tags get stored in the support bundle, but the vendor portal does have a way to extract them yet. So nothing will "break" but tags won't get updated in air gap mode
 https://replicated.slack.com/archives/CEEFDQURM/p1715269121782909

ajp-io (approved):

PR #2332, opened 2024-05-09
Require version for Helm extensions

ajp-io:


paigecalvert (approved):
Thanks @ajp-io , approved. Feel free to merge if you're okay with the edits (I removed the part about it most likely deploying the latest version even if you don't include it in favor of emphasizing how upgrades will be messed up if you omit the version)

PR #2361, opened 2024-05-20
HOLD FOR RELEASE add airgap to embedded-cluster

laverya:
Adds airgap docs for embedded-cluster. Wait until it hits GA first.


paigecalvert:
Gonna close this out as the air gap docs are now live here: https://docs.replicated.com/enterprise/installing-embedded-air-gap

PR #2374, opened 2024-05-22
Clarify that KOTS 1.92.1 supports air gap telemetry

paigecalvert:
https://deploy-preview-2374--replicated-docs.netlify.app/vendor/telemetry-air-gap


docs/vendor/telemetry-air-gap.mdx
```diff
@@ -14,17 +14,29 @@ Air gap instances run in environments without outbound internet access. This lim
 
 <AirGapTelemetry/>
 
-The following diagram demonstrates how air gap telemetry is collected and stored by the SDK in a customer environment, and then shared to the vendor portal in a support bundle:
+The following diagram demonstrates how air gap telemetry is collected and stored by the Replicated SDK in a customer environment, and then shared to the vendor portal in a support bundle:
 
 <img alt="Air gap telemetry collected by the SDK in a support bundle" src="/images/airgap-telemetry.png" width="800px"/>
```
paigecalvert, line 19: Seemed like the diagram was still serving it's purpose, since the intro does clarify that the diagram is showing how the SDK collects/stores telemetry (and we want to recommend including the SDK). I could also update this to add a box for KOTS in the cluster

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -14,17 +14,29 @@ Air gap instances run in environments without outbound internet access. This lim
...
+## Requirements
 
-The Replicated SDK must be installed in _air gap mode_ in the environment where the instance is running for air gap telemetry to be collected. For more information, see [Installing the SDK in Air Gap Environments](/vendor/replicated-sdk-airgap).
+Air gap telemetry has the following requirements:
+
+* To collect telemetry from an air gap instance, one of the following must be installed in the cluster where the instance is running:
+   
+   * The Replicated SDK installed in _air gap mode_. See [Installing the SDK in Air Gap Environments](/vendor/replicated-sdk-airgap).
+   
+   * KOTS v1.92.1
```
AmberAlston, line 33: should add an "or later" here

docs/vendor/telemetry-air-gap.mdx
```diff
@@ -14,17 +14,29 @@ Air gap instances run in environments without outbound internet access. This lim
...
+   
+   :::note
+   If both the SDK and KOTS v1.92.1 are installed in the cluster (such as when a Helm chart that includes the SDK is installed by KOTS), then both collect and store instance telemetry.
+   :::
```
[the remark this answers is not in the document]
AmberAlston: Yes, we should add something there. Reading as-is I'd be worried I'm going to overload that secret size limitation because this reads like we are storing data twice. 

However in this docs page /vendor/instance-insights-event-data#about-reporting, the first bullet also reads that way, so I think we still need some clarity on the actual behavior, particularly for this first bullet of concern. If both are actually sending to VP today, is it VP that is de-duping? And are both actually writing to a secret if both are present in airgap mode? If so, is VP de-duping in this scenario as well? @brianrhea This is something the team should be able to help clarify 

image

docs/vendor/telemetry-air-gap.mdx
paigecalvert: @AmberAlston I added a statement to the bottom of the requirements section about how the SDK is strongly recommended.

My goal was to avoid implying that the SDK was recommended _instead of_ KOTS. Also wanted to make sure the base requirements remained clear.

AmberAlston: lgtm thanks

PR #2379, opened 2024-05-24
Clarify usage of kotsadm namespace

ajp-io:
This PR might not be perfect style-wise, but I just wanted to do it before I forgot. We want to stop documenting `--kotsadm-namespace`, because it's confusing, and instead just tack the namespace onto the registry host, which also works.

Additionally, the `--kotsadm-namespace` flag is poorly named because it's the registry namespace. So I updated some help menus to make that clearer.

I don't know if the `[/REGISTRY_HOST]` syntax is clear. Often CLIs will put stuff in brackets if it's optional. But if you want to use a different way to show that it's optional, or not show that it's optional and people can leave it off if they want, that's fine.


paigecalvert (approved):
Nice, thanks @ajp-io. Gonna be a lot simpler to just tell people to use the one flag

I did the following:
* made some updates in the relevant kots CLI pages so that the kotsadm registry/namespace partials could be used in all the tables. (this made the Files changes a lot less easy to review...sorry about that)
* Added an (Optional) label next to the REGISTRY_NAMESPACE placeholder description in the relevant snapshots procedures that you had edited
* Added a note to the description of the `kotsadm-namespace` flag to say that replicated recommends using the `kotsasm-registry` flag instead to override both the hostname and namespace at once

ajp-io:
@paigecalvert I hate to put this on you, but would you mind resolving the conflicts here? Maybe they're easy—I didn't look. Just didn't have time to look more closely.

paigecalvert:
> @paigecalvert I hate to put this on you, but would you mind resolving the conflicts here? Maybe they're easy—I didn't look. Just didn't have time to look more closely.
> 
> We are good to merge otherwise.

@ajp-io no worries, will do

PR #2408, opened 2024-05-30
Upload support bundle from Admin Console docs

alicenstar:
https://app.shortcut.com/replicated/story/105720/support-bundle-upload-enabled-to-beta

Relevant doc: https://docs.google.com/document/d/1pJM4u1Tp37oL5YacifOr7ESM_Y5vLOwJhynS7jM_XS4/edit

Docs preview:
- https://deploy-preview-2408--replicated-docs-upgrade.netlify.app/vendor/support-enabling-direct-bundle-uploads
- https://deploy-preview-2408--replicated-docs-upgrade.netlify.app/enterprise/troubleshooting-an-app#generate-a-support-bundle-using-the-admin-console


alicenstar:
@brianrhea Started on these docs, they still need a bit of work.

AmberAlston:
I think this still needs a bit of work before we would want to make this live. 

This as-is content is written to an end-user audience vs the Vendor, so this page (/vendor/support-inspecting-support-bundles) which is all about How a vendor should inspect already provided bundles doesn’t feel like the right place for this content. There’s already a page for the end-user focus content (https://docs.replicated.com/enterprise/troubleshooting-an-app). It would be better to integrate the as-is content into this page, and highlight the capability with additional text like: _If your Vendor has enabled it for your online installation, you may also see a button to “send bundle to vendor” available. Clicking this button will send the support bundle directly to your vendor. We recommended additionally following up with your vendor to let them know the bundle has been provided._ 

We do need a place to have the content this docs story originally asked for, which is letting a vendor know how to enable and what the caveats are. I don’t see that info in the PR today. 

One idea below, but I’ll defer to @brianrhea  on the approach here. 

* Add a new sub-section under “Support Bundles” like “Enabling Direct Bundle Uploads (KOTS-Only) (Beta)” 
* Borrow this same content from /vendor/support-bundle-generating with a bit of a tweak, plus content additions (see below)

For KOTS installations, users can generate support bundles from the Admin Console UI or CLI, and provide those to you to assist in application troubleshooting. For more information, see [Generating Support Bundles](https://docs.replicated.com/enterprise/troubleshooting-an-app) in the KOTS documentation. 

Optionally, for online KOTS installations, you can enable a Beta feature to allow application end-users to send that support bundle directly to your Vendor Portal account rather than use an alternative support bundle sharing method. When this feature is enabled, the end-user will see a button to “send bundle to vendor.” When selected, the support bundle will be immediately uploaded and made available in your Vendor Portal team account for this customer under the Troubleshoot tab. 

Please note the following caveats:
* There is currently no notification provided when the end user takes this action. You should not enable this feature for a given customer license unless you already have an existing escalation process to avoid the potential miss of a directly uploaded customer support bundle. 
* This feature only supports online installs. If enabled, but installed in airgap mode, the upload button does not appear. 
* There is a 500mb limit for Support Bundles uploaded directly via the Admin Console

How to enable for your customer: 
………., default off, etc 

{Should probably put feedback/request note on the page as well)

AmberAlston (changes requested):
See prior comment

alicenstar:
@AmberAlston Updated this based on your feedback - let me know what you think! I tweaked some copy since some of it sounded a little repetitive, and hopefully I addressed all your points.

docs/vendor/support-enabling-direct-bundle-uploads.md
```diff
@@ -0,0 +1,32 @@
+# Enabling Support Bundle Uploads (KOTS-only) (Beta)
+
+:::note
+Direct bundle uploads is in beta. The functionality, requirements, and limitations of direct bundle uploads are subject to change.
+:::
+
+When this feature is enabled, customers using online KOTS installations can upload support bundles directly through the Admin Console UI or CLI, eliminating the need to share the generated bundle with you manually.
```
AmberAlston, line 7: I don't believe it's correct to say you can do it via the CLI. This Alpha feature only covered the UI interaction.

AmberAlston (approved):

AmberAlston:
@alicenstar Thanks for the changes. I did remove the reference to the CLI as commented here https://github.com/replicatedhq/replicated-docs/pull/2408#pullrequestreview-2097557215 because the Alpha feature only covered the UI. I don't know of a way to direct upload via the CLI. If I'm incorrect on that you'll want to add the CLI reference back, but I wanted to be able to send these live for you and the rest of the content looked good. cc @brianrhea

AmberAlston:
Force merging in place of Docs team review because the team member is on vacation an unavailable to apporve.

PR #2416, opened 2024-06-07
EC proxy docs

jdewinne:


ajp-io:
We should also document the limitation that the proxy values cannot be changed after installing

paigecalvert (dismissed):
LGTM. Pushed some style edits but no material changes

ajp-io:
Thanks, @paigecalvert. Made a tiny change. Feel free to reapprove whenever you have a second.

PR #2420, opened 2024-06-12
update role requirements for GAR private repos

adamancini:
the correct role is Artifact Registry Reader, not Storage Object Viewer


paigecalvert (approved):
thanks @adamancini !

PR #2428, opened 2024-06-13
Say how to download a specific version

ajp-io:
I'm fine if this deserves its own section or should be worded differently. Just wanted to be sure to capture it.


paigecalvert (approved):
Oh nice, I was going to ask you about this. Let's merge this in and then I'll capture it in the other PR for EC GA that I'm drafting

PR #2441, opened 2024-06-21
Update gitops-workflow.mdx

grantmiller:


paigecalvert (approved):
Rebranded the rest of the docs to use Auto-GitOps instead of just GitOps

PR #2460, opened 2024-07-02
Add additional limitations

ajp-io:
@paigecalvert I wrote this quickly, so totally fine if you reword, add links, etc. Just wanted to write it while I was thinking about it.


paigecalvert (dismissed):

docs/vendor/embedded-overview.mdx
```diff
@@ -56,7 +56,13 @@ Embedded Cluster has the following limitations:
...
 
-* **KOTS GitOps workflow not supported**: Embedded Cluster does not support using the KOTS GitOps workflow. We do not intend to support the GitOps workflow for Embedded Cluster in the future. If an end-user is interested in GitOps, consider the Helm install method instead.
+* **Custom domains not supported**: Embedded Cluster does not fully use custom domains even if they are configured. For more information about custom domains, see [About Custom Domains](/vendor/custom-domains).
```
paigecalvert, line 59: ^ Is there anything else we should say about "fully" using? Is the guidance to not expect custom domains to work at all with EC, or are there particular domains that can be customized and others not?

paigecalvert (dismissed):

docs/vendor/embedded-overview.mdx
ajp-io: You might get some that get customized and some that don't. So it's probably fine to just say it doesn't work. I don't know for sure or have confidence that any work fully.

paigecalvert (approved):

PR #2501, opened 2024-07-16
Registry known issue

paigecalvert:


docs/vendor/private-images-replicated.mdx
```diff
@@ -24,6 +24,10 @@ The Replicated registry has the following limitations:
 
 * When using Docker Build to build and push images to the Replicated registry, provenance attestations are not supported. To avoid a 400 error, include the `--provenance=false` flag to disable all provenance attestations. For more information, see [docker buildx build](https://docs.docker.com/engine/reference/commandline/buildx_build/#provenance) and [Provenance Attestations](https://docs.docker.com/build/attestations/slsa-provenance/) in the Docker documentation.
 
+## Known Issue
+
+There is a known issue with the Replicated registry that can cause a timeout error when attempting to push large images. This results in an error message similar to the following: `received unexpected HTTP status: 524`. If you see a 524 error, build your image with smaller image layers and then push the image again. If you continue seeing a 524 error, reduce the size of the image layers again.
```
xavpaice, line 29: can we mention the actual size limit, rather than "large"?  Approx. 2GB is the limit, it's apparently a time limit more than size.

paigecalvert: Agree that it would be preferable to be more specific ("large" doesn't help too much). There was some back and forth in Slack about what that number is...I shared your question there: https://replicated.slack.com/archives/CEEFDQURM/p1721314046677119?thread_ts=1721137115.789949&cid=CEEFDQURM

PR #2504, opened 2024-07-18
describe EC host storage locations

laverya:


laverya:
https://deploy-preview-2504--replicated-docs.netlify.app/vendor/embedded-overview#host-storage

paigecalvert (dismissed):
Looks good thanks @laverya

docs/vendor/embedded-overview.mdx
```diff
@@ -302,6 +302,16 @@ When you skip host preflight checks, the Admin Console still runs any applicatio
...
 
+### Host Storage
+
+Embedded Cluster uses disk space in the following locations:
+
+* `/var/lib/embedded-cluster` stores Embedded Cluster binaries, logs, and temporary files.
+* `/var/lib/k0s` stores the Kubernetes configuration, data, and container images.
+* `/var/openebs` stores PVC data.
+
+Replicated does not recommend using symlinked directories for these locations.
```
chris-sanders, line 313:
```suggestion
Embedded Cluster stores data in three primary locations, all of which must be writable:

### 1. `/var/lib/embedded-cluster`

- Used to store your installer
- Includes:
  * Logs
  * Temporary files
  * Supporting binaries (e.g., kubectl and troubleshoot)
- Should be large enough to fit at least 2 installer binaries

### 2. `/var/lib/k0s`

- Where the Kubernetes cluster runs
- Includes:
  * All Kubernetes components and binaries
  * Pod log files
  * Images currently in use by containerd
- Storage requirements:
  * The total space should be sufficient to hold the following while remaining below 85% utilization:
    1. All pod logs
    2. Two copies of infrastructure images included in k0s
    3. All pod images for your application
  * Kubernetes will initiate garbage collection of images when utilization reaches 85%
- Calculation example:
  If the sum of required storage for logs, images, and components is X GB, the total allocated space should be at least X / 0.85 GB to ensure it remains below 85% utilization.

### 3. `/var/openebs`

- Where PVC data is stored
- Needs enough storage to support all volumes used in your application
- For airgap installs:
  * Registry uses a volume
  * Requires space for two copies of your images (for registry)
  * Additional space needed for other application volume requirements
  * If the `--enable-ha` flag is used all control planes have the same requirements, data will be replicated to the other control planes.

### 3. `/tmp`
- Temporary space during installation

**Note:** Replicated does not support using symlinked directories for these locations.
```

chris-sanders: I don't know all of the storage requirments, but what bout being a little more verbose about what's located at each location. Rasa for example wanted to know where the registry stored it's data. Also it's good to at least clarify these have to be writable.

ajp-io (approved):

paigecalvert (approved):
Looks good.

I just got rid of the numbered lists because they are notoriously easy to mess up (for example, I noticed that there already had two of these headings that were erroneously labeled as "3.", and we said "three primary locations" but then listed four locations). Anyway, just for context on why ordered lists are typically reserved for step-by-step procedures

PR #2506, opened 2024-07-18
Explain when to use builders key

paigecalvert:


docs/vendor/helm-packaging-airgap-bundles.mdx
```diff
@@ -11,13 +11,13 @@ This topic describes how to package and build air gap bundles for releases that
...
 
 ## Configure the `builder` Key
 
-To build an `.airgap` bundle for a release that contains one or more Helm charts, you must configure the [`builder`](/reference/custom-resource-helmchart-v2#builder) key in the KOTS HelmChart custom resource for each Helm chart in the release.
+The purpose of the KOTS HelmChart custom resource [`builder`](/reference/custom-resource-helmchart-v2#builder) key is to set Helm values for rendering chart templates so that the output includes any required images for the `.airgap` bundle.
 
-<BuilderAirgapIntro/>
+You should configure the `builders` key if you need to change any default values in your Helm chart so that the `.airgap` bundle for the release includes all images needed to successfully deploy the chart. For example, if there is a `required` field in your Helm chart that must be set for the chart to render or if you need to change which image is used for air gap installations.
```
diamonwiggins, line 20: > For example, if there is a `required` field in your Helm chart that must be set for the chart to render or if you need to change which image is used for air gap installations.

This isn't really an example of the first sentence. It's more of an additional use case imo.

diamonwiggins (approved):
lgtm

PR #2510, opened 2024-07-19
add warning about storing secrets in a text field

St0rmz1:
add warning about storing secrets in a text field


paigecalvert (approved):
Thanks @St0rmz1 ! Just added some guidance to use the password item type instead, plus a xref. Feel free to merge if you agree with that addition

PR #2525, opened 2024-07-25
Specifically call out EKS Extended Support Versions

AmberAlston:
May need to add more specific pricing detail here in the future, but that was getting a bit content heavy. To start let's just be clear which versions are in this "Extended Support EKS" category


paigecalvert (dismissed):

paigecalvert:
Ah actually, looks like an issue with the HTML. I will fix that and re-approve

AmberAlston:
> Ah actually, looks like an issue with the HTML. I will fix that and re-approve

doh! thank you!

paigecalvert (approved):

PR #2556, opened 2024-08-09
Explain default regex for redactors

paigecalvert:
https://deploy-preview-2556--replicated-docs-upgrade.netlify.app/reference/custom-resource-redactor


docs/reference/custom-resource-redactor.md
```diff
@@ -1,14 +1,16 @@
...
 
-The default redactors can be disabled using the command line only. Replicated recommends leaving the redactors enabled.
+## Overview
 
-For Replicated KOTS, you can add custom redactors to support bundles using the Redactor custom resource manifest file. For example, you can redact API keys or account numbers, depending on your customer needs. For more information about redactors, see [Redacting Data](https://troubleshoot.sh/docs/redact/) in the Troubleshoot documentation.
+Preflight checks and support bundles include built-in redactors. These built-in redactors use regular expressions to identify and hide potentially sensitive data before it is analyzed. For example, the built-in redactors hide values like data sources, passwords, and user IDs that are commonly found in standard database connection strings. They also hide environment variables with names that begin with words like token, password, or user. To view the complete list of regex patterns for the built-in redactors, see [`redact.go`](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/redact/redact.go#L204) in the open-source Troubleshoot GitHub repo.
```
paigecalvert, line 7: New description:

>These built-in redactors use regular expressions to identify and hide potentially sensitive data before it is analyzed. For example, the built-in redactors hide values like data sources, passwords, and user IDs that are commonly found in standard database connection strings. They also hide environment variables with names that begin with words like token, password, or user. To view the complete list of regex patterns for the built-in redactors, see [`redact.go`](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/redact/redact.go#L204) in the open-source Troubleshoot GitHub repo.

docs/reference/custom-resource-redactor.md
```diff
@@ -1,14 +1,16 @@
 # Redactor (KOTS-Only)
 
-Preflight checks and support bundles include built-in redactors that hide sensitive customer data before it is analyzed. These default redactors hide passwords, tokens, AWS secrets, database connection strings, and URLs that contain usernames and passwords.
+This topic describes the Redactor custom resource. The information in this topic applies to Replicated KOTS installations only.
 
-The default redactors can be disabled using the command line only. Replicated recommends leaving the redactors enabled.
```
paigecalvert, line 5: >The default redactors can be disabled using the command line only. Replicated recommends leaving the redactors enabled.

I think this is referring to the --redact flag that you can run with support-bundle / preflight, but that is not super clear from this statement. Since --redact=false is not recommended, there's info about it elsewhere like when you run --help, and this statement doesn't even explain how you are supposed to be able to turn off redaction from the command line, I decided to just remove it

docs/reference/custom-resource-redactor.md
```diff
@@ -1,14 +1,16 @@
...
+## Overview
 
-For Replicated KOTS, you can add custom redactors to support bundles using the Redactor custom resource manifest file. For example, you can redact API keys or account numbers, depending on your customer needs. For more information about redactors, see [Redacting Data](https://troubleshoot.sh/docs/redact/) in the Troubleshoot documentation.
+Preflight checks and support bundles include built-in redactors. These built-in redactors use regular expressions to identify and hide potentially sensitive data before it is analyzed. For example, the built-in redactors hide values that match common patterns for data sources, passwords, and user IDs that can be found in standard database connection strings. They also hide environment variables with names that begin with words like token, password, or user. To view the complete list of regex patterns for the built-in redactors, see [`redact.go`](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/redact/redact.go#L204) in the open-source Troubleshoot GitHub repo.
+
```
paigecalvert, line 8: ^ new description with a link to the regex

Rewrote this so that it explains better how it's pattern matching, not magically redacting every piece of sensitive data in every situation

chris-sanders (approved):
This looks great. It's clear that we match on a lot of common patterns, and let's them review those patterns themself.

PR #2568, opened 2024-08-13
WAIT FOR RELEASE: Updates for new service node port range

ajp-io:
We are adjusting the service node port range so that people don't have to adjust it themselves as much anymore. So I added a new unsupported overrides examples using WireGuard.

I also capitalized instances of Embedded Cluster.


paigecalvert (approved):

ajp-io:
I just resolved conflicts but didn't actually make changes, so I'm going to merge without you approving again.

PR #2586, opened 2024-08-22
WAIT FOR RELEASE: Add missing requirements

ajp-io:
Technically we haven't updated the preflights to check for available space yet, so let's wait on merging this.

Also if you have thoughts on whether every little thing we preflight for should be in this list, versus having more details info separately, that would be fine too. "Linux" is a very different requirement than "no `nameserver localhost` in `resolv.conf`.


paigecalvert:
@ajp-io  Sounds good. And yeah I think for things that are less likely to cause issues in the preflights, we can at a certain point decide to say: "EC also checks for some additional things which you can see in this preflight spec over here" so the list doesn't get hard to read/update

ajp-io:
That's a good idea. The SELinux one can stay, but the nameserver one can go then.

paigecalvert (approved):

paigecalvert (dismissed):

paigecalvert:
@ajp-io checking in on this one. Still waiting to merge?

ajp-io:
I got rid of the no longer or not yet relevant stuff. Feel free to merge if it's good.

PR #2588, opened 2024-08-22
Clarify proxy settings don't apply to extensions

ajp-io:
Totally fine to tweak the wording. The point I'm trying to make is that containerd has the proxy settings, so there won't be issues pulling the images for your Helm extensions. There's only a problem if your Helm extensions themselves reach out to the internet. Then you would have to manually configure those Helm charts to use the proxy, which is likely done via the Helm vales.


paigecalvert (approved):
lmk what you think of the edit

ajp-io:
The person who brought this up was confused about the containers for the extensions getting pulled without proxying. Which isn't the case. I think this wording (and maybe even some of my original wording) doesn't clarify that enough. Like I hoped "propagated to the containers" and "that make requests to the Internet" would clarify that, but I worry they won't.

paigecalvert:
@ajp-io "getting pulled without proxying": so, they did not know that images for Helm extensions would need to be proxied in the case of a proxy install? as in, they did understand that those images would not get proxied, but they didn't realize this was a problem?

Workshopping:

>For any of your Helm extensions make requests to the internet, the given Helm charts need to be manually configured so that the containers are pulled through the proxy. Typically, this involves updating the Helm values to [point to the location of the image on the proxy?].
>
>This is required because the proxy settings that are passed during a proxy install only apply to ___. So, any Helm extensions with Helm charts that attempt to pull containers from the internet are not automatically updated to use the supplied proxy.

ajp-io:
Nope, a little backwards. They read that proxy stuff wasn't propagated to the Helm extensions, and they thought, well how do the containers get pulled then? But in reality, the containers will get pulled fine because containerd has the proxy info. What won't work is any network requests the Helm extensions make themselves (not pulling the images to run the extensions), unless they pass the proxy info along.

ajp-io:
Bump on my above comment in case you missed it. I can take a swing at wording too if you'd like.

paigecalvert:
@ajp-io here's an updated draft (left some blank spaces where I think clarification could be helpful for the user but I don't know exactly what to put)

```
If any of the charts included in your Helm extensions make requests to the internet, [such as in ___ situation or when you want to___], the given charts need to be manually configured so that those requests are made to the user-supplied proxy server instead. Typically, this requires updating the Helm values [to set the hostname/port...then templating the URL used in the requests?]. 

Note that this limitation applies only to network requests made by charts in your Helm extensions. No additional configuration is required to pull the containers required to run Helm extensions through the user-supplied proxy server during proxy installations.
```

Also wasn't sure which language is more correct: `requests made by Helm extensions` or `requests made by charts included in your Helm extensions`. They might be the same, but wasn't sure if we were talking about requests directly in the `extensions` sections of the Config yaml, or made from the charts included in the extensions

ajp-io:
I don't even have a good example for the first blank. Most charts people would use in extensions don't need to access the Internet, which is why this is an acceptable limitation for us. So while I think a clarification there could help, maybe we just skip it for now.

For the language, either could work. I might prefer "requests made by Helm extensions" since it's shorter.

The rest seemed pretty good. Let me know what you think of this:
>If any of your Helm extensions make requests to the internet, the given charts need to be manually configured so that those requests are made to the user-supplied proxy server instead. Typically, this requires updating the Helm values to set HTTP proxy, HTTPS proxy, and no proxy.
>
>Note that this limitation applies only to network requests made by your Helm extensions. The proxy settings supplied to the `install` command are used to pull the containers required to run your Helm extensions.

ajp-io:
Since we just updated this page a bunch, let me force push this change and you can approve merge if it looks good

ajp-io:
Actually, this was a bullet before, but now it's two paragraphs. Let me know what that should look like @paigecalvert

paigecalvert (approved):

PR #2592, opened 2024-08-26
Clarify --no-proxy

paigecalvert:
https://deploy-preview-2592--replicated-docs.netlify.app/vendor/embedded-overview#install-behind-a-proxy-server

### DIFF
--- a/docs/vendor/embedded-overview.mdx
+++ b/docs/vendor/embedded-overview.mdx
@@ -201,7 +201,7 @@ The following flags can be used with the Embedded Cluster `install` command to i
      <td>`--no-proxy`</td>
      <td>
        <p>Comma-separated list of hosts for which not to use a proxy.</p>
-       <p>For single-node installations, pass the IP address of the node where you are installing. For multi-node installations, when deploying the first node, pass the list of IP addresses for all nodes in the cluster (typically in CIDR notation). The proxy settings are automatically configured when joining additional nodes.</p>
+       <p>For single-node installations, pass the IP address of the node where you are installing. For multi-node installations, when deploying the first node, pass the list of IP addresses for all nodes in the cluster (typically in CIDR notation).</p>
        <p>The following are never proxied:</p>
        <ul>
          <li>Internal cluster communication (`localhost`, `127.0.0.1`, `.default`, `.local`, `.svc`, `kubernetes`)</li>
--- a/docs/vendor/embedded-overview.mdx
+++ b/docs/vendor/embedded-overview.mdx
@@ -201,7 +201,7 @@ The following flags can be used with the Embedded Cluster `install` command to i
      <td>`--no-proxy`</td>
      <td>
        <p>Comma-separated list of hosts for which not to use a proxy.</p>
-       <p>For single-node installations, pass the IP address of the node where you are installing. For multi-node installations, pass the list of IP addresses for all nodes in the cluster, typically in CIDR notation.</p>
+       <p>For single-node installations, pass the IP address of the node where you are installing. For multi-node installations, when deploying the first node, pass the list of IP addresses for all nodes in the cluster, typically in CIDR notation. The proxy settings are automatically configured when joining additional nodes.</p>
        <p>The following are never proxied:</p>
        <ul>
          <li>Internal cluster communication (`localhost`, `127.0.0.1`, `.default`, `.local`, `.svc`, `kubernetes`)</li>
--- a/docs/vendor/embedded-overview.mdx
+++ b/docs/vendor/embedded-overview.mdx
@@ -199,11 +199,19 @@ The following flags can be used with the Embedded Cluster `install` command to i
   </tr>
   <tr>
      <td>`--no-proxy`</td>
-     <td>Comma-separated list of hosts for which not to use a proxy</td>
+     <td>
+       <p>Comma-separated list of hosts for which not to use a proxy</p>
+       <p>The following are never proxied:</p>
+       <ul>
+         <li>Internal cluster communication (`localhost`, `127.0.0.1`, `.default`, `.local`, `.svc`, `kubernetes`)</li>
+         <li>Communiation to the KOTS database (`kotsadm-rqlite`)</li>
+         <li>The CIDRs used for assigning IPs to Kubernetes Pods and Services. By default, the Pod CIDR is `10.96.0.0/12` and the Service CIDR `10.244.0.0/16`. For information about how to change these defaults, see [Set IP Address Ranges for Pods and Services](#set-ip-address-ranges-for-pods-and-services).</li>
+       </ul>
+     </td>
   </tr>
 </table>
 
-**Example:**
+#### Example
 
 ```bash
 sudo ./APP_SLUG install --license LICENSE_FILE \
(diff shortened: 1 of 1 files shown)


docs/vendor/embedded-overview.mdx
```diff
@@ -199,11 +199,19 @@ The following flags can be used with the Embedded Cluster `install` command to i
...
      <td>`--no-proxy`</td>
-     <td>Comma-separated list of hosts for which not to use a proxy</td>
+     <td>
+       <p>Comma-separated list of hosts for which not to use a proxy</p>
+       <p>The following are never proxied:</p>
+       <ul>
+         <li>Internal cluster communication (`localhost`, `127.0.0.1`, `.default`, `.local`, `.svc`, `kubernetes`)</li>
+         <li>Communiation to the KOTS database (`kotsadm-rqlite`)</li>
+         <li>The CIDRs used for assigning IPs to Kubernetes Pods and Services. By default, the Pod CIDR is `10.96.0.0/12` and the Service CIDR `10.244.0.0/16`. For information about how to change these defaults, see [Set IP Address Ranges for Pods and Services](#set-ip-address-ranges-for-pods-and-services).</li>
+       </ul>
```
paigecalvert, line 209: ^ This seemed clear enough that either the default or customized CIDRs are never proxied, but LMK if you think we should explicitly state "You don't have to pass your customized CIDRs to --no-proxy"

docs/vendor/embedded-overview.mdx
```diff
@@ -199,11 +199,19 @@ The following flags can be used with the Embedded Cluster `install` command to i
   </tr>
   <tr>
      <td>`--no-proxy`</td>
-     <td>Comma-separated list of hosts for which not to use a proxy</td>
+     <td>
+       <p>Comma-separated list of hosts for which not to use a proxy</p>
+       <p>The following are never proxied:</p>
```
paigecalvert, line 204: ^ moved this list to the description of the --no-proxy flag so it's easier to find when you need it

docs/vendor/embedded-overview.mdx
```diff
@@ -218,18 +226,16 @@ Where:
 * `HOST:PORT` is the host and port of the proxy server
 * `LIST_OF_HOSTS` is the list of hosts to not proxy
 
-For multi-node clusters, the proxy settings are automatically configured when joining additional nodes. When deploying the first node of a multi-node cluster, you should pass the node IP addresses (typically in CIDR notation) to `--no-proxy`.
+#### Install Behind a Proxy in Multi-Node Clusters
```
paigecalvert, line 229: ^ added a heading so the multi-node stuff is easier to find

docs/vendor/embedded-overview.mdx
ajp-io: I thought about it when I read it, but I think it would work even if you chose to manually pass those CIDRs to `--no-proxy`. It would just get deduped or listed twice, but as long as it's there, it's good. So it's likely fine.

docs/vendor/embedded-overview.mdx
```diff
@@ -218,18 +226,16 @@ Where:
...
 
-The following are never proxied:
-* Internal cluster communication (`localhost`, `127.0.0.1`, `.default`, `.local`, `.svc`, `kubernetes`)
-* Communiation to the KOTS database (`kotsadm-rqlite`)
-* The CIDR used for assigning IPs to Kubernetes Pods and Services (`10.0.0.0/8`)
+For multi-node clusters, the proxy settings are automatically configured when joining additional nodes. When deploying the first node of a multi-node cluster, pass the node IP addresses (typically in CIDR notation) to `--no-proxy`.
```
ajp-io, line 231: Just realized in our planning today that if you're using a proxy, you should always pass the node IP address, even for single-node. It's just that for multi-node you need to pass a list of IPs or a CIDR that covers all the nodes you will later add.

docs/vendor/embedded-overview.mdx
```diff
@@ -218,18 +226,16 @@ Where:
...
 
-**Requirement:**
+#### Requirement
 
 Proxy installations require Embedded Cluster 1.5.1 or later with Kubernetes 1.29 or later. For example, Embedded Cluster 1.6.0+k8s-1.29 supports installing behind a proxy, and 1.6.0+k8s-1.28 does not. For the latest version information, see the [embedded-cluster](https://github.com/replicatedhq/embedded-cluster/releases) repository in GitHub.
```
ajp-io, line 235: Now that we have release notes, we should update these messages here and elsewhere to point to RNs instead.

docs/vendor/embedded-overview.mdx
ajp-io: So we might not need a heading for multi-node, we should just say to always pass the node IPs, where it's only the current node, or a list of IPs/a CIDR for all the nodes.

docs/vendor/embedded-overview.mdx
```diff
@@ -199,11 +199,20 @@ The following flags can be used with the Embedded Cluster `install` command to i
   </tr>
   <tr>
      <td>`--no-proxy`</td>
-     <td>Comma-separated list of hosts for which not to use a proxy</td>
+     <td>
+       <p>Comma-separated list of hosts for which not to use a proxy.</p>
+       <p>For single-node installations, pass the IP address of the node where you are installing. For multi-node installations, pass the list of IP addresses for all nodes in the cluster, typically in CIDR notation.</p>
```
paigecalvert, line 204: ^ added info about passing the node IP(s)

docs/vendor/embedded-overview.mdx
```diff
@@ -216,20 +225,14 @@ Where:
 
 * `LICENSE_FILE` is the customer's license
 * `HOST:PORT` is the host and port of the proxy server
-* `LIST_OF_HOSTS` is the list of hosts to not proxy
-
-For multi-node clusters, the proxy settings are automatically configured when joining additional nodes. When deploying the first node of a multi-node cluster, you should pass the node IP addresses (typically in CIDR notation) to `--no-proxy`.
```
paigecalvert, line 221: >For multi-node clusters, the proxy settings are automatically configured when joining additional nodes.

I had this in a Note block, but then thought this might need more clarification:
* is this just saying that you don't have to tell the new node the host/port of the proxy server?
* When joining additional nodes after installation, would they need to pass the IP(s) of the nodes to the no proxy list somehow?

ajp-io: There are no proxy-related flags for the join command. You just set those on install. That's why on install you have to pass all the node IPs you'll have in the cluster. You won't/can't add those when you join nodes.

ajp-io: My original wording was meant to convey that you won't add NODE_2_IP_ADDRESS when joining node 2. You have to pass that on install. All the proxy settings are set on install. I didn't want people to think, oh I'll just no proxy this node and then set the additional nodes when I join them. You have to do it all up front.

paigecalvert: Added this language back in

ajp-io:
I'm actually with you that that language was needed, so I removed it and I'll merge

PR #2603, opened 2024-08-30
Update testing-about.md

AmberAlston:
We say here that ipv6 is not supported but we do support choosing dual stack network on kind now. Changing wording accordingly


docs/vendor/testing-about.md
```diff
@@ -69,7 +69,7 @@ Compatibility Matrix has the following limitations:
 - Multi-node support is not available for every distribution. For distribution-specific details, see [Supported Compatibility Matrix Cluster Types](/vendor/testing-supported-clusters).
 - ARM instance types are only supported on Cloud Clusters. For distribution-specific details, see [Supported Compatibility Matrix Cluster Types](/vendor/testing-supported-clusters).
 - GPU instance types are only supported on Cloud Clusters. For distribution-specific details, see [Supported Compatibility Matrix Cluster Types](/vendor/testing-supported-clusters).
-- There is no support for IPv6.
+- IPv6 is only supported for kind clusters.
```
jdewinne, line 72: This is maybe not 100% correct. Kind clusters support ipv4 and dual stack. However, Kind clusters do not support ipv6 as a single stack.

AmberAlston: Updated language with latest edit

jdewinne (approved):

PR #2626, opened 2024-09-09
Updated download portal password instructions

divolgin:
Updated generating Download Portal password section to match the new UI

Screenshot 2024-09-09 at 1 10 11 PM


paigecalvert (approved):
thanks @divolgin !

PR #2638, opened 2024-09-13
docusaurus upgrade for security fixes

paigecalvert:


St0rmz1:
Ran locally, I do get some deprecation warnings, but the pages are all rendering ok.
Also confirmed the preview deploy looks good.

PR #2647, opened 2024-09-16
Add default node port range

paigecalvert:


docs/reference/embedded-config.mdx
```diff
@@ -146,6 +146,8 @@ If a Helm extension is removed from the Embedded Cluster Config, the associated
 
 * If you need to install multiple charts in a particular order, set the `order` field to a value greater than or equal to 10. Numbers below 10 are reserved for use by Embedded Cluster to deploy things like a storage provider and the Admin Console. If an `order` is not provided, Helm extensions are installed with order 10. 
 
+* The default service node port range is `80-32767`. For any node port services you include in your Helm extensions, use ports within this range.
+
```
paigecalvert, line 150: ^ It's not necessary to tell people how to edit this default range anymore, right?

ajp-io: Yeah, that's fine. We expanded it so people generally wouldn't need to adjust it. Someone might want to change it, but until that's a significant use case, I don't think it's necessary to doc it.

paigecalvert: 👍 going to close this out

PR #2657, opened 2024-09-18
Update limitations

ajp-io:
Remove air gap limitation, because it's beta and so is the rest of EC. Update kURL migration limitation so it doesn't sound like it's impossible.

### DIFF
--- a/docs/vendor/embedded-overview.mdx
+++ b/docs/vendor/embedded-overview.mdx
@@ -43,7 +43,7 @@ As shown in the diagram above, the Embedded Cluster Config is included in the ap
 
 Embedded Cluster has the following limitations:
 
-* **No automated migration from kURL**: There is no automated migration from an existing kURL instance to Embedded Cluster. There is a relatively simple manual migration procedure for those interested.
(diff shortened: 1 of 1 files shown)


docs/vendor/embedded-overview.mdx
```diff
@@ -43,13 +43,11 @@ As shown in the diagram above, the Embedded Cluster Config is included in the ap
 
 Embedded Cluster has the following limitations:
 
-* **No migration from kURL**: There is not a supported path to migrate an existing kURL instance to Embedded Cluster.
+* **No automated migration from kURL**: There is no automated migration from an existing kURL instance to Embedded Cluster. There is a relatively simple manual migration procedure for those interested.
```
paigecalvert, line 46: @ajp-io should we tell them to reach out to you for the steps?

ajp-io: Sure!

ajp-io: See what you think

paigecalvert (approved):
@ajp-io if you're cool with the edits to the manual migration procedure content, this is good to merge 👍

PR #2528, opened 2024-07-29
WAIT FOR EC GA: Onboarding docs updates

paigecalvert:
KOTS/EC Quick start: https://deploy-preview-2528--replicated-docs-upgrade.netlify.app/vendor/quick-start-embedded

~~Helm only quick start: https://deploy-preview-2528--replicated-docs-upgrade.netlify.app/vendor/quick-start-helm~~

KOTS/EC onboarding: https://deploy-preview-2528--replicated-docs-upgrade.netlify.app/vendor/distributing-workflow

~~Helm only onboarding: https://deploy-preview-2528--replicated-docs-upgrade.netlify.app/vendor/replicated-onboarding~~

PR #2665, opened 2024-09-19
Remove helm only release

paigecalvert:


paigecalvert:
It looks like the whole Quick Start topic is going to need to be replaced in order to fully remove the concept of "helm-only" releases. That was going to be done as part of reorg-ing the docs around Embedded Cluster-first with: https://github.com/replicatedhq/replicated-docs/pull/2528

I'm going to close out this PR and make a smaller one that just removes the Helm-only option from the UI procedures, removes the --chart flag from the main procedure on creating releases with the CLI, but leaves in the --chart flag in the CLI docs so that I don't have to totally rip out the existing quick start yet.

PR #2672, opened 2024-09-23
newer version of EC

grantmiller:
we should also automate this


paigecalvert:
Bumped the version in a few other places: https://github.com/replicatedhq/replicated-docs/pull/2674

PR #2682, opened 2024-09-25
Make Embedded Cluster GA

ajp-io:


paigecalvert (approved):
GA?! A-yo!

ajp-io:
Found out that the beta status is starting to be a problem for some customers like ITRS. And since we're already effectively treating it as GA as as supporting it, and customers are using it as if it's GA and will expect GA-level support and fixes, it's fine to say it's GA.

But we'll have a separate marketing launch for EC when we feel it's ready for a more widespread GA.

PR #2700, opened 2024-09-28
chore(embedded-cluster): list directories and files written

emosbaugh:


docs/partials/embedded-cluster/_requirements.mdx
```diff
@@ -8,4 +8,30 @@
   * The filesystem at `/var/openebs` has 5Gi or more of total space
   * The filesystem at `/tmp` has 5Gi or more of total space
 * (Online installations only) Access to replicated.app and proxy.replicated.com or your custom domain for each
-* Embedded Cluster is based on k0s, so all k0s system requirements and external runtime dependencies apply. See [System requirements](https://docs.k0sproject.io/stable/system-requirements/) and [External runtime dependencies](https://docs.k0sproject.io/stable/external-runtime-deps/) in the k0s documentation.
\ No newline at end of file
+* Embedded Cluster is based on k0s, so all k0s system requirements and external runtime dependencies apply. See [System requirements](https://docs.k0sproject.io/stable/system-requirements/) and [External runtime dependencies](https://docs.k0sproject.io/stable/external-runtime-deps/) in the k0s documentation.
+* Additional directories and files created by the Embedded Cluster include:
```
paigecalvert, line 12: Is there any action that the user needs to take @emosbaugh or was this just informational?

ajp-io: Just information. We have a few directories listed below this with longer explanations. I would remove those directories from this list (e.g., `/var/lib/embedded-cluster`) and take the remaining directories and tack them on after the directories below. Something like "the following directories are also created" and just list them without explanations.

emosbaugh: updated

PR #2703, opened 2024-09-30
Remove the reference to BETA in Air gap telemetry

brianrhea:
Remove the reference to BETA in Air gap telemetry title and update the screenshot to remove the reference there as well.


paigecalvert (approved):
thank you @brianrhea !

PR #2678, opened 2024-09-24
Onboarding docs

paigecalvert:
New quick start: https://deploy-preview-2678--replicated-docs.netlify.app/vendor/quick-start

New onboarding flow: https://deploy-preview-2678--replicated-docs.netlify.app/vendor/replicated-onboarding

(The rest is partials, images, and editing links in other topics)

PR #2721, opened 2024-10-03
Move air gap bundle info to the relevant install topics

paigecalvert:
This PR removes the preexisting topic on "how to download air gap bundles" in favor of putting that info directly in the relevant air gap installation topics. The reasoning for this change is because:
* People are most likely to need information on how to download air gap bundles when they are actually trying to do an air gap install
* When looking for info on how to download air gap bundles, people will need info about one or two types of bundles in particular, and don't need to see info about all the different air gap bundles. This can cause confusion
* If you landed on the topic about downloading air gap bundles, you'd be missing a lot of context and likely get confused about what you are supposed to do with the information

The solution was to do the following:
* Remove the topic on "how to build to air gap bundles"
* Redirect to the topic about how to access/download stuff from the Download Portal: https://deploy-preview-2721--replicated-docs-upgrade.netlify.app/vendor/releases-share-download-portal
* Add the steps on how to build release air gap bundles to a partial: https://github.com/replicatedhq/replicated-docs/pull/2721/files#diff-62506499cfb4131a5a34e3e5d307020bc0d8d2580cae5b5a29a7b3916178fcbc
* Add the steps on how to download air gap licenses to a partial as well: https://github.com/replicatedhq/replicated-docs/pull/2721/files#diff-8ec0735b0c6ed66e64cdd2afd783b42c68aae84985eee33f1db76a53ec9b7cef
* Replace the old xrefs to the topic on "how to build air gap bundles" with these partials instead. That way, the steps are right in the relevant procedures 
* This also involves replacing some content in the updating and CLI install topics in addition to the regular air gap install topics


docs/vendor/releases-download-airgap-bundles.mdx
```diff
@@ -2,69 +2,119 @@ import DownloadPortal from "../partials/kots/_download-portal-about.mdx"
...
-* **`.airgap`**: Assets for a specific release of the application, such as Kubernetes YAML files and Docker images.
-* **`kotsadm.tar.gz`**: Container images for the Replicated KOTS admin console. Required for existing cluster installations in air gap environments.
-* **kURL `.tar.gz`**: Open source components to run the cluster, including the container runtime, Kubernetes, the KOTS admin console, the kURL image registry and a number of other [kURL add-ons](https://kurl.sh/add-ons). Required for embedded cluster installations with kURL in air gap environments. 
+_Air gap bundles_ are required to install applications in air gap environments. Air gap bundles contain the assets needed to install and run applications in an environment that has no outbound internet access.
 
-## Build and Download the Application Bundle {#air-gap-bundle}
+The air gap bundles required for installation vary depending on the Replicated installer that is used:
```
paigecalvert, line 11: updated the overview to explain which bundles are required for which install types

docs/vendor/releases-download-airgap-bundles.mdx
```diff
@@ -2,69 +2,119 @@ import DownloadPortal from "../partials/kots/_download-portal-about.mdx"
...
 
-This section describes how to build and download the application `.airgap` bundle from the Vendor Portal. For information about building and downloading `.airgap` bundles with the Vendor API v3, see [Trigger airgap build for a channel's release](https://replicated-vendor-api.readme.io/reference/channelreleaseairgapbuild) and [Get airgap bundle download URL for the active release on the channel](https://replicated-vendor-api.readme.io/reference/channelreleaseairgapbundleurl) in the Vendor API v3 documentation.
+* **Replicated Embedded Cluster**: Installations with Embedded Cluster require a single Embedded Cluster air gap bundle. The Embedded Cluster air gap bundle contains:
 
-For information about how to download the `.airgap` bundle from the Replicated download portal, see [Download Air Gap Bundles from the Download Portal](#download-portal) below.
+    * An `embedded-cluster` directory with the assets needed to install the infrastructure (Embedded Cluster/k0s and any [extensions](/reference/embedded-config#extensions))
+    * All the assets contained in the air gap bundle for the given application release (application images, manifests, and dependencies)
```
paigecalvert, line 16: ^ are Admin Console images also included? (in the embedded-cluster directory I guess?)

docs/vendor/releases-download-airgap-bundles.mdx
```diff
@@ -2,69 +2,119 @@ import DownloadPortal from "../partials/kots/_download-portal-about.mdx"
...
+* **Replicated kURL**: Installations with kURL require the following air gap bundles:
+    * **Application release air gap bundle**: Contains the assets required to install the given application release (application images, manifests, and dependencies)
+    * **kURL bundle**: Contains the open source components to run the cluster, including the container runtime, Kubernetes, the KOTS Admin Console, the kURL image registry and a number of other [kURL add-ons](https://kurl.sh/add-ons).
+    
+      <details>
+      <summary>Why is the kURL bundle separate from the air gap bundle for the release?</summary>
+
+        * The kURL bundle can get quite large, so this method lets you update your application with a smaller bundle size.
+        * Whereas air gap bundles for an application often need to be rebuilt with each new release, kURL air gap bundles only need to be built when the underlying cluster components or add-ons must be updated.
+      </details>
```
paigecalvert, line 31: ^ found this new Docusaurus feature where you can add expanding text. Feels useful for places where you want to explain why, but you don't want to clutter up the page with too many notes

docs/vendor/releases-download-airgap-bundles.mdx
```diff
@@ -2,69 +2,119 @@ import DownloadPortal from "../partials/kots/_download-portal-about.mdx"
...
+    * **kURL bundle**: Contains the open source components to run the cluster, including the container runtime, Kubernetes, the KOTS Admin Console, the kURL image registry and a number of other [kURL add-ons](https://kurl.sh/add-ons).
+    
+      <details>
+      <summary>Why is the kURL bundle separate from the air gap bundle for the release?</summary>
+
+        * The kURL bundle can get quite large, so this method lets you update your application with a smaller bundle size.
+        * Whereas air gap bundles for an application often need to be rebuilt with each new release, kURL air gap bundles only need to be built when the underlying cluster components or add-ons must be updated.
+      </details>  
+
+## Download the Embedded Cluster Bundle {#ec-bundle}
```
paigecalvert, line 33: New section with info about downloading the EC bundle using the app service

docs/vendor/releases-download-airgap-bundles.mdx
```diff
@@ -2,69 +2,119 @@ import DownloadPortal from "../partials/kots/_download-portal-about.mdx"
...
+
+To build and download the air gap bundle for a release:
 
 1. In the [Vendor Portal](https://vendor.replicated.com), go to **Channels**. On the target channel, click **Release history**.
 
     <img alt="Release history link on a channel card" src="/images/release-history-link.png" width="400px"/>
 
     [View a larger version of this image](/images/release-history-link.png)
 
-1. On the **Release history** page, click **Build** to build the `.airgap` bundle.
```
paigecalvert, line 29: ^ Previously, I was leaning pretty heavily on "`.airgap`" to talk about the release air gap bundle as opposed to the Admin Console or kURL bundles. Now that the EC bundle also uses the `.airgap` extension, felt like it would be better to spell out which bundle we are talking about in each case

docs/vendor/releases-download-airgap-bundles.mdx
```diff
@@ -2,69 +2,119 @@ import DownloadPortal from "../partials/kots/_download-portal-about.mdx"
...
-The `kotsadm.tar.gz` air gap bundle for each version of KOTS is available for download on the [Releases](https://github.com/replicatedhq/kots/releases) page in the kots repository in GitHub.
+To download the KOTS Admin Console bundle, do one of the following:
 
-You can also download the `kotsadm.tar.gz` air gap bundle from the Replicated download portal to ensure its compatibility with the application `.airgap` bundle. For more information, see [Download Air Gap Bundles from the Download Portal](#download-portal) below. 
+* Download the `kotsadm.tar.gz` air gap bundle on the [Releases](https://github.com/replicatedhq/kots/releases) page in the kots repository in GitHub.
 
-## Download the kURL Bundle {#installer-bundle}
+   :::note
+   The version of the `kotsadm.tar.gz` air gap bundle used must be compatible with the version of the `.airgap` bundle for the given application release.
+   :::
```
paigecalvert, line 109: ^ Chose to stick with `.airgap` here to talk about the release air gap bundle because I wanted to be clear which two files have to be compatible version wise. And, since this isn't in the context of Embedded Cluster stuff, figured it would be fine

docs/vendor/releases-download-airgap-bundles.mdx
```diff
@@ -115,25 +165,27 @@ To download the kURL bundle for channels other than Stable:
...
 
-## Download Air Gap Bundles from the Download Portal {#download-portal}
+## Download Bundles from the Download Portal {#download-portal}
 
 <DownloadPortal/>
 
-To download air gap bundles from the download portal:
+:::note
+Air gap bundles for Embedded Cluster are not available in the Download Portal. See [Download the Embedded Cluster Bundle](#ec-bundle) above.
+:::
```
paigecalvert, line 174: ^ added this note

docs/vendor/releases-download-airgap-bundles.mdx
ajp-io: I don't love the organization of this page. Started thinking about the fact that the EC download is above the build step (and separated from the other download steps), but I think it's more than that. If I came to this page to get started with downloading an air gap bundle, I think I would feel a bit confused. I'd want to see 1) how do I build the air gap bundle, and then 2) how do I download the bundle(s) that I need for my installation method. This feels a bit strung out and you have to piece things together (KOTS bundle, kURL bundle, app bundle, etc.).

My thought is that something that targets each installation method will be easier to use.

ajp-io: I remember the topic of "how to get started with air gap" coming up recently. Do we have a page on how to do that (other than the `builder` field in HelmChart)?

paigecalvert: >I remember the topic of "how to get started with air gap" coming up recently. Do we have a page on how to do that (other than the builder field in HelmChart)?

Yeah as a first step that would be coming as part of the updating onboarding docs: https://github.com/replicatedhq/replicated-docs/pull/2678 (there's a new section on how to support air gap). Want to look at that onboarding together early next week maybe? Would be good to get it merged in 

>My thought is that something that targets each installation method will be easier to use.

The individual installation docs do all cover this. So this page could be deleted and we could just leave it up to the install steps to tell people how to download the air gap bundles that they'll need. (I think those install topics currently link here, so that would just involve copying over the steps).

The goal of this topic as I see it was to help give a map on all the different air gap bundles we have, which is a lot to wrap your head around. I could see something that like being valuable as part of more content on how to support air gap installs, when that exists. But maybe it's not useful as is outside of the context of a larger workflow.

paigecalvert:
Note from convo with Alex this morning:
* We are going to remove this topic all together in favor of putting the info directly in the air gap installation topics. For example, the info about how to download the air gap bundle for the release will appear as a partial in the existing cluster KOTS and kURL air gap install topics.
* We also discussed making this information an actual step in the procedure rather than a prereq. This gives more space to provide screenshots/additional info, and makes it harder to miss
* The reasoning for this decision is that it can be overwhelming to see all the air gap bundle info together on a single topic. And, it's most likely that the only time people need this info is when they are actually getting ready to do an air gap install

sidebars.js
```diff
@@ -320,16 +322,6 @@ const sidebars = {
...
-            {
-              type: 'category',
-              label: 'Downloading and Sharing Assets for KOTS Releases',
-              items: [
-                'vendor/releases-download-airgap-bundles',
-                'vendor/licenses-download',
-                'vendor/releases-share-download-portal',
-                'vendor/releases-configvalues',
-              ],
-            },
```
paigecalvert, line 332: ^ removed this section.

* releases-download-airgap-bundles was deleted
* licenses-download was moved to the Customer Licenses section
* releases-share-download-portal was moved to the Channels and Releases section
* releases-configvalues was moved to the CLI install section (the only workflow where vendors need it)

PR #2730, opened 2024-10-10
Add documentation for download portal session length

alicenstar:
This comes from a question a vendor posed here: https://github.com/replicated-collab/grammatech-replicated/issues/50#issuecomment-2405787343

I'm open to better ways to present/phrase this. Wasn't sure exactly where to put it.

### DIFF
--- a/docs/vendor/releases-share-download-portal.md
+++ b/docs/vendor/releases-share-download-portal.md
@@ -81,3 +81,7 @@ and preview your customer's experience.
 1. Click the download button to download each asset.   
 
 1. To share installation files with a customer, send the customer their unique link and password for the Download Portal.
+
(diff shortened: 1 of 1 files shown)


paigecalvert (approved):
@alicenstar let me know what you think of this edit...I moved it up the Limitations section with a sentence that says the session length is not configurable (which I guess would be real "limitation" here). I was thinking this might make sense since it looked like the vendor was asking about changing the session details. If the feature is every added, then this limitation would be removed in favor of more detailed docs on how to configure it. WDYT?

paigecalvert:
@alicenstar going to merge this one. LMK if it shouldn't be and I can revert/edit as needed.

PR #2759, opened 2024-10-21
Remove 1.15

paigecalvert:


docs/release-notes/rn-embedded-cluster.md
```diff
@@ -10,6 +10,44 @@ This topic contains release notes for the [Replicated Embedded Cluster](/vendor/
 
 Additionally, these release notes list the versions of Kubernetes and Replicated KOTS that are available with each version of Embedded Cluster.
 
+## 1.15.0 - Removed
+
+:::important
+Embedded Cluster 1.15.0 has been removed and is not available for use.
```
ajp-io, line 16: I think it might be worth a vague mention as well as a note that it'll continue to work for anyone already using it.
```suggestion
Embedded Cluster 1.15.0 has been removed and is not available for use because of issues with upgrades. It continues to work for anyone already using it.
```

PR #2760, opened 2024-10-21
Update kURL to clarify version expectation

AmberAlston:
The current link to "creating a kURL installer" doesn't actually help me find the info I need. I need the kURL installer ID slug. Outside of trying to use kURL.sh to get this ID, I need to go find it in my installers history. Still not the easiest to discover ,but a small improvement towards pointing me in the right direction. 

I'm not sure if this page is deep linkable or not since it's context specific, but this is an example of what the URL for the right page looks like https://vendor.replicated.com/apps/{app-id}/channels/{channel-id)/kubernetes-installers?page=0

### DIFF
--- a/docs/vendor/testing-supported-clusters.md
+++ b/docs/vendor/testing-supported-clusters.md
@@ -267,7 +267,7 @@ Compatibility Matrix supports creating [kURL](https://kurl.sh) clusters.
   </tr>
   <tr>
     <th>Supported kURL Versions</th>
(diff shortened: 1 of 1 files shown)


paigecalvert (approved):
thanks @AmberAlston ! Approving but not merging since it's marked as Draft (wasn't sure if you were wanting to hold it)

AmberAlston:
Didn't mean to leave as draft, oops. Thanks for the approval @paigecalvert !

PR #2761, opened 2024-10-22
WAIT FOR RELEASE: Doc new --cidr flag

ajp-io:
Adds docs for the new --cidr flag, which replaces the --pod-cidr and --service-cidr flags. Those still exist, but they're hidden, and we recommend this one flag now. (This flag just asks for one CIDR and we split it between the Pods and Services, which is easier than asking people for two ranges.)


paigecalvert (dismissed):
Cool, looks good

In the release notes for the version that includes this new flag, we can explain the details about how the new cidr flag replaces the two separate pod/service flags and how those are now hidden (so people understand where they went)

ajp-io:
Good idea

-------------

*Alex Parker*
Staff Product Manager

ajp-io:
@paigecalvert I made another slight change, so this needs a re-review.

paigecalvert (dismissed):

paigecalvert (approved):

PR #2771, opened 2024-10-24
doc: document embedded-cluster support-bundle command

ricardomaraschini:
this is a temporary PR as XXX needs to be replaced with the actual embedded-cluster version yet to be released.

### DIFF
--- a/docs/partials/support-bundles/_ec-support-bundle-intro.mdx
+++ b/docs/partials/support-bundles/_ec-support-bundle-intro.mdx
@@ -2,4 +2,4 @@ Embedded Cluster includes a default support bundle spec that collects both host-
 
 The host-level information provided is useful for troubleshooting failures related to host configuration like DNS, networking, or storage problems. Cluster-level information includes details about the components provided by Replicated, such as the Admin Console and Embedded Cluster operator that manage install and upgrade operations. If the cluster has not installed successfully and cluster-level information is not available, then it is excluded from the bundle.
 
-In addition to the host- and cluster-level details provided by the default Embedded Cluster spec, support bundles generated for Embedded Cluster installations also include app-level details provided by application-specific specs.
+In addition to the host- and cluster-level details provided by the default Embedded Cluster spec, support bundles generated for Embedded Cluster installations also include app-level details provided by any custom support bundle specs that you included in the application release.
--- a/docs/partials/support-bundles/_ec-support-bundle-intro.mdx
+++ b/docs/partials/support-bundles/_ec-support-bundle-intro.mdx
@@ -1,5 +1,5 @@
-Embedded Cluster includes a default support bundle spec that can be used to collect both cluster- and host-level information.
+Embedded Cluster includes a default support bundle spec that collects both host- and cluster-level information.
 
 The host-level information provided is useful for troubleshooting failures related to host configuration like DNS, networking, or storage problems. Cluster-level information includes details about the components provided by Replicated, such as the Admin Console and Embedded Cluster operator that manage install and upgrade operations. If the cluster has not installed successfully and cluster-level information is not available, then it is excluded from the bundle.
 
-After generating a support bundle using the default spec, you can use the bundle for troubleshooting or provide it to Replicated for assistance.
+In addition to the host- and cluster-level details provided by the default Embedded Cluster spec, support bundles generated for Embedded Cluster installations also include app-level details provided by application-specific specs.
--- a/docs/partials/support-bundles/_ec-support-bundle-intro.mdx
+++ b/docs/partials/support-bundles/_ec-support-bundle-intro.mdx
@@ -1,5 +1,5 @@
 Embedded Cluster includes a default support bundle spec that collects both host- and cluster-level information.
 
-The host-level information provided is useful for troubleshooting failures related to host configuration like DNS, networking, or storage problems. Cluster-level information includes details about the components provided by Replicated, such as the Admin Console and Embedded Cluster operator that manage install and upgrade operations. If the cluster has not installed successfully and cluster-level information is not available, then it is excluded from the bundle.
+The host-level information is useful for troubleshooting failures related to host configuration like DNS, networking, or storage problems. Cluster-level information includes details about the components provided by Replicated, such as the Admin Console and Embedded Cluster operator that manage install and upgrade operations. If the cluster has not installed successfully and cluster-level information is not available, then it is excluded from the bundle.
 
 In addition to the host- and cluster-level details provided by the default Embedded Cluster spec, support bundles generated for Embedded Cluster installations also include app-level details provided by any custom support bundle specs that you included in the application release.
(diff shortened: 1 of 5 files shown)


ajp-io:
@paigecalvert what are your thoughts about "version before and after x.y.z" here? Should we do that or just document the latest? I can see a case for doing both since many people still do use the older versions.

paigecalvert:
> @paigecalvert what are your thoughts about "version before and after x.y.z" here? Should we do that or just document the latest? I can see a case for doing both since many people still do use the older versions.

@ajp-io I was thinking the same: let's leave in both sets of steps for now and then we can phase out the earlier steps when enough people have upgraded. The other thing is that the steps are different enough that it feels easier to just write them both out rather than trying to make a "In versions earlier than X.X" note or something.

docs/partials/support-bundles/_generate-bundle-ec.mdx
```diff
@@ -1,5 +1,35 @@
+There are different steps to generate a support bundle depending on the version of Embedded Cluster installed.
+
+### For Versions XXX and Later
+
+For Embedded Cluster XXX and later, you can run the Embedded Cluster `support-bundle` command to generate a support bundle.
+
+The `support-bundle` command uses the default Embedded Cluster support bundle spec to collect both cluster- and host-level information. It also automatically includes any application-specific support bundle specs in the generated bundle.
```
paigecalvert, line 7: >It also automatically includes any application-specific support bundle specs in the generated bundle. 

^ This was a guess

docs/partials/support-bundles/_generate-bundle-ec.mdx
```diff
@@ -1,5 +1,35 @@
...
+
+The `support-bundle` command uses the default Embedded Cluster support bundle spec to collect both cluster- and host-level information. It also automatically includes any application-specific support bundle specs in the generated bundle. 
+
 To generate a support bundle:
 
+1. SSH onto a controller node.
+
+   :::note
+   You can SSH onto a worker node to generate a support bundle that contains information specific to that node. However, when run on a worker node, the `support-bundle` command does not capture cluster-wide information.
+   :::
```
paigecalvert, line 15: ^ Could consider removing this note all together to cut down on noise if you all don't think there are valid use cases for executing the command on a worker node. I.e., if there are no real barriers or cons to just running the command on a controller node

docs/vendor/embedded-overview.mdx
```diff
@@ -380,3 +367,9 @@ toolkit:
    - name: CONTAINERD_SOCKET
      value: /run/k0s/containerd.sock
 ```     
+
+## Troubleshoot with Support Bundles
```
paigecalvert, line 371: Moved this to its own section at the bottom

docs/vendor/support-bundle-embedded.mdx
```diff
@@ -1,19 +1,16 @@
 import EmbeddedClusterSupportBundle from "../partials/support-bundles/_generate-bundle-ec.mdx"
+import SupportBundleIntro from "../partials/support-bundles/_ec-support-bundle-intro.mdx"
 
-# Generating Host Bundles for Embedded Cluster
+# Generating Support Bundles for Embedded Cluster
```
paigecalvert, line 4: ^ Updated the title since it's more than host-level info

paigecalvert (dismissed):

docs/partials/support-bundles/_generate-bundle-ec.mdx
ajp-io: It's true, but I wonder if this should be in the intro since the intro sounds like it's all about our default spec and not the vendor's.

docs/partials/support-bundles/_generate-bundle-ec.mdx
ajp-io: I'd ask CRE, but I think controller is the best default, but if the issue lies on a worker, then you want to run it on the worker to get its host-level info. Any host-level info is specific to the node your run it on. Only the cluster-level stuff would be cluster-wide.

docs/partials/support-bundles/_generate-bundle-ec.mdx
```diff
@@ -1,5 +1,35 @@
+There are different steps to generate a support bundle depending on the version of Embedded Cluster installed.
+
+### For Versions XXX and Later
+
+For Embedded Cluster XXX and later, you can run the Embedded Cluster `support-bundle` command to generate a support bundle.
```
ajp-io, line 5: This is likely 1.17, and we could add it when we merge or add as a placeholder for now

ajp-io:
@paigecalvert this is now released, but I had a few comments in my last review.

paigecalvert:
> @paigecalvert this is now released, but I had a few comments in my last review.

Ah okay thanks, missed those

docs/partials/support-bundles/_generate-bundle-ec.mdx
paigecalvert: Updated. Kept this sentence here but also updated the intro to clarify that you get app-level info too in the generated bundle

docs/partials/support-bundles/_generate-bundle-ec.mdx
paigecalvert: Makes sense, sounds like it's worth it to keep this. I can follow up with CRE after this is merged to see if they have anything to add

docs/partials/support-bundles/_ec-support-bundle-intro.mdx
```diff
@@ -0,0 +1,5 @@
+Embedded Cluster includes a default support bundle spec that collects both host- and cluster-level information.
+
+The host-level information provided is useful for troubleshooting failures related to host configuration like DNS, networking, or storage problems. Cluster-level information includes details about the components provided by Replicated, such as the Admin Console and Embedded Cluster operator that manage install and upgrade operations. If the cluster has not installed successfully and cluster-level information is not available, then it is excluded from the bundle.
+
+In addition to the host- and cluster-level details provided by the default Embedded Cluster spec, support bundles generated for Embedded Cluster installations also include app-level details provided by application-specific specs.
```
paigecalvert, line 5: ^ updated this paragraph so that it doesn't seem like you're missing out on app stuff. I kept it high level to avoid getting into how for earlier versions you would have to add --load-cluster-specs to the command etc etc

paigecalvert (approved):

docs/partials/support-bundles/_ec-support-bundle-intro.mdx
ajp-io: I'm fine to merge when this releases, but I don't think a vendor would read "provided by application-specific specs" and realize that meant "this will also collect the stuff you asked for in the support bundle you included in your release."

paigecalvert: Yeah, updated

paigecalvert (approved):

PR #2773, opened 2024-10-24
update download portal docs

alicenstar:
A recent change updated the sidebar in the download portal to say "Embedded Kubernetes" instead of "Embedded cluster" (when referencing the kURL tab)

Preview: https://deploy-preview-2773--replicated-docs.netlify.app/vendor/releases-share-download-portal


paigecalvert (approved):
thanks @alicenstar ! only edit was to also update the other tab to say "Bring my own Kubernetes" instead of "my own cluster"

PR #2788, opened 2024-10-29
Create topic for adding and managing nodes with EC

paigecalvert:
New topic for ~~Adding and Managing Nodes~~ Managing Multi-Node Clusters: https://deploy-preview-2788--replicated-docs-upgrade.netlify.app/enterprise/embedded-manage-nodes

Added some conceptual overview info about managing multi-node clusters to the main Embedded Cluster topic, with xrefs: https://deploy-preview-2788--replicated-docs-upgrade.netlify.app/vendor/embedded-overview


docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,88 @@
+import JoinNode from "../partials/embedded-cluster/_add-nodes.mdx"
+import JoinNodeAirGap from "../partials/embedded-cluster/_add-nodes-air-gap.mdx"
+
+# Adding and Managing Nodes with Embedded Cluster
```
paigecalvert, line 4: Moved node procedures to their own topic

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+## Add Nodes to Air Gap Clusters (Beta) {#add-nodes-air-gap}
+
+You can add nodes to an air gap cluster. This involves downloading the Embedded Cluster assets to each node, copying a join command from the Admin Console, and running the join command on each node.
+
+To add nodes to an air gap cluster after successfully installing the first node:
+
+<JoinNodeAirGap/>
+
+## Enable High Availability for Multi-Node Clusters (Alpha) {#ha}
```
paigecalvert, line 26: ^ put HA procedure here too. Could also be in its own topic if you feel like it's not something people would associate with "managing nodes"

docs/enterprise/installing-embedded-air-gap.mdx
```diff
@@ -123,7 +123,9 @@ To install with Embedded Cluster in an air gap environment:
 
 1. On the login page, enter the Admin Console password that you created during installation and click **Log in**.
 
-1. On the **Nodes** page, you can view details about the machine where you installed, including its node role, status, CPU, and memory. Optionally, add nodes before deploying the application. Click **Continue**.
+1. On the **Nodes** page, you can view details about the machine where you installed, including its node role, status, CPU, and memory. 
+
+     Optionally, add nodes to the cluster before deploying the application. For more information about joining nodes, see [Adding and Managing Nodes with Embedded Cluster](/enterprise/embedded-manage-nodes). Click **Continue**.
```
paigecalvert, line 128: ^ rather than having a separate heading for adding nodes in the installation topics, I added a link to the new topic from the relevant step in both installation procedures

docs/vendor/embedded-overview.mdx
```diff
@@ -164,21 +180,27 @@ When you skip host preflight checks, the Admin Console still runs any applicatio
 Skipping host preflight checks is _not_ recommended for production installations.
 :::
 
-## (Optional) Serve Installation Assets Using the Vendor API
+## About Managing Nodes with Embedded Cluster
```
paigecalvert, line 183: ^ added this conceptual overview section in the main EC topic to help point people to this information

docs/vendor/embedded-overview.mdx
```diff
@@ -132,6 +132,22 @@ To install with Embedded Cluster, you can follow the customer-specific instructi
 
 [View a larger version of this image](/images/embedded-cluster-install-dialog.png)
 
+### (Optional) Serve Installation Assets Using the Vendor API
```
paigecalvert, line 135: Just moved this to nest it under the Install H2 (it wasn't fitting in nicely with the other h2s)

sidebars.js
```diff
@@ -214,6 +214,7 @@ const sidebars = {
                   items: [
                     'vendor/embedded-overview',
                     'reference/embedded-config',
+                    'enterprise/embedded-manage-nodes',
```
paigecalvert, line 217: For now, added this to the main "vendor" Embedded Cluster section rather than putting it in the Installing with Embedded Cluster section. This will be easier to place once we do the reorg and have the single Embedded Cluster section, no longer split between vendor and enterprise tasks

My thought now is that most of the EC stuff is under this main vendor section. And adding nodes is not "strictly" just related to installation, so I thought it might be harder to find nested under install

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,121 @@
+# Adding and Managing Nodes with Embedded Cluster 
+
+The topic describes managing nodes in clusters created with Replicated Embedded Cluster, including how to add or reset nodes and enable high-availability for multi-node clusters.
+
+## Add Nodes to a Cluster (Beta) {#add-nodes}
```
paigecalvert, line 5: Created a single procedure the includes the online and air gap steps, which seem to be essentially the same (besides the fact that you have to make sure to download the air gap bundle on the node for air gap installs).

Pulled the online steps from https://docs.replicated.com/vendor/embedded-overview#add-nodes and the air gap steps from https://docs.replicated.com/enterprise/installing-embedded-air-gap

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+You can add nodes to create a multi-node cluster in online (internet-connected) and air-gapped environments. The Admin Console provides the join command that you use to join nodes to the cluster.
+
+When adding nodes, you select one or more roles for that node, depending on which roles are defined in the Embedded Cluster Config. For more information about defining node roles, see [roles](/reference/embedded-config#roles) in _Embedded Cluster Config_.
```
paigecalvert, line 9: ^ note: could consider having "define node roles" as the first steps of this procedure, since we're not worrying about splitting the docs between vendor and enterprise tasks anymore. LMK what you think

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+To reset a node:
+
+1. SSH onto the machine. Ensure that the Embedded Cluster binary is still available on that machine.
+
+1. Run the following command to reset the node and automatically reboot the machine to ensure that transient configuration is also reset:
+
+    ```
+    sudo ./APP_SLUG reset
+    ```
```
paigecalvert, line 116: ^ also removed the `--reboot` flag from this command

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+When adding nodes, you select one or more roles for that node, depending on which roles are defined in the Embedded Cluster Config. For more information about defining node roles, see [roles](/reference/embedded-config#roles) in _Embedded Cluster Config_.
+
+:::note
+Multi-node clusters are not highly available by default. For information about enabling high availability, see [Enable High Availability for Multi-Node Clusters (Alpha)](#ha) below.
+:::
+
+To add nodes to a cluster:
+
+1. Get the join command from the Admin Console:
```
ajp-io, line 17: Might be worth mentioning or also outlining that you can add nodes during the initial install. We put you on the cluster management page automatically. That's likely the more common way people do this.

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,121 @@
...
+   
+   1. In the **Add a Node** dialog, select one or more roles for the new node that you will join. If no roles are defined in the Embedded Cluster Config, the role selection will not appear and only the join command is displayed. For more information about defining node roles, see [roles](/reference/embedded-config#roles) in _Embedded Cluster Config_.
+
+       <img alt="Add node page in the Admin Console" src="/images/admin-console-add-node.png" width="600px"/>
+
+       [View a larger version of this image](/images/admin-console-add-node.png)
+     
+    1. Copy the join command.
+
+1. SSH onto the machine you want to join to the cluster.
```
ajp-io, line 31: People use air gap installs for different levels of air gapping. But in a true air gap env, you won't be able to run the command to download the assets. You'd need to download them elsewhere and move them over. Maybe we should be less prescriptive (SSH and run this command) and instead say that you run the command to get the assets and ensure they get to the machine you intend to join.

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+High availability is supported with Embedded Cluster 1.4.1 or later.
+
+### Limitations
+
+High availability has the following limitations:
+
+* The `--enable-ha` flag serves as a feature flag during the Alpha phase. In the future, the prompt about migrating to high availability will display automatically if the cluster is not yet HA and you are adding the third or more controller node. 
+
+* Support bundles are now stored in rqlite instead of MinIO because MinIO is no longer deployed. We’ve successfully stored 100 MB support bundles (which is much larger than most support bundles), but bundles over 100 MB can cause rqlite to crash and restart. We are considering making support bundles ephemeral instead of storing them to address this issue, because support bundles are rarely needed long after they’re generated.
```
ajp-io, line 76: This makes sense to someone migrating from kURL, but we should probably not mention MinIO and just say bundles are stored in rqlite and max out around 100 MB. Also remove the ephemeral part since that's just something we're considering.

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,121 @@
...
+To create a multi-node HA cluster:
+
+1. Set up a cluster with at least two controller nodes. You can do an online (internet-connected) or air gap installation. For more information, see [Online Installation with Embedded Cluster](/enterprise/installing-embedded) or [Air Gap Installation with Embedded Cluster](/enterprise/installing-embedded-air-gap).
+
+1. SSH onto a third node that you want to join to the cluster as a controller.
+
+1. Run the join command provided in the Admin Console **Cluster Management** tab and pass the `--enable-ha` flag. For example:
+
+     ```bash
+     sudo ./APP_SLUG join `--enable-ha` 10.128.0.80:30000 tI13KUWITdIerfdMcWTA4Hpf
```
ajp-io, line 89:
```suggestion
     sudo ./APP_SLUG join --enable-ha 10.128.0.80:30000 tI13KUWITdIerfdMcWTA4Hpf
```

docs/enterprise/embedded-manage-nodes.mdx
ajp-io: I think I like that idea. Because I like the idea that if I want to start doing multi-node, I go to one place and figure it out. So instead of having to backtrack to some other doc about node roles, node roles are only relevant for multi-node, so mention them here when someone starts trying multi-node.

docs/enterprise/embedded-manage-nodes.mdx
ajp-io: I think it makes sense since HA is only relevant for multi-node

ajp-io: I might reimagine this page as "multi-node clusters" rather than just "adding nodes" (though it amounts to about the same thing here)

docs/vendor/embedded-overview.mdx
```diff
@@ -164,21 +180,27 @@ When you skip host preflight checks, the Admin Console still runs any applicatio
...
+### Adding Nodes
 
-1. If you have not done so already, create an API token for the Vendor API. See [Using the Vendor API v3](/reference/vendor-api-using#api-token-requirement).
+Users can add nodes to a cluster with Embedded Cluster from the Admin Console. The Admin Console provides the join command used to add nodes to the cluster.
 
-1. Call the [Get an Embedded Cluster release](https://replicated-vendor-api.readme.io/reference/getembeddedclusterrelease) endpoint to download the assets needed to install your application with Embedded Cluster. Your customers must take this binary and their license and copy them to the machine where they will install your application.
+For more information, see [Adding and Managing Nodes with Embedded Cluster](/enterprise/embedded-manage-nodes).
 
-     Note the following:
+### Defining Node Roles
```
ajp-io, line 193: This makes more sense above "add nodes" IMO because once need to add roles to your EC Config if you want to use them when adding nodes.

sidebars.js
ajp-io: IMO this might be why this should be a section about multi-node rather than about adding nodes. adding nodes is very install-centric wording, whereas multi-node is all about supporting multi-node clusters generally.

docs/enterprise/installing-embedded.mdx
```diff
@@ -85,7 +85,9 @@ To install an application with Embedded Cluster:
 
 1. On the login page, enter the Admin Console password that you created during installation and click **Log in**.
 
-1. On the **Nodes** page, you can view details about the VM where you installed, including its node role, status, CPU, and memory. Optionally, add nodes before deploying the application. Click **Continue**.
+1. On the **Nodes** page, you can view details about the machine where you installed, including its node role, status, CPU, and memory. 
+
+     Optionally, add nodes to the cluster before deploying the application. For more information about joining nodes, see [Adding and Managing Nodes with Embedded Cluster](/enterprise/embedded-manage-nodes). Click **Continue**.
```
ajp-io, line 90: I mentioned this somewhere else, how you can do this on install too. I don't know if this changes my thoughts in the other comment or not.

docs/enterprise/installing-embedded-air-gap.mdx
ajp-io: I don't know what is best, but this feels fine

docs/enterprise/embedded-manage-nodes.mdx
paigecalvert: Ah good call

docs/enterprise/embedded-manage-nodes.mdx
paigecalvert: Gotcha that's helpful context, thanks

docs/enterprise/embedded-manage-nodes.mdx
paigecalvert: Reframing the page around multi-node makes sense to me. Only minor hiccup could be the "Reset a Node" steps, which I could see being useful outside a multi-node context, but I think it's fine for now if that lives under multi-node. We'll ultimately want to have a separate reference page for the EC binary `reset` command anyway

ajp-io: Might be best to move reset to the overview page or somewhere else, because it's not tied to multi-node. People reset single nodes all the time.

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Add Nodes to a Cluster (Beta) {#add-nodes}
+
+You can add nodes to create a multi-node cluster in online (internet-connected) and air-gapped environments. The Admin Console provides the join command that you use to join nodes to the cluster.
+
+:::note
+Multi-node clusters are not highly available by default. For information about enabling high availability, see [Enable High Availability for Multi-Node Clusters (Alpha)](#ha) below.
+:::
+
+To add nodes to a cluster:
```
paigecalvert, line 17: @ajp-io tweaked this procedure to provide info about both paths: either adding nodes during install or post-install. It's a little unusual but I think it works in this case, especially since the experience is pretty guided by the Admin Console UI anyway. Curious what you think.

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+1. Do one of the following to get the join command from the Admin Console:
+
+   1. If you have not yet installed the application, follow the steps in [Online Installation with Embedded Cluster](/enterprise/installing-embedded) or [Air Gap Installation with Embedded Cluster](/enterprise/installing-embedded-air-gap) to install. A **Nodes** screen is displayed as part of the installation flow in the Admin Console that allows you to choose a node role and copy the relevant join command.
+
+   1. Otherwise, log in to the Admin Console. Update the instance to the new release that you promoted that defines the `roles` key in the Embedded Cluster Config. For more information, see [Performing Updates in Embedded Clusters](/enterprise/updating-embedded). Then, go to **Cluster Management > Add node** at the top of the page.
```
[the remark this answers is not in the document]
paigecalvert: I didn't, but I can!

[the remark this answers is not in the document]
paigecalvert: Okay just did the following to test this:
1. Install a release containing an EC Config with no `roles` defined
2. Promote a new release containing an EC Config with a couple custom `roles` defined
3. In the Admin Console, before upgrading to the new version, confirmed that those new roles did _not_ appear in the **Cluster Management > Add node** pop up
4. Upgraded to the new version, then confirmed that the new roles did appear in the **Cluster Management > Add node** pop up

Was this what you were hoping to confirm?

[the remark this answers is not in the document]
ajp-io: yep, sounds good. I'm glad it works that way

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+            <img alt="Add node page in the Admin Console" src="/images/admin-console-add-node.png" width="600px"/>
+
+            [View a larger version of this image](/images/admin-console-add-node.png)
+   
+1. Either on the Admin Console **Nodes** screen that is displayed during installation or in the **Add a Node** dialog, select one or more roles for the new node that you will join. Copy the join command.
+
+     :::note
+     If no [roles](/reference/embedded-config#roles) are defined in the Embedded Cluster Config, all new nodes joined to the cluster are controller nodes by default.
+     :::
```
paigecalvert, line 45: ^ Rewrote this note to make it sound a little less negative/deemphasize it. Previously, it said "no roles are shown and only the join command is displayed". Now, it explains the default behavior

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+     :::note
+     Roles are not updated or changed after a node is added. If you need to change a node’s role, reset the node and add it again with the new role.
+     :::
```
paigecalvert, line 49: ^ Moved this note about changing to node role to this step where people would actually be assigning the role. Previously, this note appeared in the EC Config reference topic, where it was on-topic, but not the most useful when you are just learning about how to define roles

docs/reference/embedded-config.mdx
```diff
@@ -71,17 +71,21 @@ For a full list of versions, see the Embedded Cluster [releases page](https://gi
...
+You can optionally define node roles in the Embedded Cluster Config. If nodes roles are defined, users access the Admin Console to assign one or more roles to a node when it is joined to the cluster. For more information about assigning roles to nodes, see [Managing Multi-Node Clusters with Embedded Cluster](/enterprise/embedded-manage-nodes).
 
-:::note
-Roles are not updated or changed after a node is added. If you need to change a node’s role, reset the node and add it again.
-:::
+For multi-node clusters, roles can be useful when combined with [labels](#labels) for the purpose of assigning specific application workloads to nodes. For example, if your application has graphics processing unit (GPU) workloads, you might choose to define a role that is assigned to the node where those workloads run. Or, you could assign a role to a node that is dedicated to running a specific application workoad that is resource-intensive.
+
+If you do not define node roles, all nodes added to the cluster are assigned the controller role. Controller nodes run the Kubernetes control plane and can also run other workloads, such as application or Replicated KOTS workloads.
```
paigecalvert, line 78: ^ Added this info about the default behavior to explain how it works if you don't add roles

docs/reference/embedded-config.mdx
```diff
@@ -71,17 +71,21 @@ For a full list of versions, see the Embedded Cluster [releases page](https://gi
...
+
+Embedded Cluster supports defining _controller_ and _custom_ roles. For more information, see the sections below.
 
 ### controller
 
-The controller role is required in any cluster. Nodes with this role are “controller workers” because they run the control plane and can run other workloads too. The first node in a cluster will always have the controller role because a cluster needs a control plane. Any node that doesn't have the controller role is a worker node.
+Controller nodes run the Kubernetes control plane and can also run other workloads, such as application or Replicated KOTS workloads. For this reason, nodes with the controller role are considered _controller workers_.
+
+Defining one or more controller roles can be useful in highly-available clusters with three or more controller nodes if you want to assign a specific workload to one of the nodes. For more information about high availability with Embedded Cluster, see [Managing Multi-Node Clusters with Emebdded Cluster](/enterprise/embedded-manage-nodes).
```
paigecalvert, line 86: ^ @ajp-io  Could use your input on the use case(s) for defining controller roles. With all new nodes being controller nodes by default anyway, it seemed like the only reason I'd need/want to define a specific controller role is essentially the same as my use cases for the custom worker roles: label it so I can assign a specific workload to it. And maybe I'd going with controller roles because I only have a three node cluster and they all need to be controllers.

Are there other reasons I might want to label a controller node?

docs/reference/embedded-config.mdx
```diff
@@ -94,7 +98,9 @@ spec:
 
 ### custom
 
-You can define custom roles for other purposes in the cluster. This is particularly useful when combined with labels.
+Custom roles can be assigned to any non-controller, or _worker_, nodes in the cluster. Worker nodes can run any workloads that are deployed to the cluster by Embedded Cluster, such as application or Replicated KOTS workloads. Unlike controller nodes, worker nodes do not run the Kubernetes control plane.
+
+Creating custom roles for worker nodes (as opposed to creating controller roles) can be useful if you need to have one or more worker nodes in your cluster that are dedicated to only running a specific application workload.
```
paigecalvert, line 103: ^ added a use cases description of why you'd want to do a custom role instead of a controller role. I figure it's "this node should run this app workload and only this app workload"

docs/vendor/embedded-overview.mdx
```diff
@@ -164,21 +180,27 @@ When you skip host preflight checks, the Admin Console still runs any applicatio
...
+This section describes managing nodes in multi-node clusters created with Embedded Cluster.
 
-To serve Embedded Cluster installation assets with the Vendor API:
+### Defining Node Roles for Multi-Node Clusters
 
-1. If you have not done so already, create an API token for the Vendor API. See [Using the Vendor API v3](/reference/vendor-api-using#api-token-requirement).
+You can optionally define node roles in the Embedded Cluster Config. For multi-node clusters, roles can be useful for the purpose of assigning specific application workloads to nodes. If nodes roles are defined, users access the Admin Console to assign one or more roles to a node when it is joined to the cluster.
```
paigecalvert, line 189: ^ updated to emphasize optionality

docs/reference/embedded-config.mdx
paigecalvert: Ah also I am just remembering that you can assign more than one role to a node...so maybe there's the use case where you want to label all your controller nodes as such (and then also add labels for the workloads that run on them)

docs/reference/embedded-config.mdx
```diff
@@ -71,17 +71,28 @@ For a full list of versions, see the Embedded Cluster [releases page](https://gi
 
 ## roles
 
-You can define node roles in the Embedded Cluster Config. In multi-node clusters, roles are used to determine which nodes run the Kubernetes control plane, and to assign application workloads to particular nodes. One or more roles can be selected and assigned to a node when it is joined to the cluster. 
+You can optionally customize node roles in the Embedded Cluster Config using the `roles` key.
```
paigecalvert, line 74: **NEW**: Rewrote this section following our convo on 11/15 to change the "defining" roles language. Also to correct info that implies you might "define" more than one controller role

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+* At least three _controller_ nodes that run the Kubernetes control plane are required for HA. This is because clusters use a quorum system, in which more than half the nodes must be up and reachable. In clusters with three controller nodes, the Kubernetes control plane can continue to operate if one node fails because a quorum can still be reached by the remaining two nodes. By default, with Embedded Cluster, all new nodes added to a cluster are controller nodes. For information about customizing the `controller` node role, see [roles](/reference/embedded-config#roles) in _Embedded Cluster Config_.
+
+* Always use an odd number of controller nodes in HA clusters. Using an odd number of controller nodes ensures that the cluster can make decisions efficiently with quorum calculations. Clusters with an odd number of controller nodes also avoid split-brain scenarios where the cluster runs as two, independent groups of nodes, resulting in inconsistencies and conflicts.
+
+* You can have any number of _worker_ nodes in HA clusters. Worker nodes do not run the Kubernetes control plane, but can run workloads such as application or Replicated KOTS workloads.
```
paigecalvert, line 114: **NEW**: Added this last bullet to clarify you can have any number of worker nodes

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+### Best Practices for High Availability
+
+Consider the following best practices and recommendations for creating HA clusters:
+
+* At least three _controller_ nodes that run the Kubernetes control plane are required for HA. This is because clusters use a quorum system, in which more than half the nodes must be up and reachable. In clusters with three controller nodes, the Kubernetes control plane can continue to operate if one node fails because a quorum can still be reached by the remaining two nodes. By default, with Embedded Cluster, all new nodes added to a cluster are controller nodes. For information about customizing the `controller` node role, see [roles](/reference/embedded-config#roles) in _Embedded Cluster Config_.
+
+* Always use an odd number of controller nodes in HA clusters. Using an odd number of controller nodes ensures that the cluster can make decisions efficiently with quorum calculations. Clusters with an odd number of controller nodes also avoid split-brain scenarios where the cluster runs as two, independent groups of nodes, resulting in inconsistencies and conflicts.
```
paigecalvert, line 112: ^ **NEW**: Clarified that it's an odd number of _controller_ nodes that's required. Not just an odd number of nodes in general

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -0,0 +1,135 @@
...
+         
+         1. Go to **Cluster Management > Add node** at the top of the page.
+
+            <img alt="Add node page in the Admin Console" src="/images/admin-console-add-node.png" width="600px"/>
+
+            [View a larger version of this image](/images/admin-console-add-node.png)
+   
+1. Either on the Admin Console **Nodes** screen that is displayed during installation or in the **Add a Node** dialog, select one or more roles for the new node that you will join. Copy the join command.
+
+     Note the following:
```
paigecalvert, line 43: **NEW**: Added some bullet points with the notes relevant to selecting one or more roles when joining a node.

PR #2791, opened 2024-10-29
Add Harbor/Artificatory documentation to Helm Air Gap docs

pandemicsyn:
Add's high level doc's around using Harbor and Artifactory to the Helm Air Gap docs. 

~~https://deploy-preview-2791--replicated-docs-upgrade.netlify.app/vendor/helm-install-airgap#pull-through-registry-proxies-and-mirrors~~

Edit from Paige: Moved this content to a new topic here https://deploy-preview-2791--replicated-docs.netlify.app/vendor/using-third-party-registry-proxy


grantmiller (dismissed):

docs/vendor/helm-install-airgap.mdx
```diff
@@ -101,3 +101,19 @@ After logging into the registry, the customer exports their current version and
 With the list of images the provided `bash` script will automate the process of pulling updated images from the repository, tagging them with a name for an internal registry, and then pushing the newly tagged images to their internal registry. 
 
 Unless the customer has set up the `values` to preserve the updated tag (for example, by using the `latest` tag), they need to edit the `values.yaml` to reference the new image tags. After doing so, they can log in to the OCI registry and perform the commands to install the updated chart.
+
+## Pull through registry proxies and mirrors
+
+The Replicated proxy registry supports additional functionality that can help streamline the air gap installation and update process for customers. For authenticated users, the Replicated proxy registry exposes standard catalog and tag listing endpoints that can be used to generate a list of images required for installation. These endpoints will list images for the last 10 releases.
```
paigecalvert, line 107: >For authenticated users, the Replicated proxy registry **exposes standard catalog and tag listing endpoints** that can be used to generate a list of images required for installation.

Does this refer to something like `'updates.wp-e.10/my-app/1.2.3/images'`? (just basing this guess on what's shown in the helm air gap installation instructions in the download portal)

Asking because I think it would be helpful to connect the dots for people here to explain how these endpoints could them be called after integrating with Harbor or Artifactory

docs/vendor/helm-install-airgap.mdx
```diff
@@ -101,3 +101,19 @@ After logging into the registry, the customer exports their current version and
...
+
+### Integration with Harbor
+
+[Harbor](https://goharbor.io) is a popular open-source container registry that supports both pull through proxy caching and repository mirroring (via pull based replication rules). Customers can use Harbor to cache images on demand and set up replication rules to proactively mirror images.
```
paigecalvert, line 111: Terminology question: My gut feeling is that replication and mirroring are different enough that these terms shouldn't be used interchangeably. (For example, I'm seeing "replicate" used in the Harbor docs, but the term "mirror" does not appear).

Is this assumption true? Or, is there an important reason to use both "replication" and "mirroring"?

docs/vendor/helm-install-airgap.mdx
pandemicsyn: Yep exactly. Its the same concept as our proxy registry, just - yet another level of indirection. Its basically the customer bringing their own proxy registry to the party - that they then configure to pull from the one we supply for the vendor. 

SCR-20241029-onkw-2

docs/vendor/helm-install-airgap.mdx
pandemicsyn: I was using both because both are pretty common - its gonna vary a little by software. Similar to how Harbor has you configure "registries" but Artifactory has you configure "repositories". Replications overloaded here because these also have a replication service concepts - where for example one Harbor instance can replicate its config/registry/contents proactively to other Harbor instances.

Docker themself's use `mirror` (https://docs.docker.com/docker-hub/mirror/).
 
I was thinking of it in terms of what you're doing (mirroring images) by scheduling a replication rule and how you're doing it (harbor replication). 

I'm totally cool rewording this to just say `Customers can use Harbor to cache images on demand and set up replication rules to proactively _copy_ images.` or whatever you think is best.

The only really important thing to keep in some form is the `via pull based replication rules` piece. Because in all these cases we support the pull based actions only.

docs/vendor/helm-install-airgap.mdx
```diff
@@ -101,3 +101,19 @@ After logging into the registry, the customer exports their current version and
...
+
+## Pull through registry proxies and mirrors
+
+The Replicated proxy registry supports additional functionality that can help streamline the air gap installation and update process for customers. For authenticated users, the Replicated proxy registry exposes standard catalog and tag listing endpoints that can be used to generate a list of images required for installation. These endpoints will list images for the last 10 releases.
+
+### Integration with Harbor
+
+[Harbor](https://goharbor.io) is a popular open-source container registry that supports both pull through proxy caching and repository mirroring (via pull based replication rules). Customers can use Harbor to cache images on demand and set up replication rules to proactively mirror images. 
+
+The first step is simply adding the Replicated proxy registry with the customer's credentials (the `license_id`) as a standard Docker Registry ([Harbor - Configure Proxy Cache](https://goharbor.io/docs/2.11.0/administration/configure-proxy-cache/)). That alone will allow the customer to pull images through Harbor on demand and utilize Harbor's caching, security, and other features.
```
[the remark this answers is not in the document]
pandemicsyn: Yep it would proxy.replicated.com or a custom hostname if the vendor's using one. The customer email and license id would go into the access id and access secret fields.

Its this screenshot from https://app.shortcut.com/replicated/story/114300/download-portal-should-allow-enterprises-to-replicate-images-charts-to-popular-registries-artifactory-harbor-etc#activity-114978

Screenshot 2024-10-29 at 16 36 07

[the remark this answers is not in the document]
pandemicsyn: Want me to update to include explicit mentions of the fields ?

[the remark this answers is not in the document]
paigecalvert: Ah nice okay got it! Thank you I can make this update along with my edits, no worries.

docs/vendor/helm-install-airgap.mdx
paigecalvert: Wow what a handy visual! Thanks for that. Makes sense to me.

So, sounds like one main use case for a proxy cache might be: "some enterprise customers might want to bring their own proxy registry (ie Harbor or Artifactory) in order to cache images ahead of installation. This is easier especially for environments with limited/no outbound internet connection because it allows users to cache images in their own proxy registry before installation, and then pull the cached images at the time of installation, saving time during the actual install."

Does that sound about right?

docs/vendor/helm-install-airgap.mdx
paigecalvert: >The only really important thing to keep in some form is the via pull based replication rules piece. Because in all these cases we support the pull based actions only.

Perfect. And I'm not too worried about using both terms as long as it's easily understood by users; just wanted to make sure I was following correctly. Your description makes sense 👍

docs/vendor/helm-install-airgap.mdx
pandemicsyn: Yea, thats pretty good. These proxies may end up being the bridge between the the internet and the air gapped environment. 

The other big thing is probably the security scanning and access controls. Harbor and friends typically support scanning images for security vulnerabilities and help you deal with/inventory all that.

docs/vendor/helm-install-airgap.mdx
```diff
@@ -101,3 +101,44 @@ After logging into the registry, the customer exports their current version and
...
+
+For authenticated users, the [Replicated proxy registry](/vendor/private-images-about) exposes standard catalog and tag listing endpoints that are used during Helm installations in air-gapped environments to generate the list of images required for installation. These endpoints are provided in the Helm air gap installation instructions in the Replicated Download Portal. For example, an `'updates.wp-e.10/my-app/1.2.3/images'` endpoint would return images for version 1.2.3 of an application from the proxy registry. 
+
+Users can optionally connect the Replicated proxy registry with [Harbor](https://goharbor.io) or Artifactory in order to use these endpoints to cache images. Integrating with Harbor also allows users to set up replicated rules for mirroring.
+
+### Limitations
+
+Connecting the Replicated proxy registry with Harbor and Artifactory has the following limitations:
+
+* Neither the catalog or tags listing endpoints respect pagination requests yet but harbor requests 1000 items at a time. (Documented in "Harbor _catalog use": https://docs.google.com/document/d/1Ece7PaFOV12EVhMWFMHkrVZLDHFQfWUv4rWzi_wRjAM/edit#heading=h.wxiwuxpqb3pw)
```
pandemicsyn, line 115: @paigecalvert thats  our internal google doc. The harbor page size isn't documented anymore other than code - so not sure its worth linking too since it could change at any time. 

But https://github.com/goharbor/harbor/blob/9e55afbb9a124dbe47562d56769ad4e9f14cb5ed/src/pkg/registry/client.go#L705 is the permalink for the current harbor version where this is set.

docs/vendor/helm-install-airgap.mdx
```diff
@@ -101,3 +101,44 @@ After logging into the registry, the customer exports their current version and
 With the list of images the provided `bash` script will automate the process of pulling updated images from the repository, tagging them with a name for an internal registry, and then pushing the newly tagged images to their internal registry. 
 
 Unless the customer has set up the `values` to preserve the updated tag (for example, by using the `latest` tag), they need to edit the `values.yaml` to reference the new image tags. After doing so, they can log in to the OCI registry and perform the commands to install the updated chart.
+
+## Using a Pull-Through Proxy  
+
+For authenticated users, the [Replicated proxy registry](/vendor/private-images-about) exposes standard catalog and tag listing endpoints that are used during Helm installations in air-gapped environments to generate the list of images required for installation. These endpoints are provided in the Helm air gap installation instructions in the Replicated Download Portal. For example, an `'updates.wp-e.10/my-app/1.2.3/images'` endpoint would return images for version 1.2.3 of an application from the proxy registry.
```
pandemicsyn, line 107:
```suggestion
For authenticated users, the [Replicated proxy registry](/vendor/private-images-about) exposes standard catalog and tag listing endpoints that are used during Helm installations in air-gapped environments to generate the list of images required for installation. These endpoints are provided in the Helm air gap installation instructions in the Replicated Download Portal. For example, an `'updates.wp-e.10/my-app/1.2.3/images'` endpoint would return images for version 1.2.3 of an application from the proxy registry. 
```

These aren't really used as part of the installation. They're just registry endpoints that some registry client can use to get listing of images available in a registry. I think i might have led you a astray a little. The end result is that the customer can pull an image like `'updates.wp-e.10/my-app/1.2.3/images'` from *their* Harbor or Artifactory instance. 

What these two new endpoints return is different and not even strictly related to Helm. The catalog endpoint returns a list of repositories built from images of the last 10 releases.  The tags listing, lists the tags available in a given repository for those same releases. Those two combined are what Harbor uses to build up a list of images to cache and then serve. What this enables is basically asking proxy.replicated.com "hey let me build up a list of all images i need to know about, im going to download them and save them so i can hand them out myself"

paigecalvert: @pandemicsyn Does the user need to understand the format of those endpoints in order to make the request to the Replicated proxy registry? I think that's the part I'm missing: what are the mechanics of actually pinging these endpoints from the Harbor or Artifactory instance?

paigecalvert: Also, want to make sure I'm understanding these two concepts:
1. "Those two [endpoints] combined are what Harbor uses to build up a list of images to cache and then serve. What this enables is basically asking proxy.replicated.com 'hey let me build up a list of all images i need to know about, im going to download them and save them so i can hand them out myself'"
2. "neither catalog or tags listing are required for Pull through caching by Artifactory or Harbor. Thats purely a piece to enable Harbor's and friends replication feature."

So, if I'm just adding the Replicated proxy registry as an endpoint in Harbor/Artifactory for the purpose of pull-through caching, then the catalog/tag listing endpoints are not used at all?

docs/vendor/using-third-party-registry-proxy.mdx
```diff
@@ -0,0 +1,72 @@
+# Using a Registry Proxy for Helm Air Gap Installations (Alpha)
```
paigecalvert, line 1: Left "for Helm Air Gap Installations" in the title since it sounds like that is the main use case this is intending to server for now. I figure we could always expand and move this topic as the scope of likely use cases grows, but lmk if that makes sense to you

On that note, not sure if "Alpha" is an appropriate label here

paigecalvert (approved):

docs/vendor/using-third-party-registry-proxy.mdx
pandemicsyn: think its fine, our main Helm Air Gap doc flags as being alpha too and this is probably also just the first pass for this bit (hence things like us not respecting pagination requests yet).

PR #2799, opened 2024-11-01
define install types option & include single tenant managed

grantmiller:


docs/vendor/licenses-using-builtin-fields.mdx
```diff
@@ -17,15 +17,19 @@ License fields labeled "KOTS Only" in the table below apply only to Replicated K
...
 | `licenseID`, `licenseId` | ID of the installed license.  This value never changes. |
 | `licenseSequence` | Every time a license is updated, its sequence number is incremented. This value represents the license sequence that the client currently has. |
 | `licenseType` | This field contains a string value that describes the type of the license. This is currently one of `paid`, `trial`, `dev` or `community`. For more information about license types, see [Managing License Type](licenses-about-types).|
 | `signature` | The base64 encoded license signature. This value will change when license is updated. |
+| `isHelmAirgapEnabled` | If a license supports Helm air gap installations, then this field is set to `true` or missing. If Helm air gap is not supported, this field is missing. When you enable this feature, the `license.yaml` file will have license metadata embedded in it and must be re-downloaded. This field requires that the "Install Types" feature is enabled for your Vendor Portal team. Reach out to your Replicated account representative to get access.|
```
paigecalvert, line 32: >then this field is set to `true` or missing

Looks like each of the new fields says that they might be `true` or missing to support the given install type. Does the "missing" option just refer to vendors that don't have Install Types feature enabled on their account?

paigecalvert (approved):

PR #2835, opened 2024-11-19
Update KOTS install overview

paigecalvert:


paigecalvert:
Doing as part of https://github.com/replicatedhq/replicated-docs/pull/2849

PR #2967, opened 2025-01-14
Embedded Cluster single-node and mutli-node architecture diagrams

paigecalvert:
New: https://deploy-preview-2967--replicated-docs.netlify.app/vendor/embedded-overview#architecture-and-add-ons


docs/vendor/embedded-overview.mdx
```diff
@@ -1,15 +1,11 @@
...
 import KurlComparison from "../partials/embedded-cluster/_kurl-comparison.mdx"
 import Requirements from "../partials/embedded-cluster/_requirements.mdx"
-import UpdateOverview from "../partials/embedded-cluster/_update-overview.mdx"
-import SupportBundleIntro from "../partials/support-bundles/_ec-support-bundle-intro.mdx"
-import EmbeddedClusterSupportBundle from "../partials/support-bundles/_generate-bundle-ec.mdx"
-import EcConfig from "../partials/embedded-cluster/_ec-config.mdx"
 import EmbeddedClusterPortRequirements from "../partials/embedded-cluster/_port-reqs.mdx"
 
-# Using Embedded Cluster
+# Embedded Cluster Overview
```
paigecalvert, line 6: Moved the intro material like overview, architecture, reqs, and to a new Overview topic. The "Using Embedded Cluster" was reaching a critical mass (still need to break that page up more, but this seemed like a good start rather than continuing to pile on)

docs/vendor/embedded-overview.mdx
```diff
@@ -19,29 +15,63 @@ If you are instead looking for information about creating Kubernetes Installers
...
 <EmbeddedCluster/>
 
-The following diagram demonstrates how Kubernetes and an application are installed into a customer environment using Embedded Cluster: 
+## Architecture and Add-ons
 
-![Embedded Cluster installs an app in a customer environment](/images/embedded-cluster-install.png)
+This section describes the Embedded Cluster architecture, including the add-ons deployed by Embedded Cluster.
 
-[View a larger version of this image](/images/embedded-cluster-install.png)
+### Single-Node Architecture
```
paigecalvert, line 22: added single- and multi- node diagrams. My thought was that this would make it easier to see at a glance what aspects of EC installs are specific to single versus multi node installs. For example, if we just showed multi-node, then the question might be, "are these components deployed in this way specifically because it's multi-node, or would it look like this for single node too?"

docs/vendor/embedded-overview.mdx
```diff
@@ -19,29 +15,63 @@ If you are instead looking for information about creating Kubernetes Installers
...
+* **(Air Gap Only) Image registry:** For air gap installations in environments with limited or no outbound internet access, Embedded Cluster installs an image registry where the images required to install and run the application are pushed. For more information about installing in air-gapped environments, see [Air Gap Installation with Embedded Cluster](/enterprise/installing-embedded-air-gap).
+
+Any Helm extensions that were included in the [`extensions`](/reference/embedded-config#extensions) field of the Embedded Cluster Config are also installed in the cluster. The namespace or namespaces where extensions are installed are defined by the vendor in the Embedded Cluster Config.  
+
+After the add-ons and Helm extensions are installed, the Embedded Cluster binary deploys a second systemd service on the node named `APP_SLUG-manager` (for example, `gitea-manager`). This manager service orchestrates Embedded Cluster and communicates with the KOTS instance running in the cluster through a websocket.
```
paigecalvert, line 42: ^ I believe you mentioned on the diagramming call that the manager service was deployed after add-ons, but wasn't sure if you were 100% confident on that. 

(Might not also be super important to articulate the install order)

docs/vendor/embedded-overview.mdx
```diff
@@ -19,29 +15,63 @@ If you are instead looking for information about creating Kubernetes Installers
...
 
-### Comparison to kURL
+![Embedded Cluster single-node architecture](/images/embedded-architecture-single-node.png)
+
+[View a larger version of this image](/images/embedded-architecture-single-node.png)
+
+As shown in the diagram above, the Embedded Cluster binary first installs the k0s cluster as a systemd service. This systemd service is named using the slug of the application (for example, `gitea`).
+
+After all the Kubernetes components for the cluster are available, the Embedded Cluster binary then installs the Embedded Cluster add-ons, including:
+
```
paigecalvert, line 33: Listing/explaining the add-ons as part of the diagram description...

docs/vendor/embedded-overview.mdx
```diff
@@ -19,29 +15,63 @@ If you are instead looking for information about creating Kubernetes Installers
...
+:::
+
+![Embedded Cluster multi-node architecture](/images/embedded-architecture-multi-node.png)
+
+[View a larger version of this image](/images/embedded-architecture-multi-node.png)
+
+As shown in the diagram above, in multi-node installations, an instance of the Embedded Cluster manager systemd service runs on each node and communicates with the KOTS instance running on the primary node through a websocket. This allows Embedded Cluster and KOTS to manage installations where workloads are running on multiple nodes in a cluster.
+
+Additionally, for installations that include disaster recovery with Velero, the Velero Node Agent also runs on each node in the cluster. The Node Agent is a Kubernetes DaemonSet that performs backup and restore tasks such as creating snapshots and transferring data during restores.
+
```
paigecalvert, line 59: ^ just called out multi-node specific points/the pieces that are distinct from single-node installs

docs/vendor/embedded-overview.mdx
```diff
@@ -19,29 +15,63 @@ If you are instead looking for information about creating Kubernetes Installers
...
+High availability (HA) for multi-node installations with Embedded Cluster is Alpha and is not enabled by default. For more informaiton about enabling HA, see [Enable High Availability for Multi-Node Clusters (Alpha)](/enterprise/embedded-manage-nodes#ha).
+:::
+
+![Embedded Cluster multi-node architecture](/images/embedded-architecture-multi-node.png)
+
+[View a larger version of this image](/images/embedded-architecture-multi-node.png)
+
+As shown in the diagram above, in multi-node installations, an instance of the Embedded Cluster manager systemd service runs on each node and communicates with the KOTS instance running on the primary node through a websocket. This allows Embedded Cluster and KOTS to manage installations where workloads are running on multiple nodes in a cluster.
+
+Additionally, for installations that include disaster recovery with Velero, the Velero Node Agent also runs on each node in the cluster. The Node Agent is a Kubernetes DaemonSet that performs backup and restore tasks such as creating snapshots and transferring data during restores.
```
ajp-io, line 58: Could consider whether calling out node agent specifically versus Velero generally is worthwile. I think (could be wrong) that Velero runs multiple things. On the first node, we just say Velero rather than listing each piece. So it could be kind of inconsistent to then say that node agent specifically runs on other nodes, but maybe nbd too.

ajp-io: The third node doesn't show anything in gitea (k0s). Was that just to save space?

docs/vendor/embedded-overview.mdx
```diff
@@ -19,29 +15,63 @@ If you are instead looking for information about creating Kubernetes Installers
...
 
-As shown in the diagram above, the Embedded Cluster Config is included in the application release in the Replicated Vendor Portal and is used to generate the Embedded Cluster installation assets. Users can download these installation assets from the Replicated app service (`replicated.app`) on the command line, then run the Embedded Cluster installation command to install Kubernetes and the KOTS Admin Console. Finally, users access the Admin Console to optionally add nodes to the cluster and to configure and install the application.
+The following diagram shows the architecture of a single-node Embedded Cluster installation for an application named Gitea:
 
-### Comparison to kURL
+![Embedded Cluster single-node architecture](/images/embedded-architecture-single-node.png)
+
+[View a larger version of this image](/images/embedded-architecture-single-node.png)
+
+As shown in the diagram above, the Embedded Cluster binary first installs the k0s cluster as a systemd service. This systemd service is named using the slug of the application (for example, `gitea`).
```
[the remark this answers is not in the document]
paigecalvert: Yeah couldn't hurt!

docs/vendor/embedded-overview.mdx
paigecalvert: It was to save space, but I could see how people might assume that means there's nothing on nodes 3+

docs/vendor/embedded-overview.mdx
ajp-io: I would check with an engineer, but I assume the manager comes earlier.

docs/vendor/embedded-overview.mdx
paigecalvert: Cool, I can go with "built-in extensions"

docs/vendor/embedded-overview.mdx
paigecalvert: And good point re: the node agent. I could probably just say "velero" in the 2nd and 3rd nodes and explain that it's a node agent in the description

docs/vendor/embedded-overview.mdx
paigecalvert: from salah:
>manager service gets installed at the end as it’s not involved in the installation process

https://replicated.slack.com/archives/C036BTS7JCE/p1737152140234869?thread_ts=1736451935.270869&cid=C036BTS7JCE

ajp-io (approved):
I'm good with this. I could see an argument for just saying "node" instead of "controller worker node" since you can have controller workers and just workers, depending on the node roles you define. and that seems more in depth than we're trying to show right now in this diagram.

paigecalvert:
@ajp-io Yeah I went back and forth on that...I landed on controller worker for all of them since this particular diagram happens to show a 3-node cluster, and if you are theoretically doing HA (which I know isn't show in this diagram either), having 3 controller nodes is necessary. I thought maybe showing one controller with two workers (or something like that) would be misleading since idk if we want to imply that that's a good pattern to follow. Not sure if that's sound logic, but just to give context :)

ajp-io:
It makes sense. And I think that's fine, unless we want to say that this architecture diagram isn't touching node roles like that, so we just say node to keep it generic.

paigecalvert:
@ajp-io true...you know what, I do think it might be better to just keep it generic since I don't explain what controller workers are in the description. Probably just invites confusion

paigecalvert:
I'll edit that in the diagrams then merge

PR #2968, opened 2025-01-15
Update port requirements for EC

paigecalvert:
https://deploy-preview-2968--replicated-docs.netlify.app/enterprise/installing-embedded-requirements#port-requirements

### DIFF
--- a/docs/partials/embedded-cluster/_port-reqs.mdx
+++ b/docs/partials/embedded-cluster/_port-reqs.mdx
@@ -1,24 +1,111 @@
-Embedded Cluster requires that the following ports are open and available:
+This section lists the ports used by Embedded Cluster. These ports must be open and available for both single- and multi-node installations.
 
-* 2379/TCP &#42;
+#### Ports Used by Local Processes
+
+The following ports must be open and available for use by local processes running on the same node. It is not necessary to create firewall openings for these ports.
+
+* 2379/TCP
+* 9099/TCP
+* 10248/TCP
+* 10257/TCP
+* 10259/TCP
+
+<table>
+  <tr>
+    <th>Port/Protocol</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>2379/TCP</td>
+    <td></td>
+  </tr>
+  <tr>
+    <td>9099/TCP</td>
+    <td></td>
+  </tr>
+  <tr>
+    <td>10248/TCP</td>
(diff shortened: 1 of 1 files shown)


docs/partials/embedded-cluster/_port-reqs.mdx
```diff
@@ -1,24 +1,41 @@
...
+* 10257/TCP
+* 10259/TCP
+
+#### Ports Required for Bidirectional Communication Between Nodes
+
+The following ports are used for bidirectional communication between nodes.
+
+For multi-node installations, create firewall openings between nodes for these ports.
+
+For single-node installations, ensure that there are no other processes using these ports. Although there is no communication between nodes in single-node installations, having these ports open and available ensures that Embedded Cluster preflight checks pass.
```
paigecalvert, line 19: ^ @laverya question for you

I added this description of how, for single-node installs, you don't need to worry about creating firewall openings for these ports since it's not even possible to open a port on a node that doesn't exist...by way of explanation, I also described how the preflight checks would fail in single-node installs if these ports were unavailable 

Is this right? and is this the appropriate level of detail that users would need?

laverya: I would instead say:
```suggestion
For single-node installations, ensure that there are no other processes using these ports. Although there is no communication between nodes in single-node installations, these ports are still required.
```

I am not at all confident that these ports are unneeded/unused on single node installs, merely that they need to be usable between nodes. I expect they are still used within a single node for in-node communication!

docs/partials/embedded-cluster/_port-reqs.mdx
```diff
@@ -1,24 +1,41 @@
...
 * 10256/TCP
-* 10257/TCP &#42;
-* 10259/TCP &#42;
-* 30000/TCP &#42;&#42;&#42;
-* 50000/TCP &#42; &#42;&#42; &#42;&#42;&#42;
 
-&#42; These ports are used only by processes running on the same node. Ensure that there are no other processes using them. It is not necessary to create firewall openings for these ports. 
+#### Admin Console Port
+
+The KOTS Admin Console requires that port 30000/TCP is open and available. Create a firewall opening for port 30000/TCP so that the Admin Console can be accessed by the end user.
```
laverya, line 33: 30000 also needs to be accessible by nodes joining the cluster

PR #2984, opened 2025-01-21
Update Replicated CLI docs for v0.88.1

divolgin:


sidebars.js
```diff
@@ -611,51 +611,72 @@ const sidebars = {
   'reference/replicated-sdk-apis',
   {
     type: 'category',
-    label: 'Replicated CLI',
+    label: 'Replicated CLI', // This label is generated. Do not edit.
     items: [
-      'reference/replicated-cli-installing',
```
paigecalvert, line 616: ^ This topic (how to install the Replicated CLI) got deleted (we need it): https://docs.replicated.com/reference/replicated-cli-installing

If it's easier for the generation, I can rename or relocate this topic so that it doesn't get squashed

sidebars.js
```diff
@@ -611,51 +611,72 @@ const sidebars = {
...
+      // This list is generated. Do not edit.
+      'reference/replicated',
```
paigecalvert, line 617: ^ The description in this new overview topic is not great

>Manage your Commercial Software Distribution Lifecycle using Replicated
>
>## Synopsis
>The 'replicated' CLI allows Replicated customers (vendors) to manage their Commercial Software Distribution Lifecycle (CSDL) using the Replicated API.

We probably don't need much of a definition of the CLI TBH, but something like "the CLI for the Vendor Portal" would prob suffice

docs/reference/replicated-cli-api-get.md
```diff
@@ -0,0 +1,43 @@
+# replicated api get
```
paigecalvert, line 1: General comment: If possible, .mdx might be preferred over .md for Docusaurus (both are file but they switched to mdx with Docusaurus v3)

docs/reference/replicated-cli-app-rm.md
```diff
@@ -0,0 +1,52 @@
+# replicated app rm
```
paigecalvert, line 1: ^ could be helpful to list delete as an alias?

docs/reference/replicated-cli-channel-enable-semantic-versioning.md
```diff
@@ -0,0 +1,38 @@
+# replicated channel enable-semantic-versioning
+
+Enable semantic versioning for CHANNEL_ID
+
+### Synopsis
+
+Enable semantic versioning for the CHANNEL_ID.
```
paigecalvert, line 7: > Enable semantic versioning for the CHANNEL_ID.

I think this would be better as "Enable semantic versioning for the channel."

docs/reference/replicated-cli-app-create.md
```diff
@@ -1,28 +1,52 @@
...
-Create an application.
+### Synopsis
 
-## Usage
-```bash
-replicated app create NAME
+Create a new application in your Replicated account.
+
+This command allows you to initialize a new application that can be distributed
+and managed using the KOTS platform. When you create a new app, it will be set up
```
paigecalvert, line 10: >...application that can be distributed
and managed using the KOTS platform. ...

^ not sure what the KOTS platform means here

docs/reference/replicated-cli-app-delete.md
paigecalvert, line 1: Note: Paige to add redirect for this page to `app rm` in the netlify toml before merging

docs/reference/replicated-cli-channel-enable-semantic-versioning.md
paigecalvert: same comment for disable-semver above

docs/reference/replicated-cli-completion-bash.md
```diff
@@ -1,22 +0,0 @@
-import Help from "../partials/replicated-cli/_help.mdx"
```
paigecalvert, line 1: ^ looks like each of the completion topics got deleted in favor of a single completion topic that lists bash, fish, powershell, zsh as options 

I'm fine with this as long as there wasn't anything that you all think it would be helpful to go into more detail about in individual topics for each option

docs/reference/replicated-cli-vm.md
paigecalvert, line 1: We haven't put anything about the cmx vm feature on the main docs site yet, but maybe we could just comment these `vm` pages out from the docs website sidebar if it would be awkward to try to exclude them from the generated docs

docs/reference/replicated-cli-api-get.md
divolgin: Done

docs/reference/replicated-cli-completion-bash.md
divolgin: Indeed... They will need to be defined as separate subcommands in order to be in separate topics.

docs/reference/replicated-cli-vm.md
divolgin: Any reason for docs site to not match CLI? The only way to remove something here is to make that command hidden, so it won't show up in `replicated --help` either

divolgin:
Will close this PR once all comments are resolved. Changes to address content comments need to be made in the Replicated CLI project.

docs/reference/replicated-cli-app-rm.md
divolgin: This function right now does not add aliases to the output: https://github.com/spf13/cobra/blob/6c3c1163c6df19860515103199e880122ba49e80/doc/md_docs.go#L57-L117

Which is inconsistent with Go template they use for rendering help on command line: https://github.com/spf13/cobra/blob/6c3c1163c6df19860515103199e880122ba49e80/command.go#L558-L559

We can fix that.

docs/reference/replicated-cli-channel-enable-semantic-versioning.md
divolgin: Once doc generating PR is merged, this can be addressed

docs/reference/replicated-cli-app-create.md
divolgin: Once doc generating PR is merged, this can be addressed

docs/reference/replicated-cli-app-delete.md
divolgin: Good catch!

docs/reference/replicated-cli-vm.md
divolgin: Warnings can be added to Cobra docs too, and then they will be in both places

paigecalvert: Just following up that Amber has no problem with the vm commands being in the docs. We don't need to worry about adding a warning

paigecalvert:
Closing. The updated cli docs were added with https://github.com/replicatedhq/replicated-docs/pull/2987

PR #3007, opened 2025-01-27
feat: [sc-108273] GTM Enablement: Improve the docs to make it easier to discover/understand how to get your cluster usage history via API & UI

squizzi:
Story details: https://app.shortcut.com/replicated/story/108273


squizzi:
I put the article under `Vendor Portal / Insights & Telemetry` but I'm open to wherever you two feel we should put it.

https://deploy-preview-3007--replicated-docs-upgrade.netlify.app/vendor/compatibility-matrix-usage

docs/vendor/compatibility-matrix-usage.md
```diff
@@ -0,0 +1,260 @@
...
+used.  This endpoint returns a list of terminated clusters and VMs:
+
+```shell
+curl --request GET \
+     --url 'https://api.replicated.com/vendor/v3/cmx/history' \
+     --header 'Authorization: $REPLICATED_API_TOKEN' \
+     --header 'accept: application/json'
+```
+
+You can specify `total-only=true` to get a quick total count of terminated
```
squizzi, line 163: I should probably move this `total-only` section under the `Unique to this endpoint` text/bullets.

docs/vendor/compatibility-matrix-usage.md
```diff
@@ -0,0 +1,260 @@
...
+- **Duration:** The total time the resource was running, this is the time
+  between the `running` and `terminated` states.
+- **Tag:** Any tags that were applied to the resource.
+
+Each of these fields can be filtered and sorted to help you find the information
+you are looking for.  To sort by a specific field, click on the column header.
+The column will display an up or down arrow to denote that the column is being
+sorted and the direction of the sort:
+
+![Compatibility Matrix History Page, column header clicked](/images/compatibility-matrix-column-sort.png)
```
AmberAlston, line 62: this section of screenshots take up too much page real estate now. @paigecalvert may have a better idea of how to optimize them, but I currently lean to removing them given that they don't provide a ton of value as the filtering is fairly easy to understand conceptually and in the product.

docs/vendor/compatibility-matrix-usage.md
```diff
@@ -0,0 +1,260 @@
+# Compatibility Matrix Usage
+This topic describes using the Replicated Vendor Portal to understand
+Compatibility Matrix usage across your team.
+
+## Currently Active Clusters and Virtual Machines
+You can find current active clusters and virtual machines (VMs) on the Compatibility
+Matrix overview page which is accessible by clicking the **Compatibility Matrix**
+tab and then selecting **Overview** from the sidebar.
```
AmberAlston, line 8: This is only currently true for someone that has the VM feature flag turned on right now. For everyone else, this doesn't say "Overview" it says "Clusters". However, I'm not opposed to this remaining as-is. It will become correct soon, and it's not that hard to figure out the mapping if someone is looking for the overview page.

docs/vendor/compatibility-matrix-usage.md
squizzi: I know @paigecalvert stated using callouts to images, but I wasn't sure what the syntax was on that, that's a potential other way we could do this, or we can remove them, whatever works...

AmberAlston:
> I put the article under `Vendor Portal / Insights & Telemetry` but I'm open to wherever you two feel we should put it.
> 
> https://deploy-preview-3007--replicated-docs-upgrade.netlify.app/vendor/compatibility-matrix-usage

That doesn't feel like the right section because all of that telemetry is currently exclusively about end customer data visibility and insights, and not about team actions taken within VP. 

I think the best section is within the CMX section. We know these docs are a bit rough, and we already need to touch them again to add support for VMs, so for now it probably makes sense to go with a sub-section like "Viewing Compatibility Matrix Usage" under CMX docs, unless @paigecalvert has a better wording suggestion than me :)

docs/vendor/compatibility-matrix-usage.md
paigecalvert: That works for me--we don't mention VMs anywhere currently in the cmx docs but I'm not worried about it if we're going to start documenting that feature soon

docs/vendor/compatibility-matrix-usage.md
paigecalvert: I left in the one showing the result of filtering and removed the others

docs/vendor/compatibility-matrix-usage.md
```diff
@@ -0,0 +1,232 @@
...
+## Filter and Sort Usage History
+
+Each of the fields on the **History** page can be filtered and sorted. To sort by a specific field, click on the column header.
+
+To filter by a specific field, click on the filter icon in the column header, then use each specific filter input to filter the results, as shown below:
+
+![Compatibility Matrix History Page, filter input](/images/compatibility-matrix-column-filter-input.png)
+[View a larger version of this image](/images/compatibility-matrix-column-filter-input.png)
+
+## Get Usage History with the Vendor API v3
```
paigecalvert, line 53: I usually don't document API stuff in the Replicated docs site, as it typically ends up becoming another place to remember to update outside of the API docs themselves (as in, another place that is bound to get out of date 😅 ). Ideally any gaps in information/docs for the API would be edited in the API docs rather than patched in the docs site

My pattern in other topics is to say something like "this topic shows you how to do this in the [Vendor Portal/CLI]. For information about how to do this with the Vendor API v3, see LINK in the Vendor API v3 documentation"

That said, an alternative might be providing a few key examples of using the API if you all think that would be helpful. I'm mainly just wanting to avoid you all having to update the API docs for the usage endpoints in multiple different locations

sidebars.js
```diff
@@ -220,6 +220,7 @@ const sidebars = {
         'vendor/testing-pricing',
         'vendor/testing-supported-clusters',
         'vendor/testing-cluster-addons',
+        'vendor/compatibility-matrix-usage',
```
paigecalvert, line 223: ^ moved up to the cmx section of the docs since this is explaining one the cmx-specific vendor portal pages

docs/vendor/compatibility-matrix-usage.md
```diff
@@ -0,0 +1,232 @@
+# Viewing Compatibility Matrix Usage History
+This topic describes using the Replicated Vendor Portal to understand
+Compatibility Matrix usage across your team.
```
paigecalvert, line 3: ^ I opted to remove the Overview information that was in here previously in favor of keeping the topic focused on usage history specifically (also felt mostly self-explanatory to people using the GUI)

We can put that back here or find another home for it though if you all think it should be documented

docs/vendor/compatibility-matrix-usage.md
squizzi: That's fair, I was trying to follow the key examples, I'll try to make a couple examples that are useful and remove the rest.

docs/vendor/compatibility-matrix-usage.md
squizzi: No, I don't think its necessary, it kind of is self-explanatory 😆

paigecalvert (approved):
This looks good to go to me! Not sure if @AmberAlston , you wanted to take another look. 

Approved on my end

AmberAlston:
It looks good to me as well. Thanks @squizzi

PR #3012, opened 2025-01-29
Update policies-vulnerability-patch.md

ajp-io:
not sure if you have a preference on the order of the products in this list


paigecalvert (dismissed):
Eh I have thought about it but we're not consistent. That works

PR #3046, opened 2025-02-11
Add step for rewriting image names in backup and restore resources

paigecalvert:


docs/vendor/embedded-disaster-recovery.mdx
```diff
@@ -95,7 +95,38 @@ To configure Velero Backup and Restore custom resources for Embedded Cluster dis
        - '*'
      ```
 
-1. (Optional) You can use Velero functionality like [backup](https://velero.io/docs/main/backup-hooks/) and [restore](https://velero.io/docs/main/restore-hooks/) hooks to customize the backup and restore process as needed.
+1. For any image names that you include in your Backup and Restore resources, rewrite the image name using the Replicated KOTS [HasLocalRegistry](/reference/template-functions-config-context#haslocalregistry), [LocalRegistryHost](/reference/template-functions-config-context#localregistryhost), and [LocalRegistryNamespace](/reference/template-functions-config-context#localregistrynamespace) template functions. This ensures that the image name is rendered correctly during deployment, allowing the image to be pulled from the user's local image registry (such as in air gap installations) or through the Replicated proxy registry. If any images referenced in Backup and Restore resources cannot be pulled, the application will fail to deploy after a restore.
```
ajp-io, line 98: I don't know that the last sentence is necessarily true (app will fail to deploy). it means that that container won't run, so it's purpose won't be fulfilled (like if there's a hook, then it won't run). IMO we can leave off that sentence because it'll be self-explanatory that if the image can't be pulled, that part won't work.

ajp-io (commented):
one comment. this also made me realize we need to be sure that the air gap builder picks up images from the backup and restore resources. i asked the team: https://replicated.slack.com/archives/C0499EGNPNY/p1739316078888169

PR #3047, opened 2025-02-12
Remove Alpha and Beta label references on Helm Air Gap 

AmberAlston:


AmberAlston:
Given we recently announced Helm Air Gap as generally available in this month's product email it felt wrong to keep all these alpha and beta labels on the functionality in the docs. Tagging in @grantmiller as a reviewer since this is his product area, want to make sure he's good with these edits before merging.

paigecalvert (approved):

PR #3054, opened 2025-02-17
Release notes for EC 2.1.0

ajp-io:


paigecalvert (approved):
looks good @ajp-io . I split out those kernel preflight checks into their own bullets to make them more visible. good to merge if you are ok with the changes

PR #3066, opened 2025-02-19
Migrate existing installations to HelmChart v2

paigecalvert:


docs/release-notes/rn-app-manager.md
```diff
@@ -79,7 +79,7 @@ Support for Kubernetes: 1.29, 1.30, and 1.31
 
    This flag allows Helm to take ownership of existing resources that were installed without Helm, like resources deployed with HelmChart v1beta1 and `useHelmInstall: false`.
 
-   To migrate an existing installation from `kots.io/v1beta1` with `useHelmInstall: false` to `kots.io/v1beta2`, reach out to your Replicated account representative in Slack or over email.
+   For information about how to migrate an existing installation to KOTS HelmChart `v1beta2`, see [Migrating Existing Installations to HelmChart v2](/vendor/helm-v2-migrate).
```
paigecalvert, line 82: ^ added link to the migration topic

docs/vendor/helm-native-v2-using.md
```diff
@@ -362,8 +362,8 @@ The `kots.io/v1beta2` HelmChart custom resource has the following differences fr
...
 
-### Migrate Existing HelmChart v1 Installations to v2
+### Migrate Existing KOTS Installations to HelmChart v2
 
-After you complete the tasks in this topic to configure the `kots.io/v1beta2` HelmChart custom resource, you can migrate any existing installations that were deployed with `kots.io/v1beta1` with `useHelmInstall: true` to use `kots.io/v1beta2` instead. For more information, see [Migrating Existing Installations to HelmChart v2](helm-v2-migrate).
+Existing KOTS installations can be migrated to use the KOTS HelmChart v2 method, without having to reinstall the application.
 
-To migrate an existing installation from `kots.io/v1beta1` with `useHelmInstall: false` to `kots.io/v1beta2`, reach out to your Replicated account representative in Slack or over email.
\ No newline at end of file
+There are different steps for migrating to HelmChart v2 depending on the application deployment method used previously. For more information, see [Migrating Existing Installations to HelmChart v2](helm-v2-migrate).
```
paigecalvert, line 369: ^ Replaced "reach out to your rep" with a link to the migration steps

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
 
-## How to Migrate Existing Installations
+* The Helm `--take-ownership` flag is supported with KOTS v1.124.0 and later.
 
-The HelmChart custom resource version `kots.io/v1beta2` is available only for installations that use KOTS v1.99.0 or later. To support existing users that have not yet upgraded to KOTS v1.99.0 or later, Replicated recommends that you include both a `kots.io/v1beta2` and a `kots.io/v1beta1` (with `useHelmInstall: true`) HelmChart custom resource for each Helm chart in your release. This allows you to support both installation methods from the same release.
+* The `kots.io/keep` annotation is supported with KOTS v1.122.0 and later.
 
-When you include both versions in a single release for the same Helm chart, installations with KOTS v1.98.0 or earlier use `kots.io/v1beta1`. Installations with KOTS v1.99.0 or later use `kots.io/v1beta2`.
+### Migrate From HelmChart v1 with `useHelmInstall: true`
```
paigecalvert, line 17: ^ we didn't touch on this before since it's pretty straightforward, but it felt helpful to explicitly state that this migration path just involves configuring the HelmChart v2 resource then upgrading

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+1. Configure each HelmChart custom resource in the release to update `apiVersion` to `kots.io/v1beta2`, rewrite images, inject image pull secrets, and add backup labels. See [Configuring the HelmChart Custom Resource v2](helm-native-v2-using).
+
+1. Promote the release to an internal-only channel that your team uses for testing.
+
+1. In your development environment, log in to the Admin Console and confirm that you can upgrade to the new HelmChart v2 release.
+
+1. When you are done testing, promote the release to one or more of your customer-facing channels. Customers can follow the standard upgrade process in the Admin Console to update their instance.
+
+### Migrate From HelmChart v1 with `useHelmInstall: false`
```
paigecalvert, line 33: ^ new steps on how to migrate from useHelmInstall: false. Including info about kots.io/keep and --take-ownership

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+     1. Promote the release to an internal-only channel that your team uses for testing.
+
+1. In a development environment, install the release to test that you can upgrade to the new release successfully.
+
+1. When you are done testing, promote the release to one or more of your customer-facing channels.
+
+    Your customers can follow the standard upgrade process in the Admin Console to update their instance to the new version.
+
+### Migrate From Standard Kubernetes Manifests
```
paigecalvert, line 68: ^ I also added a separate set of steps for migrating to HelmChart v2 when you previously had a release packaged with standard manifests that's not adopted in to one or more Helm charts

I made this separate since it uses the 2-step upgrade process that Alex outlined here: https://github.com/replicated-collab/julia-kots/issues/74#issuecomment-2664108873

It also seems like it could be a fairly common migration path that people could have questions about

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+1. Create a new release containing the Kubernetes manifests for your application.
+
+1. Add the `kots.io/keep` annotation to each of the application manifests in the release. The `kots.io/keep` annotation prevents KOTS from uninstalling resources that were previously installed without Helm when upgrading using the HelmChart v2 method.
+
+1. Save the release.
+
+1. Create another new release:
+
+     1. In the release, add your application Helm chart or charts. Remove the application manifests for resources that were adopted into the Helm chart(s).
```
paigecalvert, line 82: ^ no need to also add the kots.io/keep annotation to the resources that were adopted in to helm chart(s), right?

docs/vendor/helm-v2-migrate.md
paigecalvert: Note: To keep things simpler, I just documented the second method that Alex suggested (the one that the vendor went with). Seemed like doc'ing both potentially migration paths would just add extra potential for confusion, but lmk if there's value in including both

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+     1. Save the release.
+
+1. Test the upgrade process:
+
+    1. Promote the first release to an internal-only channel that your team uses for testing.
+
+         :::note
+         Replicated recommends that you mark the release as required by enabling **Prevent this release from being skipped during upgrades** because customers must upgrade to this release first before they can upgrade to the release that migrates their installation to HelmChart v2.
+         :::
```
paigecalvert, line 107: ^ recommend using required releases

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+  
+     :::note
+     Replicated recommends that you mark the release as required by enabling **Prevent this release from being skipped during upgrades** because customers must upgrade to this release first before they can upgrade to the release that migrates their installation to HelmChart v2.
+     :::
+
+1. Promote the second release containing your application Helm chart to the same channels.
+
+1. Instruct customers to migrate by first upgrading to the release containing the standard manifests, then upgrading to the release packaged with Helm second.
+
+## Support Customers on KOTS Versions Earlier Than v1.99.0 {#support-both-v1-v2}
```
paigecalvert, line 125: pre-existing content about what to do when you have customers that aren't on 1.99 or later

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+The HelmChart v2 installation method requires KOTS v1.99.0 or later. If you have existing customers that have not yet upgraded to KOTS v1.99.0 or later, Replicated recommends that you support both the HelmChart v2 and v1 installation methods from the same release until all installations are running KOTS v1.99.0 or later.
+
+To support both installation methods from the same release, include both versions of the HelmChart custom resource for each Helm chart in your application releases (HelmChart `kots.io/v1beta2` and HelmChart `kots.io/v1beta1` with `useHelmInstall: true`).
+
+When you include both versions of the HelmChart custom resource for a Helm chart, installations with KOTS v1.98.0 or earlier use the v1 method, while installations with KOTS v1.99.0 or later use v2.
+
+After all customers are using KOTS v1.99.0 or later, you can remove the HelmChart v1 custom resources so that all customers are using the HelmChart v2 method.
+
+## HelmChart v2 Migration FAQs
```
paigecalvert, line 135: ^ not sure if we really need these faqs, but with all the subtle differences I thought it might be helpful. I was inspired by Gerard's slack message that cleared up our uncertainty: https://replicated.slack.com/archives/C68QDSQ64/p1739917976982519?thread_ts=1739898768.942019&cid=C68QDSQ64

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+Because the HelmChart v1 with `useHelmInstall: false` method does not deploy resources with Helm, it is necessary to use the `kots.io/keep` annotation and the Helm `--take-ownership` flag when migrating to the HelmChart v2 installation method. These ensure that Helm can take ownership of existing resources and that the resources are not uninstalled during upgrade.
+
+For more information about how KOTS deploys Helm charts, including information about the deprecated HelmChart v1 installation methods, see [About Distributing Helm Charts with KOTS](helm-native-about).
+
+### Where can I find additional help migrating existing installations to HelmChart v2?
+
+For additional support migrating existing installations to HelmChart v2, Replicated recommends that you post your question in the [Replicated Community](https://community.replicated.com/) site. You can also reach out to your Replicated account representative.
```
paigecalvert, line 169: ^ post on community or reach out to your rep if you need help

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+         apiVersion: kots.io/v1beta2
+         kind: HelmChart
+         metadata:
+           name: samplechart
+         spec:
+           helmUpgradeFlags:
+             - --take-ownership
+         ```   
+
+         When the `--take-ownership` flag is enabled, Helm automatically takes ownership of resources that were previously deployed without Helm during upgrade. The Helm `--take-ownership` flag requires KOTS 1.122.0 or later.
```
nvanthao, line 56: nit: requires KOTS 1.124.0 or later

paigecalvert: good catch!

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
 # Migrating Existing Installations to HelmChart v2
 
-This topic provides guidance for migrating existing installations that were deployed with the HelmChart custom resource `kots.io/v1beta1` with `useHelmInstall: true` to use `kots.io/v1beta2` instead.
+This topic describes how to migrate existing Replicated KOTS installations to the KOTS HelmChart `kots.io/v1beta2` (HelmChart v2) installation method, without having to reinstall the application. It also includes information about how to support both HelmChart v1 and HelmChart v2 installations from a single release, and lists frequently-asked questions (FAQs) related to migrating to HelmChart v2.
 
-## Overview
+For additional support migrating existing installations to HelmChart v2, reach out to your Replicated account representative.
```
ajp-io, line 5: do we feel like we still need this? on the one hand, i definitely agree with it, on the other hand, the details are here and we would always hope people reach out if they run into issues.

ajp-io: maybe up to chris if he feels that's necessary still

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+### Migrate From HelmChart v1 with `useHelmInstall: false`
+
+This section describes how to migrate existing HelmChart v1 installations with `useHelmInstall: false`. These migration steps ensure that KOTS does not uninstall any resources that were previously deployed without Helm and that Helm takes ownership of these existing resources.
+
+To migrate existing installations from HelmChart v1 and `useHelmInstall: false` to HelmChart v2:
+
+1. Create a new release containing your application files:
+
+     1. Add the `kots.io/keep` annotation to any resources that were previously installed with `kubectl apply`, including resources defined in Kubernetes manifests or in your Helm `templates`. The `kots.io/keep` annotation prevents KOTS from uninstalling resources that were previously installed without Helm when upgrading using the HelmChart v2 method.
```
ajp-io, line 41: Customers might not realize that v1 false uses `kubectl apply`. from their pov, they think it's a helm chart. so i wonder if we should make that clearer here. i might read this and say, well i had a helm chart, so no resources were applied with kubectl. but in reality, everything in the chart was kubectl applied.

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+     1. In the `helmUpgradeFlags` field of the Embedded Cluster Config, add the `--take-ownership` flag, as shown below:
```
ajp-io, line 43: not EC Config, helm chart v2

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+         When the `--take-ownership` flag is enabled, Helm automatically takes ownership of resources that were previously deployed without Helm during upgrade. The Helm `--take-ownership` flag requires KOTS 1.122.0 or later.
+
+     1. Configure each HelmChart custom resource in the release to update `apiVersion` to `kots.io/v1beta2`, rewrite images, inject image pull secrets, and add backup labels. See [Configuring the HelmChart Custom Resource v2](helm-native-v2-using).
+
+     1. Promote the release to an internal-only channel that your team uses for testing.
+
+1. In a development environment, install the release to test that you can upgrade to the new release successfully.
+
+1. When you are done testing, promote the release to one or more of your customer-facing channels.
```
ajp-io, line 64: would you want to remove the keep annotation in a subsequent release though? I'm not sure if it matters or if @nvanthao tested it.

docs/vendor/helm-v2-migrate.md
ajp-io: nope

docs/vendor/helm-v2-migrate.md
ajp-io: i don't think the second one needs to be required, unless they're going to remove the `--take-ownership` flag in a later release. it would need to be required if only this version has `--take-ownership` applied. which could be a good idea.

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,169 @@
...
+
+After all customers are using KOTS v1.99.0 or later, you can remove the HelmChart v1 custom resources so that all customers are using the HelmChart v2 method.
+
+## HelmChart v2 Migration FAQs
+
+This section includes FAQs related to migrating existing installations to the KOTS HelmChart v2 method.
+
+### Which migration scenarios require the `kots.io/keep` annotation?
+
+When applied to a resource in a release, the `kots.io/keep` annotation prevents the given resource from being uninstalled. The `kots.io/keep` annotation can be used to prevent KOTS from deleting resources that were either adopted into Helm charts, or that were previously deployed without Helm.
```
ajp-io, line 141: @nvanthao does this annotation work if you put it on a resource in a Helm chart, or does it only work for manifests in a release?

docs/vendor/helm-v2-migrate.md
chris-sanders: I'm fine dropping this since technically you can always reach out if you have further questions. I'm fine helping people as well I'd say do whatever is commensurate with the rest of our documentation. If we tend to reinforce you can ask for help this is fine if we don't it's fine to drop it as well.

docs/vendor/helm-v2-migrate.md
paigecalvert: oops good catch

docs/vendor/helm-v2-migrate.md
paigecalvert: The workflow I was trying to capture was testing the first (required) release, then promoting that same release to a customer-facing channel and again marking it as required.

But that said, the _second_ release does have the `--take-ownership` flag...so it sounds like maybe it could be helpful to actually mark both the first and second releases are required?

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,198 @@
...
+    
+       ```yaml
+       apiVersion: apps/v1
+       kind: Statefulset
+       metadata:
+         name: postgresql
+         # Add the kots.io/keep annotation
+         annotations:
+           kots.io/keep: "true"
+       ```
```
paigecalvert, line 51: ^ added these examples of what the keep annotation looks like

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,198 @@
...
+   
+   1. In your development environment, access the Admin Console to upgrade to the second release.
+
+1. When you are done testing, promote the first release to one or more of your customer-facing channels. Replicated recommends that you mark the release as required by enabling **Prevent this release from being skipped during upgrades**. For more information about required releases, see [Properties](/vendor/releases-about#properties) in _About Channels and Releases_.
+
+1. Promote the second release to the same customer-facing channel(s). Replicated recommends that you mark the release as required by enabling **Prevent this release from being skipped during upgrades**.
+
+1. Instruct customers to migrate by first upgrading to the release where the `kots.io.keep` annotation is applied to your resources, then upgrading to the release with HelmChart v2.
+
+1. In subsequent releases, remove the `helmUpgradeFlags` field (including the `--take-ownership` flag) from the HelmChart custom resource. This flag is only required during one upgrade to allow Helm to take ownership of your application resources.
```
paigecalvert, line 92: ^ added step to remove the flag for later releases since it's not needed

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,198 @@
...
+
+   1. In a development environment, install the first release.
+
+   1. Promote the second release to the same channel.
+   
+   1. In your development environment, access the Admin Console to upgrade to the second release.
+
+1. When you are done testing, promote the first release to one or more of your customer-facing channels. Replicated recommends that you mark the release as required by enabling **Prevent this release from being skipped during upgrades**. For more information about required releases, see [Properties](/vendor/releases-about#properties) in _About Channels and Releases_.
+
+1. Promote the second release to the same customer-facing channel(s). Replicated recommends that you mark the release as required by enabling **Prevent this release from being skipped during upgrades**.
```
paigecalvert, line 88: ^ recommend marking both releases as required (which is important especially if we want them to remove the helm upgrade take ownership flag in subsequent releases)

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,198 @@
...
+1. For each Helm chart in the release, configure the corresponding HelmChart custom resource to update `apiVersion` to `kots.io/v1beta2`, rewrite images, inject image pull secrets, and add backup labels. See [Configuring the HelmChart Custom Resource v2](helm-native-v2-using).
+
+1. Promote the release to an internal-only channel that your team uses for testing.
+
+1. In your development environment, log in to the Admin Console and confirm that you can upgrade to the new HelmChart v2 release.
+
+1. When you are done testing, promote the release to one or more of your customer-facing channels. Customers can follow the standard upgrade process in the Admin Console to update their instance.
+
+### Migrate From HelmChart v1 with `useHelmInstall: false`
+
```
paigecalvert, line 32: Note: updated these steps to use the same "two step" upgrade that is used when migrating from manifests to helmchart v2 (kots keep then helm take ownership)

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,198 @@
...
+1. For each Helm chart in the release, configure the corresponding HelmChart custom resource to update `apiVersion` to `kots.io/v1beta2`, rewrite images, inject image pull secrets, and add backup labels. See [Configuring the HelmChart Custom Resource v2](helm-native-v2-using).
+
+1. Promote the release to an internal-only channel that your team uses for testing.
+
+1. In your development environment, log in to the Admin Console and confirm that you can upgrade to the new HelmChart v2 release.
+
+1. When you are done testing, promote the release to one or more of your customer-facing channels. Customers can follow the standard upgrade process in the Admin Console to update their instance.
+
+### Migrate From HelmChart v1 with `useHelmInstall: false`
```
ajp-io, line 31: I wonder if it's worth mentioning that false is the default value, because probably most people using this method won't have `useHelminstall` specified at all in their HelmChart resource.

docs/vendor/helm-v2-migrate.md
ajp-io: I'd just say to remove the take ownership flag, since they could be specifying other helm upgrade flags that they want to keep

ajp-io: If I'm thinking of this correctly, they should also remove the `keep` annotation from the templates. Since this is the same chart, the annotation would be there, and we don't want that in the future (though I don't think KOTS will actually do anything about it anyway since it's in a Helm chart). But for cleanup purposes, that might be best.

docs/vendor/helm-v2-migrate.md
```diff
@@ -1,19 +1,198 @@
...
-When you include both versions in a single release for the same Helm chart, installations with KOTS v1.98.0 or earlier use `kots.io/v1beta1`. Installations with KOTS v1.99.0 or later use `kots.io/v1beta2`.
+To migrate existing installations from HelmChart v1 with `useHelmInstall: true` to HelmChart v2:
 
-After all customer environments where the Helm chart was previously installed with `kots.io/v1beta1` and `useHelmInstall: true` are upgraded to KOTS v1.99.0 or later, you can remove version `kots.io/v1beta1` of the HelmChart custom resource from your releases.
\ No newline at end of file
+1. In a development environment, install an application release using the KOTS HelmChart v1 with `useHelmInstall: true` method. You will use this installation to test the migration to HelmChart v2.
+
+1. Create a new release containing your application files.
+
+1. For each Helm chart in the release, configure the corresponding HelmChart custom resource to update `apiVersion` to `kots.io/v1beta2`, rewrite images, inject image pull secrets, and add backup labels. See [Configuring the HelmChart Custom Resource v2](helm-native-v2-using).
```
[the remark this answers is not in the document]
paigecalvert: I like it 👍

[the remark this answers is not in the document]
ajp-io: I actually thought i deleted this haha, so nbd either way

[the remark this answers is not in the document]
ajp-io: Note that it's in a few spaces if you do change it though

docs/vendor/helm-v2-migrate.md
paigecalvert: Oh that's good to know. Yeah I'll add that in

docs/vendor/helm-v2-migrate.md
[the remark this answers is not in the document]
paigecalvert: Ha I updated it in both places 😄 (the third place was slightly different because it was instructing them to create the resource when migrating from standard manifests to helm charts)

PR #3086, opened 2025-03-03
perms set for data dir

ajp-io:


paigecalvert (approved):
@ajp-io I swapped these around (so that it's: we set the data dir permissions then we verify the permissions). Removed the info about if the install fails or not

PR #3136, opened 2025-03-26
HOLD FOR RELEASE: Updates for Embedded Cluster HA

paigecalvert:
https://deploy-preview-3136--replicated-docs.netlify.app/enterprise/embedded-manage-nodes#ha

~~Docs edits to the description of the `roles` key for clarity (no changes to the details of how that feature actually works): https://deploy-preview-3136--replicated-docs.netlify.app/reference/embedded-config#roles~~


docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -83,13 +75,11 @@ To add nodes to a cluster:
 
 1. Repeat these steps for each node you want to add.    
 
-## Enable High Availability for Multi-Node Clusters (Alpha) {#ha}
+## Enable High Availability for Multi-Node Clusters{#ha}
 
-Multi-node clusters are not highly available by default. The first node of the cluster is special and holds important data for Kubernetes and KOTS, such that the loss of this node would be catastrophic for the cluster. Enabling high availability (HA) requires that at least three controller nodes are present in the cluster. Users can enable HA when joining the third node.
```
paigecalvert, line 88: Replaced _"Users can enable HA when joining the third node."_ with:
>Users are automatically prompted to enable HA when joining the third controller node to a cluster. Alternatively, users can enable HA after adding three or more controller nodes with the `enable-ha` command.

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -101,47 +91,56 @@ For more information about the Embedded Cluster built-in extensions, see [Built-
 
 Enabling high availability has the following requirements:
 
-* High availability is supported with Embedded Cluster 1.4.1 or later.
+* High availability is supported with Embedded Cluster **VERSION** and later.
```
paigecalvert, line 94: ^ probably should change this to the GA version

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -101,47 +91,56 @@ For more information about the Embedded Cluster built-in extensions, see [Built-
...
 
 * Always use an odd number of controller nodes in HA clusters. Using an odd number of controller nodes ensures that the cluster can make decisions efficiently with quorum calculations. Clusters with an odd number of controller nodes also avoid split-brain scenarios where the cluster runs as two, independent groups of nodes, resulting in inconsistencies and conflicts.
 
 * You can have any number of _worker_ nodes in HA clusters. Worker nodes do not run the Kubernetes control plane, but can run workloads such as application or Replicated KOTS workloads.
 
-### Create a Multi-Node HA Cluster
+### Create a Multi-Node Cluster with High Availability {#create-ha}
+
+You can enable high availability for a multi-node cluster when joining the third controller node. Alternatively, you can enable HA for an existing cluster with three or more controller nodes. For more information, see [Enable High Availability For an Existing Cluster](#enable-ha-existing) below.
```
paigecalvert, line 114: ^ Split the options into two different headings: "Create a Multi-Node Cluster with High Availability" and "Enable High Availability for an Existing Cluster".

I also tried combining them in the same procedure but this felt easier to skim from looking at the TOC for the page

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -101,47 +91,56 @@ For more information about the Embedded Cluster built-in extensions, see [Built-
...
+
+### Enable High Availability For an Existing Cluster {#enable-ha-existing}
+
+To enable high availability for an existing Embedded Cluster installation with three or more controller nodes, run the following command:
+
+```bash
+sudo ./APP_SLUG enable-ha
+```
+
+Where `APP_SLUG` is the unique slug for the application.
```
paigecalvert, line 146: TBD on options we should specify for this command

docs/reference/embedded-config.mdx
```diff
@@ -68,83 +68,30 @@ You must specify which version of Embedded Cluster to install. Each version of E
...
-apiVersion: embeddedcluster.replicated.com/v1beta1
-kind: Config
-spec:
-  roles:
-    controller:
-      name: management
-      labels:
-        management: "true" # Label applied to "management" nodes
-```      
+## nodeRoles
```
paigecalvert, line 71: rewrote this section and renamed to nodeRoles

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -101,47 +91,56 @@ For more information about the Embedded Cluster built-in extensions, see [Built-
...
-
-Enabling high availability has the following limitations:
-
-* High availability for Embedded Cluster in an Alpha feature. This feature is subject to change, including breaking changes. For more information about this feature, reach out to Alex Parker at [alexp@replicated.com](mailto:alexp@replicated.com).
+### Limitation
 
-* The `--enable-ha` flag serves as a feature flag during the Alpha phase. In the future, the prompt about migrating to high availability will display automatically if the cluster is not yet HA and you are adding the third or more controller node. 
-
-* HA multi-node clusters use rqlite to store support bundles up to 100 MB in size. Bundles over 100 MB can cause rqlite to crash and restart.
+HA multi-node clusters use rqlite to store support bundles up to 100 MB in size. Bundles over 100 MB can cause rqlite to crash and restart.
```
ajp-io, line 100: I don't think this is specific to HA. I think this is true across the board.

docs/enterprise/embedded-manage-nodes.mdx
ajp-io: doesn't look like it'll require options

ajp-io (commented):
not final, but it looks like we might stick with our current node roles set up in the EC Config and continue letting people multi-select. I'll get back to you with a final word. I'm not confident enough that the proposal is making things better enough to warrant changing it.

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -83,13 +75,11 @@ To add nodes to a cluster:
 
 1. Repeat these steps for each node you want to add.    
 
-## Enable High Availability for Multi-Node Clusters (Alpha) {#ha}
+## Enable High Availability for Multi-Node Clusters{#ha}
 
-Multi-node clusters are not highly available by default. The first node of the cluster is special and holds important data for Kubernetes and KOTS, such that the loss of this node would be catastrophic for the cluster. Enabling high availability (HA) requires that at least three controller nodes are present in the cluster. Users can enable HA when joining the third node.
+Multi-node clusters are not highly available by default. The first node of the cluster is special and holds important data for Kubernetes and KOTS, such that the loss of this node would be catastrophic for the cluster. Enabling high availability requires that at least three controller nodes are present in the cluster.
```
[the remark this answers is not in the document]
paigecalvert: You are no longer special, first node.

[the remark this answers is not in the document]
ajp-io: lol exactly

docs/enterprise/embedded-manage-nodes.mdx
paigecalvert: Oh good catch. I can move to main limitations and get eng confirmation

ajp-io:
plan is to stick with the current roles setup, as well as the multi-select of roles per node

docs/reference/embedded-config.mdx
```diff
@@ -72,61 +75,12 @@ For a full list of versions, see the Embedded Cluster [releases page](https://gi
 
 You can optionally customize node roles in the Embedded Cluster Config using the `roles` key.
```
paigecalvert, line 76: I reverted this back to `roles` (from `nodeRoles`)

As part of reverting back to the original description of how this feature works, I also did a bit of a rewrite/reorg to simplify things. This meant reducing word count, using one full example for setting roles.controller and roles.custom with names and labels rather than a few scattered examples, and changing the subheadings so that "labels" isn't it's own h3 anymore

docs/enterprise/embedded-manage-nodes.mdx
```diff
@@ -125,23 +105,38 @@ Consider the following best practices and recommendations for creating HA cluste
...
+     **Example:**
 
      ```bash
-     sudo ./APP_SLUG join --enable-ha 10.128.0.80:30000 tI13KUWITdIerfdMcWTA4Hpf
+     sudo ./APP_SLUG join 10.128.0.80:30000 tI13KUWITdIerfdMcWTA4Hpf
      ```
+     Where `APP_SLUG` is the unique slug for the application.
 
-1. After the third node joins the cluster, type `y` in response to the prompt asking if you want to enable high availability.
+1. In response to the prompt asking if you want to enable high availability, type `y` or `yes`:
```
ajp-io, line 127: the screenshot here is different, but we can make it a follow up to update it if you'd like.

docs/reference/embedded-config.mdx
```diff
@@ -135,9 +89,11 @@ kind: Config
 spec:
   roles:
     controller:
+      # Optionally change the name for the default controller role
       name: management
```
ajp-io, line 93: thinking about it more, i don't like management as a name because that implies it's just the controller. but people's workloads run here too.

like in these examples, your db runs on dedicated db nodes, and everything else runs on controllers. so it's not really just a management node then.

i think app could be a good example name here.

docs/reference/embedded-config.mdx
```diff
@@ -147,6 +103,22 @@ spec:
         gpu: "true" # Label applied to "gpu" nodes    
 ```
 
+### roles.controller
+
+In the `roles.controller` key, you can set the following fields to customize the default controller role:
+* `name`: Set the name that is assigned to controller nodes. By default, controller nodes are named “controller”.
+    :::note
+    If you plan to create any custom roles, Replicated recommends that you change the default name for the controller role to a term that is easy to understand, such as "management". This is because, when you add custom roles, both the name of the controller role and the names of any custom roles are displayed to the user when they join a node.
```
ajp-io, line 111: i don't think i'd offer a recommendation like management. app is what i suggested for our examples above, and we could say that here if you think the guidance is helpful. but I don't know what people will run on controllers, so it's hard to give a good name.

paigecalvert: We could get rid of that note too. It was a carry-over

ajp-io: the note might be worthwhile so people know that controller will be the role name otherwise. i guess they'll see that when they test anyway though. so we can remove if you think it's easier/clearer.

paigecalvert: We'd still have "By default, controller nodes are named 'controller' in the `name` description, so that might cover it

docs/reference/embedded-config.mdx
paigecalvert: Sounds good, I'll use app

PR #3162, opened 2025-04-04
Document ability to configure workerProfiles

paigecalvert:
https://deploy-preview-3162--replicated-docs.netlify.app/reference/embedded-config#configure-kubelet-with-k0s-worker-profiles

### DIFF
--- a/docs/reference/embedded-config.mdx
+++ b/docs/reference/embedded-config.mdx
@@ -285,15 +285,20 @@ spec:
         key: value
 ```
 
-### Configure Kubelet with k0s Worker Profiles
+### Configure the Kubelet with k0s Worker Profiles
 
-k0s supports worker profiles, which allow you to configure Kubelet with different settings for different worker nodes. One common use case for setting a worker profile is that you need more pods on a single node than the default limit of 100. In this case, you could define a worker profile to set `maxPods` to 150. Another common example is reducing startup time by setting `maxParallelImagePulls` to increase the maximum number of image pulls that can be done in parallel.
+k0s supports worker profiles, which allow you to configure the Kubelet with different settings for different worker nodes. One common use case for setting a worker profile is that you need more pods on a single node than the default limit of 100. In this case, you could define a worker profile to set `maxPods` to 150. Another common example is reducing startup time by setting `maxParallelImagePulls` to increase the maximum number of image pulls that can be done in parallel.
 
 When a worker profile is defined in the Embedded Cluster Config, Embedded Cluster uses the profile for every node in the cluster during initial installation and when joining nodes to the cluster.
 
 You can add a worker profile in the Embedded Cluster Config under `unsupportedOverrides.k0s.config.spec.workerProfiles[]`. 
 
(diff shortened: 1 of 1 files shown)


docs/reference/embedded-config.mdx
```diff
@@ -284,3 +284,35 @@ spec:
...
+  unsupportedOverrides:
+    k0s: |
+      config:
+        spec:
+          # Define a default profile that sets maxPods to 150
+          workerProfiles:
+          - name: default
+            values:
+              maxPods: 150
+```
```
paigecalvert, line 313: based on https://github.com/replicatedhq/embedded-cluster/pull/1974/files#diff-8246356184ddd84ea018be5d8d52aa4a7526f4106cd06ff20325d0ae78a6c574L40

docs/reference/embedded-config.mdx
```diff
@@ -284,3 +284,35 @@ spec:
       values: |
         key: value
 ```
+
+### Configure Kubelet with k0s Worker Profiles
```
ajp-io, line 288: seems like the k8s docs tend to say "the kubelet" https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

docs/reference/embedded-config.mdx
```diff
@@ -284,3 +284,35 @@ spec:
...
+apiVersion: embeddedcluster.replicated.com/v1beta1
+kind: Config
+spec:
+  unsupportedOverrides:
+    k0s: |
+      config:
+        spec:
+          # Define a default profile that sets maxPods to 150
+          workerProfiles:
+          - name: default
```
ajp-io, line 310: i think the name "default" is actually reserved by k0s and you explicitly should not name it this. confirm with andrew or ash

hedge-sparrow: Yeh, we should avoid using the `default` name as it's used by k0s.

paigecalvert: Used custom-maxpods instead based on examples I saw in the k0s docs for worker profiles

docs/reference/embedded-config.mdx
```diff
@@ -284,3 +284,35 @@ spec:
       values: |
         key: value
 ```
+
+### Configure Kubelet with k0s Worker Profiles
+
+k0s supports worker profiles, which allow you to configure Kubelet with different settings for different worker nodes. One common use case for setting a worker profile is that you need more pods on a single node than the default limit of 100. In this case, you could define a worker profile to set `maxPods` to 150. Another common example is reducing startup time by setting `maxParallelImagePulls` to increase the maximum number of image pulls that can be done in parallel.
+
+When a worker profile is defined in the Embedded Cluster Config, Embedded Cluster uses the profile for every node in the cluster during initial installation and when joining nodes to the cluster.
```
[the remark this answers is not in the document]
paigecalvert: I did mention this over in the Limitations, but I could see how those would be easy to miss. I moved limitations up above the example so they are more apparent

PR #3167, opened 2025-04-07
add note on worker profiles

hedge-sparrow:


ajp-io:
Paige is actually working on adding docs for this. Sorry that you didn't realize that. https://github.com/replicatedhq/replicated-docs/pull/3162

There was one outstanding question for you though. Her example used "default" as the name for the profile, but we need to document that you should _not_ use that name, right?

PR #3188, opened 2025-04-14
Fixing a dead link in the SDK values page

jtuchscherer:
Fixing a dead link to https://github.com/replicatedhq/replicated-sdk/blob/main/chart/values.yaml


paigecalvert:
thank you @jtuchscherer !

PR #3194, opened 2025-04-18
add migration doc

ajp-io:
the wording here is probably bad, but i didn't know what to say and didn't want to spend too much tine thinking about it. please change it.


paigecalvert (dismissed):
@ajp-io made some updates. I also linked directly to the migration google doc versus the slide deck, but we could also include a link to the slide deck with another sentence like "For more information about the benefits of Embedded Cluster compared to kURL, see DECK" (and that could even go in the Comparison to kURL section versus this limitation)

paigecalvert (approved):

PR #3207, opened 2025-04-25
chore: Remove need to have wget in install instructions

banjoh:
In the command to install replicated CLI, `curl` should be sufficient to download the binary. There is no need for having users have both in their environments


paigecalvert (approved):
thanks @banjoh :)

PR #3217, opened 2025-04-30
Update VM SSH-ENDPOINT to show we can access by id or name

marccampbell:
Updating the `ssh-endpoint` subcommand with the fact that you can use id or name in the cli


paigecalvert:
Closing. These docs will be automatically generated when the new version of the cli is released. see https://replicated.slack.com/archives/C02BPJQQP28/p1746032626315779?thread_ts=1746031693.038779&cid=C02BPJQQP28

PR #3272, opened 2025-05-27
Reformat "open in chatgpt" dropdown

paigecalvert:


src/components/HomepageFeatures.js
paigecalvert, line 1: these are legacy components from the default docusaurus site. we're not using them and they can be removed

src/theme/DocItem/Layout/index.js
paigecalvert, line 1: we're now swizzling DocItem/Content instead of Layout

PR #3278, opened 2025-05-29
feat(ec): new PrivateCACert template function behavior

emosbaugh:


docs/reference/template-functions-static-context.md
```diff
@@ -16,7 +16,7 @@ This topic provides a list of the KOTS template functions in the Static context.
 func PrivateCACert() string
 ```
 
-PrivateCACert returns the name of a ConfigMap that contains private CA certificates provided by the end user. For Embedded Cluster installations, these certificates are provided with the `--private-ca` flag for the `install` command. For KOTS installations, the user provides the ConfigMap using the `--private-ca-configmap` flag for the `install` command.
+For KOTS installations, PrivateCACert returns the name of a ConfigMap that contains CA certificates provided by the end user using the `--private-ca-configmap` flag for the `install` command. For Embedded Cluster installations, the ConfigMap contains a single file which is the CA trust store from the host.
```
diamonwiggins, line 19: Should we mention how we determine the CA trust store from the host here? Should be pretty straightforward, and wonder if it'll be useful in the future during support.

emosbaugh: added

paigecalvert (approved):

PR #3288, opened 2025-06-04
Revert "ETCD port 2379 must be open between nodes"

sgalsaleh:
Reverts replicatedhq/replicated-docs#3287

This is not true for k0s.


diamonwiggins (approved):
lgtm

PR #3349, opened 2025-06-30
K8s 1.30 out of support now

AmberAlston:


AmberAlston:
Note that the next version to support, K8s 1.34, looks to be coming out around Aug 27th per https://www.kubernetes.dev/resources/release/#:~:text=Wednesday%2027th%20August%202025:%20Week,34.0%20released

paigecalvert (approved):

PR #3365, opened 2025-07-08
Removing an old alpha reference on EC doc

AmberAlston:
Still said alpha in this section of the docs, but has been GA since 2.4.0 https://docs.replicated.com/release-notes/rn-embedded-cluster#240


paigecalvert:
thanks @AmberAlston !

PR #3400, opened 2025-07-24
Fix Linux install commands

siggy:
The Linux install commands relied on grepping for
`browser_download_url` in the latest release JSON. Unfortunately there are more than one occurrence of that string, and it produces an incorrect download URL.

Modify the Linux install commands to instead grep for the latest version, and then explicitly generate the download URL based on that.


AmberAlston (approved):
Thank you @siggy !

PR #3439, opened 2025-08-11
Quick start updates

paigecalvert:
https://deploy-preview-3439--replicated-docs-upgrade.netlify.app/vendor/quick-start


docs/vendor/quick-start.mdx
```diff
@@ -199,50 +269,114 @@ Before you begin, ensure that you have access to a VM that meets the requirement
...
+             ![DNS record for a VM](/images/compatibility-matrix-dns-record.png)
+             [View a larger version of this image](/images/compatibility-matrix-dns-record.png)
+
+         1. (Optional) Copy the URL for the DNS record with a target port of 443 and save it somewhere on your local machine. You will be asked to provide this hostname in the Admin Console as part of the installation process for SlackerNews.
+         
+         1. In a browser, navigate to the URL for the DNS record with a target port of 30000.
+
+            The Admin Console opens.
    
    1. On the Admin Console landing page, click **Start**.
```
paigecalvert, line 381: Note: For some reason, this page only appears for me when I use a non-CMX VM. Seems like when I use a CMX VM, I get jumped right to the log in screen. Curious if this happens for any of you when you go through the steps.

docs/vendor/quick-start.mdx
```diff
@@ -252,66 +386,96 @@ Before you begin, ensure that you have access to a VM that meets the requirement
...
 
-   This instance telemetry is automatically collected and sent back to the Vendor Portal by both KOTS and the Replicated SDK. For more information, see [About Instance and Event Data](/vendor/instance-insights-event-data).
+       ![Slackernews instance on customers page](/images/quick-start-slackernews-customer-instance.png)
 
-1. Under **Instance ID**, click on the ID to view additional insights including the versions of Kubernetes and the Replicated SDK running in the cluster where you installed the application. For more information, see [Instance Details](/vendor/instance-insights-details).    
+       [View a larger version of this image](/images/quick-start-slackernews-customer-instance.png)
+
+       This instance telemetry is automatically collected and sent back to the Vendor Portal by Embedded Cluster and the Replicated SDK. For more information, see [About Instance and Event Data](/vendor/instance-insights-event-data).
+
+    1. (Optional) Under **Instance ID**, click on the ID to view additional insights including the versions of Kubernetes and the Replicated SDK running in the cluster where you installed the application.
```
paigecalvert, line 441: ^ Not really relevant to this quick start, but something that I thought was weird was that I included v1.7.2 of the SDK in my release, but it was showing up as v1.7.0 in the instance insights. Curious if the same thing happens for you all.

docs/vendor/quick-start.mdx
```diff
@@ -339,26 +504,38 @@ Before you begin, ensure that you have access to a VM that meets the requirement
...
-   1. From the `manifests` directory, create and promote a new release, setting the version label of the release to `0.0.2`:  
+   1. From the `manifests` directory, lint the release manifests:
+
+      ```bash
+      replicated release lint --yaml-dir .
+      ```
+
+      :::note
+      If you see a `helm-archive-missing` error, confirm that you have only one chart archive in `manifests` for version 0.4.15, and that the `chartVersion` in the HelmChart resource is also set to `0.4.15`. If you notice that any of the errors are from files in your `.history/` directory, run `rm -rf .history` and then lint the files again.
+      :::
```
paigecalvert, line 536: >If you notice that any of the errors are from files in your `.history/` directory, run `rm -rf .history` and then lint the files again.

These seems to happen to me a lot. Not sure if there's a better way to handle this (maybe customizing the linter to ignore .history?)

docs/vendor/quick-start.mdx
```diff
@@ -24,119 +25,178 @@ The goals of this quick start are to introduce new Replicated users to the follo
...
+* **Option 1: Use Compatibility Matrix.** To use Replicated Compatibility Matrix to create a VM, do the following before proceeding:
+   
+   * Request Compatibility Matrix credits. You can request credits by creating a Vendor Portal account and then going to [**Compatibility Matrix > Request more credits**](https://vendor.replicated.com/compatibility-matrix) in the Vendor Portal. For more information about creating an account, see [Create a Vendor Account](vendor-portal-creating-account). For more information about Compatibility Matrix credits, see [Billing and Credits](/vendor/testing-about#billing-and-credits).
+   
+   * Ensure that you have an SSH key in your GitHub account. Then, add your GitHub username to your Vendor Portal [**Account Settings**](https://vendor.replicated.com/account-settings). This will provide SSH access to VMs that you create with Compatibility Matrix. For more information, see [Set Up SSH Access](/vendor/testing-vm-create#set-up-ssh-access) in _Create VMs_.
+
+   After you complete the prerequisites above, continue to the [Quick Start](#quick-start). You will create the VM with Compatibility Matrix as part of the tutorial.
+
```
paigecalvert, line 39: ^ adds CMX option

docs/vendor/quick-start.mdx
```diff
@@ -199,50 +269,114 @@ Before you begin, ensure that you have access to a VM that meets the requirement
...
+      :::note
+      If you are prompted to add the fingerprint for replicatedvm.com, type `yes` and press Enter.
+      :::
+
+   1. When prompted, provide the passphrase for the SSH key in your linked GitHub account. Compatibility Matrix uses GitHub SSH to provide access to the VM. 
+
 1. Install the application with Embedded Cluster:
      
-    1. On the page for the customer that you created, click **Install instructions > Embedded Cluster**.
+    1. In the Vendor Portal, on the page for the customer that you created, click **Embedded Cluster install instructions**.
```
paigecalvert, line 317: ^ I opted for showing people how to find the install instructions in the vendor portal versus going through the steps to access the enterprise portal. 

This was already feeling pretty long for a "quick" start. I'm going to need to update the other tutorials too, so can do a version that shows enterprise portal

crdant: I had a note about this in my notes, it's a great point.

paigecalvert:
@crdant thanks! comments below

> Some notes that are as much about the Quick Start itself since a lot of my comments aren't on changes
> 
> * Is there an appetite for linking to/incorporating [Set Up Your Environment](https://docs.replicated.com/vendor/environment-setup)?

For sure, I can add a link. I was a little light-handed on links in general for this quick start with the intention to help keep people on this page and limit "distractions"

> * I don't think I care about the Helm chart version, but it's kind of arbitrary and not necessarily consistent with the production SlackerNews

Very true. Might as well be 1.0.0 or something like that.

> * Maybe we should change the `data:` URL to an `https` one to make `Chart.yaml` seem less intimidating
>   
>   * Use the GitHub avatar?
>   * Actually, we have [this one](https://uploads-ssl.webflow.com/6310ad0e6a18aa1620da6ae8/6330e04f42bc6a7ba03b4725_snicon.png) in `replicated-app.yaml`

Good idea. That's an easy one.

> * EC version is updated and we should probably use it since the 1.31 version of k8 is almost out of support

Thanks 👍 

> * Should be bring them to Enterprise Portal instead?
>   > In the Vendor Portal, on the page for the customer that you created, click **Embedded Cluster install instructions**.

Covered in the other comment thread

> * Do we talk about "Admin Console" anywhere before this? If not, does it need to be described a bit?
>   > When prompted by the `install` command, set a password for accessing the Admin Console. You will use this password to log in to the Admin Console in a later step.

We do talk about it in the step for configuring the manifests for the release. But it's not really explained well there either. I'll add something to the intro to call out this UI specifically

> * There's a wrinkle here if you its shows the private IP address, since you need to access it from the public one
>   > - If you brought your own VM, go to the URL provided in the output of the install command.

@crdant Ah okay. I didn't realize that--I always just go to link so I suppose I'm not sure then why that works. Is that something where we'd be able to tell them to make a change in their VM settings to allow them to access this? Or do you have a suggestion?

> * My browser switched to `443` automatically when I went to the amdin console, so I missed the steps around the certificate, maybe a note there?

I think this was another one that I added a comment on in the Files changed tab. This happens to me too when I use CMX, but not when I use a google cloud VM. Were you also using CMX?

> * I have broken icons, there was a fix I made to SlackerNews in the last couple of days to address that that should be in your image but may not be

I noticed that too...I did pull the latest down from your repo before repackaging the helm chart for this tutorial, but sounds like it might have been a change that's just in the securebuild image and not the public one that this points to?

> * Personal preference, but I like to use the long flag in tutorials and scripts when I'm not sure if someone will know the command, so `helm package -u .` would be `helm package --dependency-update .`

I'm down to change to the full flag

crdant:
> > I have broken icons, there was a fix I made to SlackerNews in the last couple of days to address that that should be in your image but may not be

> I noticed that too...I did pull the latest down from your repo before repackaging the helm chart for this tutorial, but sounds like it might have been a change that's just in the securebuild image and not the public one that this points to?

This is in the image. Let me make sure the public one matches the latest private one.

docs/vendor/quick-start.mdx
```diff
@@ -199,50 +269,114 @@ Before you begin, ensure that you have access to a VM that meets the requirement
...
+
+      http://10.0.0.11:30000
+      -----------------------------------------------------------------
       ```
 
       At this point, the cluster is provisioned and the Admin Console is deployed, but the application is not yet installed.
 
-   1. Go to the URL provided in the output to access to the Admin Console.
+   1. Do one of the following to access the Admin Console:
+       * If you brought your own VM, go to the URL provided in the output of the install command.
```
crdant, line 357: We could do something like
> :::note The IP address in the provided URL may be the private IP address of your virtual machine. You may need to edit the URL to use the public IP address. :::

PR #3507, opened 2025-09-02
Marking a release as no longer required in KOTS 1.126.0 and later

paigecalvert:
Screenshot 2025-09-03 at 9 12 01 AM

https://deploy-preview-3507--replicated-docs.netlify.app/vendor/releases-about#release-properties


docs/partials/releases/_required-releases-limitations.mdx
```diff
@@ -3,4 +3,4 @@ Required releases have the following limitations:
   * Required releases are supported in KOTS v1.68.0 and later.
   * After users deploy a required version, they can no longer redeploy (roll back to) versions earlier than the required version, even if `allowRollback` is true in the Application custom resource manifest. For more information, see [`allowRollback`](/reference/custom-resource-application#allowrollback) in the Application custom resource topic.
   * If you change the channel an existing customer is assigned to, the Admin Console always fetches the latest release on the new channel, regardless of any required releases on the channel. For more information, see [Channel Assignment](licenses-about#channel-assignment) in _About Customers_.
-  * Required releases are supported for KOTS installations only and are not supported for releases installed with Helm. The **Prevent this release from being skipped during upgrades** option has no affect if the user installs with Helm.  
\ No newline at end of file
+  * Required releases are not supported for Helm CLI installations. The **Prevent this release from being skipped during upgrades** option has no affect when the user installs with Helm.
```
paigecalvert, line 6: Was drafting this new limitation to clarify how the requirement status doesn't get updated after an upstream update check in 1.125.2 and earlier:

> * For installations using KOTS 1.125.2 or earlier, if you disable the **Prevent this release from being skipped during upgrades** option for a promoted release, the release will still be marked as required in the Admin Console. 

Wasn't sure if there is some sort of workaround in this case (ie, the status doesn't change in the Admin Console but you can skip the previously required version if you upgrade from the CLI or something), or if you'd just need to upgrade to KOTS 1.126.0.

paigecalvert: Gotcha. I'll add one more sentence to make sure that's clear

PR #3545, opened 2025-09-23
Documents new custom field types

crdant:
TL;DR
-----

Adds the new types of custom license fields

Details
-------

Updates the docs on custom license fields to include the "Password" and "Enum" types. Passwords are essentially strings that are considered sufficiently sensitive to be masked in the UI, while enums allow you to specify a list of options for the users to choose from.

Provides more description of each type by moving the list of types into its own section and putting the types in a table with the name, an explanation, and the type of widget used on the "Manage Customer" screen.


crdant:
> a little reordering so that this could be a h3 and visible from the right hand toc on the page

I was wondering if that might make more sense, then decided it wasn't enough content. I like it split out, though, it also leaves room for more detail on the future.

PR #3572, opened 2025-10-07
Updated troubleshoot docs to add brew instructions

NoaheCampbell:
Troubleshoot docs now instruct mac users to install preflight and support bundle via homebrew instead of krew, and still say to use krew for windows and linux.

Not looking to merge this right away since v1beta3 has not been shipped yet, this is just in preparation of it shipping

### DIFF
--- a/docs/partials/support-bundles/_install-plugin.mdx
+++ b/docs/partials/support-bundles/_install-plugin.mdx
@@ -22,7 +22,7 @@ For Windows and Linux users, use Krew to install the support-bundle plugin:
     curl https://krew.sh/support-bundle | bash
     ```
 
-* If krew is installed already, run the following command to install the plug-in:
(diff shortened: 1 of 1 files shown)


docs/partials/support-bundles/_install-plugin.mdx
```diff
@@ -1,30 +1,36 @@
...
-    ```
+
+    ```bash
     curl https://krew.sh/support-bundle | bash
     ```
-    
-* If krew is installed already, run the following command to install the plug-in: 
 
-    ```
+* If krew is installed already, run the following command to install the plug-in:
```
paigecalvert, line 25:
```suggestion
* If krew is installed already, run the following command to install the plugin:
```

paigecalvert (approved):
lgtm! minor edit

paigecalvert (approved):

PR #3620, opened 2025-10-28
chore: add blurb about additional AKS annotations required for loadbalancers to work

mikhailswift:
Customer reported that they were getting time outs when using the External IP to access a service they deployed to an AKS cluster on CMX. Upon investigation it was found that AKS requires some annotations to be set for Azure to know that the service is ready to receive traffic before it's ready.

https://app.shortcut.com/replicated/story/130440/document-capabilities-to-ingress-to-cloud-based-cluster

I tested the other cloud distros and found that AKS was the only one any special steps are required for, so a note for AKS in the docs felt appropriate.


paigecalvert (approved):
thank you!

PR #3623, opened 2025-10-29
SDK 1.10.0 release notes and docs

laverya:


docs/vendor/replicated-sdk-customizing.md
```diff
@@ -572,3 +572,29 @@ To enable reporting all images, set the `replicated.reportAllImages` value in yo
 replicated:
   reportAllImages: true
 ```
+
+## Proxy Configuration {#proxy}
+
+With the Replicated SDK version 1.10.0 and later, you can configure the SDK to use an HTTPS proxy when fetching license information and reporting metrics.
+
+When enabled, the SDK will use the configured proxy for requests to Replicated APIs, but will not use a proxy for requests to the Kubernetes API. Setting an additional no_proxy is not required.
```
paigecalvert: >Setting an additional no_proxy is not required.

note the minor reword here

PR #3681, opened 2025-12-01
Update documentation for installer-only annotation

AmberAlston:
Clarified the usage of the kots.io/installer-only annotation for both HelmChart and standard Kubernetes resources, including examples and troubleshooting information.

Preview link: https://deploy-preview-3681--replicated-docs.netlify.app/reference/custom-resource-helmchart-v2#kotsioinstaller-only-annotation


AmberAlston:
@paigecalvert added this PR as a result of this explanation to a customer question on the recent annotation feature (see slack https://replicated.slack.com/archives/C01CE41DDUM/p1764611853840209?thread_ts=1760389797.156219&cid=C01CE41DDUM) They thought they were restricted to helm charts only based on the content and placement, but the code shows it works for other K8s resources as well. Maybe it needs to be referenced in another section that helmchartv2. What do you think? 

cc @laverya for visibility

AmberAlston:
@paigecalvert That's a great question. It depends on the app architecture but in real application the primary use case for kots.io/installer-only on standard K8s resources is when you have resources INSIDE a Helm chart that should conditionally deploy based on installation method.

i had claude help me better summarize it in a follow up pr and then cut down the content a bit more from there. 

Take a look and see if you think this is clearer, and if keeping it here or moving the page still makes sense given this context 

https://deploy-preview-3681--replicated-docs-upgrade.netlify.app/reference/custom-resource-helmchart-v2#kotsioinstaller-only-annotation

docs/vendor/packaging-include-resources.md
```diff
@@ -1,24 +1,18 @@
...
+Software vendors often need a way to conditionally deploy one or more application resources. One common use case is giving the user the choice to use an external database or an embedded database. In this scenario, when a user chooses to use their own external database, it is not desirable to deploy the embedded database resources.
 
-There are different options for creating conditional statements to include or exclude resources based on the application type (Helm chart- or standard manifest-based) and the installation method (Replicated KOTS or Helm CLI).
+Another common use case is needing to explicitly include or exclude a resource from a customer deployment depending on if they install with the Helm CLI or with a Replicated installer. For example, some vendors provide additional services when the user deploys with a Replicated installer like Embedded Cluster, but do not want those services included when users deploy with the Helm CLI.
```
paigecalvert: ^ added this other use case example to the overview

paigecalvert (approved):
@AmberAlston details of my changes in the comments!

PR #3721, opened 2025-12-16
Add claude subagent for docs reviews

paigecalvert:


docs/templates/procedure.md
paigecalvert: these are unused

PR #3978, opened 2026-04-20
docs: clarify that the ec distro in cmx is specifically ec v2

mikhailswift:
https://app.shortcut.com/replicated/story/136304/cmx-make-it-apparent-the-ec-distro-is-ecv2


paigecalvert (approved):
thanks!

PR #4003, opened 2026-04-28
docs(custom-domains): warn that switching default custom domain breaks existing installs

Justiceleeg:
Adds an important admonition in "Set a default domain" explaining that changing the default custom domain for the registry or proxy registry causes image pulls to fail for installations of releases promoted under the previous default, and that customers should upgrade first.

from this story: https://app.shortcut.com/replicated/story/136331/custom-domain-activation-breaks-existing-sdk-releases


paigecalvert:
@Justiceleeg question on this one as well https://replicated.slack.com/archives/C02BPJQQP28/p1777929730068529

paigecalvert:
Closing this per Dmitriy's feedback in: https://replicated.slack.com/archives/C02BPJQQP28/p1777929730068529

The bug that caused image pull failures for existing installations that were using proxy.rereplicated.com after the vendor added a custom domain has been fixed. There's no action required for vendors.

PR #4008, opened 2026-04-30
feat: document how to brand replicated sdk

banjoh:


paigecalvert (approved):
thanks @banjoh ! mostly small copy edits, and also moved this to the Customize the SDK  page instead of Install the SDK

PR #4035, opened 2026-05-06
docs(vm): clarify --ssh-public-key creates user from key comment

jdewinne:
## Summary

Adds documentation explaining how the Linux username is derived from the SSH public key comment when using the `--ssh-public-key` flag with `replicated vm create`.

- The username is taken from the portion of the key's comment before the first `@`.
- Users should pass that username via `--username` when using `replicated vm ssh-endpoint` or `replicated vm scp-endpoint`.

This matches the clarification recently added in the CLI repo: https://github.com/replicatedhq/replicated/pull/708

### DIFF
--- a/docs/vendor/testing-vm-create.md
+++ b/docs/vendor/testing-vm-create.md
@@ -110,8 +110,6 @@ Or, to use multiple SSH public keys:
 replicated vm create --distribution ubuntu --version 24.04 --ssh-public-key ~/.ssh/id_rsa.pub --ssh-public-key ~/.ssh/id_ed25519.pub
 ```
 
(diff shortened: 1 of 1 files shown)


docs/vendor/testing-vm-create.md
```diff
@@ -110,6 +110,8 @@ Or, to use multiple SSH public keys:
 replicated vm create --distribution ubuntu --version 24.04 --ssh-public-key ~/.ssh/id_rsa.pub --ssh-public-key ~/.ssh/id_ed25519.pub
 ```
 
+The Linux user provisioned on the VM is derived from the comment in the SSH public key: the portion before the first `@`. For example, a key with the comment `ci@host` creates a user named `ci`. When you connect to the VM using `replicated vm ssh-endpoint` or `replicated vm scp-endpoint`, pass that username with the `--username` flag so the endpoint matches the user the key was added to.
+
```
paigecalvert: question https://replicated.slack.com/archives/C02BPJQQP28/p1778096436240619?thread_ts=1778090231.597179&cid=C02BPJQQP28

paigecalvert (approved):
@jdewinne thank you for this! I moved the info down to the steps from direct ssh and scp (we already had some info about this there, so this just expands on it)

PR #4045, opened 2026-05-08
Update Application CR for ECv3

paigecalvert:
https://deploy-preview-4045--replicated-docs.netlify.app/reference/custom-resource-application


docs/reference/template-functions-static-context.md
```diff
@@ -12,33 +12,33 @@ This topic provides a list of the Replicated template functions in the Static co
...
+For Embedded Cluster installations, the ConfigMap returned by PrivateCACert contains the CA trust store from the host. Embedded Cluster determines the CA trust store by first checking the `SSL_CERT_FILE` environment variable. If `SSL_CERT_FILE` is not set, Embedded Cluster then searches common certificate paths and uses the first valid certificate file found.
+
+For KOTS installations in existing clusters, the end user can optionally create and pass the ConfigMap to the `install` command using the `--private-ca-configmap` flag. For more information, see [install](/reference/kots-cli-install).
+
+#### About the ConfigMap and `SSL_CERT_DIR` environment variable
+
+The Replicated installer mounts the ConfigMap returned by the PrivateCACert template function as a volume. Each key in the ConfigMap is created as a file, with its value as the file's contents. The installer then sets the `SSL_CERT_DIR` environment variable to the volume. `SSL_CERT_DIR` is a common environment variable that most tools and languages support to override the trust store.
```
emosbaugh: i dont think it is correct. it was more correct before since this is unique to kots and ec v2 which runs kots. for v3 everything is run on the host so no need to pass anything to our containers. 

i wonder if this is TMI anyway? perhaps it is enough to say that kots and ec will use the private ca from the host or the flag when making outbound network requests.

docs/reference/template-functions-static-context.md
```diff
@@ -12,33 +12,33 @@ This topic provides a list of the Replicated template functions in the Static co
...
+
+#### About mounting the ConfigMap in your application containers
 
-The primary use case for the PrivateCACert template function is to make sure that KOTS trusts the private CA certificates issued by TLS man-in-the-middle (MITM) proxies in the end user's environment and can make outbound internet connects without TLS errors.
+You can use the PrivateCACert template function to mount the ConfigMap in your application containers. 
 
-<details>
-  <summary>Can the ConfigMap returned by PrivateCACert also be mounted in application containers?</summary>
+When you mount the ConfigMap in your containers, the `SSL_CERT_DIR` environment variable overrides the trust store in the container. This means that the container trusts only the CAs included in the ConfigMap.
```
emosbaugh: im not sure what this is suggesting. i dont think we explicitly set SSL_CERT_DIR in the vendors container.

emosbaugh: this was more correct prior to the update.

emosbaugh (changes requested):

docs/reference/template-functions-static-context.md
paigecalvert: >i wonder if this is TMI anyway? perhaps it is enough to say that kots and ec will use the private ca from the host or the flag when making outbound network requests.

that works for me. I'll pare it down; we can see if it's something we end up really getting any questions and revisit as needed

emosbaugh (approved):

PR #4079, opened 2026-05-19
Document config value defaults behavior change in EC v3 migration guide

AmberAlston:
## Summary

- Adds a new section to the EC v2-to-v3 migration guide documenting the config value defaults behavior change
- In v2 (KOTS), only explicitly-set values were persisted and defaults were re-evaluated on each release
- In v3, all config values (including defaults) are resolved and persisted at install time, so changing a default in a new release will not take effect on existing installations
- Recommends using `value` on readonly/disabled items instead of `default` for vendor-controlled fields that need to be updateable across releases

This was surfaced by a vendor doing a proof-of-concept migration to v3 who hit this behavioral difference and asked if it was expected.

## Test plan

- [ ] Verify the new section renders correctly in the docs site
- [ ] Confirm placement in the "Comparison to Embedded Cluster v2" section is logical alongside the other behavioral differences

Generated with [Claude Code](https://claude.com/claude-code)


AmberAlston:
Netlify is completely down right now. I'm not going to wait for it with this small of a change

PR #4148, opened 2026-06-08
Add guidance for choosing Support Bundle Analyzed over Uploaded

AmberAlston:
Preview links
* https://deploy-preview-4148--replicated-docs.netlify.app/reference/notifications-events-filters#support-events
* https://deploy-preview-4148--replicated-docs.netlify.app/vendor/event-notifications-create#customer-success-manager-support-bundle-analyzed
* https://deploy-preview-4148--replicated-docs.netlify.app/vendor/event-notifications#instance-tag-filters

## Summary

- Adds a tip callout to the events/filters reference explaining when to use Analyzed vs Uploaded
- Changes the CSM example notification from "Support Bundle Uploaded" to "Support Bundle Analyzed" with a tip explaining why Analyzed is the better default for most workflows
- Documents in the instance tag filters section that tag filtering is available on Analyzed but not Uploaded, and why

## Context

Vendors tend to default to the Support Bundle Uploaded event without realizing that Analyzed provides richer context (metadata, bundle name, resolved instance identity) and is the only support bundle event that supports instance tag filtering. The Uploaded event fires before extraction, so for manual, Enterprise Portal, and vendor upload paths, the instance hasn't been identified yet and tag-based routing won't apply.

Related: [sc-137964](https://app.shortcut.com/replicated/story/137964)

## Test plan

- [ ] Verify the tip callout renders correctly on the events/filters reference page
- [ ] Verify the updated CSM example reads clearly on the create subscriptions page
- [ ] Verify the new paragraph in the instance tag filters section flows with the existing content

### DIFF
--- a/docs/reference/notifications-events-filters.md
+++ b/docs/reference/notifications-events-filters.md
@@ -634,7 +634,7 @@ When a customer pulls a release asset (Helm chart, Embedded Cluster bundle, or p
 ## Support events
 
 :::note
-Use **Support Bundle Uploaded** when you need an immediate notification that a bundle has arrived. Use **Support Bundle Analyzed** when you need to filter or route based on instance tags, bundle metadata, or bundle name. The Analyzed event fires after the bundle is extracted and the instance is identified, so it has richer context and supports instance tag filtering. If you are unsure which to use, start with Support Bundle Analyzed.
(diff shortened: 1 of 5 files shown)


docs/reference/notifications-events-filters.md
```diff
@@ -633,6 +633,10 @@ When a customer pulls a release asset (Helm chart, Embedded Cluster bundle, or p
 
 ## Support events
 
+:::note
+Use **Support Bundle Uploaded** when you need an immediate notification that a bundle has arrived. Use **Support Bundle Analyzed** when you need to filter or route based on instance tags, bundle metadata, or bundle name. The Analyzed event fires after the bundle is extracted and the instance is identified, so it has richer context and supports instance tag filtering. If you are unsure which to use, start with Support Bundle Analyzed.
```
seanoseanohay: Analyzed conditionally attaches tags. So, the code still allows analyzed events without tag payloads. It is possible for an instance to be legitimate and still have 0 tags.

Should we avoid implying every analyzed event always has a resolved instance and tags?

seanoseanohay (approved):

PR #4170, opened 2026-06-12
Clean up EC v3 docs: remove stale limitations, legacy refs, fix examples

AmberAlston:
## Summary
- Remove stale custom node roles limitation from overview (custom roles with labels and taints have shipped)
- Replace vague "some template functions are not implemented" limitation with the specific list of 5 unsupported legacy functions and point to their replacements
- Remove unnecessary kURL/KOTS reference from air gap install overview
- Remove Embedded Cluster 2.x comparison notes from online install page (confusing for new v3 users who never used v2)
- Move v2 Setup page context to the migration page where v2 users will find it
- Fix multi-node description to include CLI and external API join methods (not just UI)
- Fix NVIDIA GPU operator example to use the v3 `extensions.helmCharts` format instead of the old `extensions.helm.repositories/charts` format
- Fix v2/v3 typo in proxy auth section of migration page ("Embedded Cluster v2 configures" should be "v3")
- Add note to migration page that v3 releases won't appear in the v2 KOTS Admin Console, so vendors need to steer customers to EP or communicate out-of-band
- Clarify that v2-to-v3 migration uses the same `upgrade` command as any v3-to-v3 upgrade

## Test plan
- [ ] Local docs preview renders all changed pages correctly
- [ ] All cross-page links resolve
- [ ] NVIDIA GPU operator example validates against EC Config schema

### DIFF
--- a/embedded-cluster/embedded-using.mdx
+++ b/embedded-cluster/embedded-using.mdx
@@ -208,25 +208,24 @@ Using the NVIDIA GPU Operator with Embedded Cluster requires configuring the con
 
 ```yaml
 # Embedded Cluster Config
-
+apiVersion: embeddedcluster.replicated.com/v1beta1
+kind: Config
+spec:
   extensions:
-    helm:
-      repositories:
-        - name: nvidia
-          url: https://nvidia.github.io/gpu-operator
-      charts:
-        - name: gpu-operator
-          chartname: nvidia/gpu-operator
-          namespace: gpu-operator
-          version: "v24.9.1"
-          values: |
-            # configure the containerd options
-            toolkit:
-             env:
-             - name: CONTAINERD_CONFIG
-               value: /etc/k0s/containerd.d/nvidia.toml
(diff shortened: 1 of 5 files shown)


embedded-cluster/embedded-using.mdx
```diff
@@ -208,25 +208,24 @@ Using the NVIDIA GPU Operator with Embedded Cluster requires configuring the con
...
-          version: "v24.9.1"
-          values: |
-            # configure the containerd options
-            toolkit:
-             env:
-             - name: CONTAINERD_CONFIG
-               value: /etc/k0s/containerd.d/nvidia.toml
-             - name: CONTAINERD_SOCKET
-               value: /run/k0s/containerd.sock
+    helmCharts:
```
sgalsaleh, line 215: should this section mention that the vendor should add the chart's .tgz file to the release now that there is no `repositories` section?

sgalsaleh (dismissed):

PR #4185, opened 2026-06-16
Document EP version picker fallback behavior

NoaheCampbell:
Documents the Enterprise Portal version picker behavior for releases that do not have matching content branches.

This adds guidance for the new Require matching content for release versions setting, including when vendors should leave fallback behavior enabled and when they should require explicit version-specific content.


AmberAlston (approved):

AmberAlston:
@NoaheCampbell this is good to merge when you're ready

PR #4206, opened 2026-06-24
Remove a note about deleting images from registry not being supported

divolgin:
Deleting image tags is now supported in Vendor Portal.

Screenshot 2026-06-25 at 4 07 42 PM


AmberAlston (changes requested):
See comment in slack, needs to convey how to delete an image rather than just delete the limitation

AmberAlston (approved):

PR #4249, opened 2026-07-10
Document Enterprise Portal service account automation

NoaheCampbell:
Documents Enterprise Portal service account automation for the new Enterprise Portal.

Adds a new Service Account Automation page that explains why vendors should include the built-in automation docs in their Enterprise Portal content repo, how the default API reference and workflow guide pages work, and what customers can use them for.

The new page covers:

  - Customer service account tokens for CI/CD and automation.
  - The difference between Enterprise Portal service account APIs and the Replicated Vendor API.
  - How to add the default APIReference and HeadlessWorkflowGuide components to an Enterprise Portal content repo.
  - How vendors can customize surrounding MDX content.
  - The supported APIReference description prop.
  - What endpoints and workflows the API reference and workflow guide document.
  - Where customers create service account tokens in the Enterprise Portal.

  Also updates the new Enterprise Portal overview and use pages to mention service account automation.


AmberAlston:
Readability nit - The intro section has 5 consecutive single-sentence paragraphs that read a bit choppy. Combining a couple would flow better.

I also had the 🤖 review the docs against the actual backend routes, Next.js proxy layer, MDX component registry, and content template repo, and it flagged the following things. Can you please take a look at these concerns it raised. 

1. update-instructions is not an SA endpoint

The swagger spec in #10160 documents GET /api/v1/install-options/{id}/update-instructions as a service-account endpoint, but the route isn't registered in apiEPServiceAccountGroup (daemon.go lines 352-367), and there's no Next.js proxy route for it. It only exists in the browser-session-authenticated group (daemon.go line 419). #10160 doesn't modify daemon.go.

This means the <APIReference /> component will render an update-instructions endpoint that returns 401 when a customer tries it with an SA token.

The docs PR references "install and update instructions" in a few places. Either the route needs to be added to the SA group + proxy layer in #10160, or the update-instructions references should come out of both PRs until that's wired up.

2. "The default Enterprise Portal content template includes an Automation section" — not yet true

The template repo (replicatedhq/enterprise-portal-content) has no automation pages, no <APIReference /> usage, and no <HeadlessWorkflowGuide /> usage. #10160 doesn't update the template repo either. Vendors would need to manually add the pages using the instructions on this docs page.

The docs page should either say "you can add" instead of "the template includes", or the template repo needs a follow-up PR to add the default pages.

AmberAlston (approved):

PR #4276, opened 2026-07-17
docs(ec): cover repo/OCI-sourced Helm extension charts in v2→v3 migration

sgalsaleh:
Fixes [SC-138905](https://app.shortcut.com/replicated/story/138905)

## Problem

Embedded Cluster v3 resolves Helm extension charts from `.tgz` archives **bundled in the release**. It does not pull extension charts at install time from an HTTP Helm repository or an OCI registry — the v3 `extensions.helmCharts` schema has no repository/registry field.

A vendor migrating a v2 extension whose chart was sourced from a repo or registry (e.g. `chartname: ingress-nginx/ingress-nginx` or `chartname: oci://quay.io/jetstack/charts/cert-manager`) hits a wall: the v2→v3 migration docs only showed the Config YAML shape change and never explained where the `.tgz` comes from. In one case a v2-shaped extensions config was promoted under a v3 apiVersion, the extension's images never made it into the air gap bundle, and the install failed.

## What this changes

Docs-only. Closes the migration-path gap by documenting the fetch-and-bundle procedure.

- **`embedded-v3-migrate.mdx`** — new comparison subsection *"Helm extension charts must be bundled in the release"* explaining the v2 (repo/OCI pull) → v3 (bundled `.tgz`) model; and expands the extensions migration step from a pure YAML swap into: `helm pull` (HTTP repo **and** OCI examples) → add the `.tgz` to the release → reference it in `extensions.helmCharts`.
- **`embedded-using.mdx`** (`#add-extensions`) — states the bundled-`.tgz` requirement and expands the steps to package/pull and reference the chart.
- **`embedded-config.mdx`** — adds a requirements bullet to the `extensions.helmCharts` reference.

Example chart sources were verified against upstream: `https://kubernetes.github.io/ingress-nginx` (valid repo, versions have no `v` prefix) and `oci://quay.io/jetstack/charts/cert-manager` (canonical OCI location; the `1.19.2` tag exists on quay.io).

## Out of scope

The related asks to add a **promote-time lint for a missing chart archive / HelmChart CR** and to **fetch-and-bundle at the platform level during promote** are code/product changes in other repos, tracked separately.

### DIFF
--- a/embedded-cluster/embedded-config.mdx
+++ b/embedded-cluster/embedded-config.mdx
@@ -202,6 +202,8 @@ Each extension chart is installed with `helm upgrade --install --wait --wait-for
 
 #### Requirements
 
+* Each chart referenced in `helmCharts` must be packaged as a `.tgz` chart archive and included in the release. Embedded Cluster resolves extension charts from bundled archives by `chart.name` and `chart.chartVersion`; it does not pull them from an HTTP Helm repository or an OCI registry at install time. See [Add Helm chart extensions](embedded-using#add-extensions).
+
 * You must provide `chart.chartVersion`. Omitting it can cause upgrade issues.
 
 * For air gap bundles, images used by Helm extensions must not refer to a multi-architecture image by digest. Air gap bundles include only x64 images, and the digest for the x64 image will be different from the digest for the multi-architecture image, preventing Kubernetes from locating the image in the bundle. For an example of how to set digests to empty strings and pull by tag only, see the `ingress-nginx` [Example](embedded-config#extension-example).
--- a/embedded-cluster/embedded-config.mdx
+++ b/embedded-cluster/embedded-config.mdx
@@ -204,6 +204,8 @@ Each extension chart is installed with `helm upgrade --install --wait --wait-for
 
 * Each chart referenced in `helmCharts` must be packaged as a `.tgz` chart archive and included in the release. Embedded Cluster resolves extension charts from bundled archives by `chart.name` and `chart.chartVersion`; it does not pull them from an HTTP Helm repository or an OCI registry at install time. See [Add Helm chart extensions](embedded-using#add-extensions).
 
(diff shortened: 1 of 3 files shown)


AmberAlston:
The `.tgz` bundling coverage looks good — this closes the gap that was causing vendors to hit install failures when migrating repo/OCI-sourced extensions.

One thing I think is missing: should the migration steps also mention creating a HelmChart v2 CR for each extension chart? The existing step 6 in the migration doc says "Ensure that your release has a corresponding HelmChart v2 custom resource for each of your Helm charts," but it's not clear whether that includes extension charts. The new extension-specific steps (helm pull → add .tgz → reference in EC Config) don't mention the CR at all.

This matters for Security Center — without a HelmChart CR, the extension chart's images don't get scanned. The validation currently exempts extension charts from the CR requirement, so vendors won't get a lint error, but they'll silently lose image scanning coverage.

If extension charts should have HelmChart CRs (and I think they should based on the docs stating every chart needs one), it would be worth adding that as a sub-step here.

AmberAlston:
Update on the HelmChart CR concern above: Salah clarified that extension charts don't need a separate HelmChart CR — the schema lives inside the EC Config, and Security Center scans extensions without one. So the migration steps in this PR are correct as-is.

That said, this is worth making explicit in a few spots since the existing docs are ambiguous:

1. **Step 6** in the migration steps ("Ensure that your release has a corresponding HelmChart v2 custom resource for each of your Helm charts") — a vendor reading this right after the new extension steps could think they also need a CR for extension charts. Worth noting that this applies to application charts, not extension charts defined in the EC Config.

2. **"HelmChart v2 required" comparison section** — same ambiguity. A note that extension charts are handled through the EC Config instead of a separate CR would help.

3. **`extensions.helmCharts` requirements in embedded-config.mdx** — could add a bullet noting no separate HelmChart CR is needed.

These are small clarifications — happy to push commits to this PR if you'd prefer.

AmberAlston:
I'll go ahead and contribute to this pr since they are small edits

AmberAlston (approved):

PR #4289, opened 2026-07-21
Clarify Login as customer is a full customer-admin session

AmberAlston:
Preview link - https://deploy-preview-4289--replicated-docs.netlify.app/vendor/enterprise-portal-v2-access#login-as-customer

## What

The Enterprise Portal "Login as customer" docs framed the feature as a benign "view as a customer" preview. It didn't convey that the button opens a **full customer-admin session** where any action is real and visible to the customer, and it didn't explain how to undo changes made in that session.

This adds to the **Login as customer** section of `enterprise-portal-v2-access.mdx`:

- A `:::note` warning that the session is full customer-admin, not read-only/preview, and that starting installs or creating/revoking service accounts affects the customer directly.
- A new **Undoing changes made while logged in as a customer** subsection, explaining how to undo the two artifacts a session can create:
  - Discard a pending installation from the installations list (the discard control was restored in Enterprise Portal v2 by vandoor #10256). Discarding also permanently deletes the service account created for that installation, so both are removed in one step.
  - Revoke a standalone service account from the team settings **Service accounts** tab.

## Why

A vendor started an install flow and created a service account while logged in as a customer, then couldn't find a way to remove them (the Vendor Portal sections are read-only). The failure mode was confusion about the privilege level and where to undo, not a permissions bug. A companion Vendor Portal UI change (info-icon tooltip near the button) is tracked separately.

### DIFF
--- a/docs/vendor/enterprise-portal-v2-access.mdx
+++ b/docs/vendor/enterprise-portal-v2-access.mdx
@@ -60,3 +60,13 @@ To view the production Enterprise Portal as a specific customer, use the **Login
 This is useful for verifying that a customer sees the correct content, entitlements, and branding without creating a permanent user account in their portal team. The system creates a temporary system user scoped to that customer and generates a magic link nonce that expires after 10 minutes.
 
 The login routes to the correct portal version automatically. If the customer is on the new Enterprise Portal, the link opens `{appSlug}.enterpriseportal.app` (or your custom domain). If the customer is on Classic, the link opens `get.replicated.com`.
+
+:::note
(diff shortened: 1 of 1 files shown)


AmberAlston:
Holding this as a draft. Verification turned up that EP v2 has no discard control for pending installations in the UI (a regression from v1, where a trash icon + confirm modal existed). Documenting a self-serve "remove the install" flow now would describe functionality that isn't reachable.

Blocked on sc-139000 (restore the EP v2 pending-install discard UI). Once that lands, I'll add the install-discard step (there's a TODO marker in the file) and mark this ready.

The service-account revoke guidance and the customer-admin warning callout in this PR are accurate today and independent of that work.

PR #4371, opened 2026-08-13
Remove Factory dependency update workflow from repo

xavpaice:
The dependency update workflow is now handled elsewhere, so remove the `.factory-workflows/dependency-update.yaml` file and the Depot jobs that published and validated it.


kriscoleman (approved):

AmberAlston (approved):
Confirmed with Xav that dependency updates still run via Elasticclaw/Factory (workflow is defined centrally now, not published from this repo). Removal is clean — code search shows no remaining refs to `factory-workflows`, the workflow_dispatch input option is updated, and the two PRs are independent with no merge-order hazard. LGTM.

PR #4389, opened 2026-08-20
docs(ep): document SAML customer SSO as generally available

Bishibop:
Part 1 of 3 for sc-139273. SAML customer SSO shipped as generally available and the docs still describe it as Alpha behind a feature flag.

## Why

Migration `20260806125346_EnableEnterprisePortalSAMLAuthForAllTeams.sql` set `feature.defaultvalue = '1'` for `enterprise_portal_saml_auth` and deleted every per-team override. SAML is enabled for all teams by default. Four pages still told vendors that a feature flag must be enabled for their team and to contact their account representative.

The flag still exists and still gates in code (`app_features.go:184-185`), so the wording here is "enabled for all teams by default" rather than "the gate was removed".

## Changes

**Enterprise Portal (New)**
- `enterprise-portal-v2-invite.mdx` — drop `(Alpha)` from the SAML heading; state that SAML is enabled for all teams by default; add a Just-in-time user provisioning subsection
- `enterprise-portal-v2-use.mdx` — drop `(Alpha)` from the SAML SSO bullet and correct the provisioning rule
- `enterprise-portal-v2-troubleshooting.mdx` — new SAML login errors section

**Enterprise Portal (Classic)** — the same flag serves both portals (`EP SAML Auth (Both)`), so the Classic pages carried the same stale claim
- `enterprise-portal-invite.mdx`, `enterprise-portal-use.mdx` — remove three feature-flag notes, drop `(Alpha)` from three headings and four link texts

## Corrections to previously published behavior

Two statements were wrong, not just stale:

1. **JIT provisioning.** The docs said an account is activated when the email "matches a pending invitation". Verified in `saml_acs.go:254-259`: that requirement applies only to portal-initiated logins. IdP-initiated logins provision with no invitation.
2. **Domain allowlist.** Undocumented, and it governs SAML. `CheckCustomerDomainAllowed` runs inside `if !userExists` (`saml_acs.go:241-253`), so it gates account creation rather than login, and it runs *before* the invitation check — a new user from a disallowed domain gets `domain_not_allowed`, not `saml_invite_required`. The troubleshooting section lists the four codes in evaluation order for that reason.

Error codes verified verbatim: `domain_not_allowed` (`customer_allowed_domains.go:17`), `saml_invite_required` / `saml_invalid_email` / `saml_auth_failed` (`replv3/errors.go:19-21`). All four are user-visible (`MagicLinkLogin.tsx:39-54`). `saml_auth_failed` is documented as covering both the initiation and assertion steps, because it is raised from both.

## Out of scope

Does not touch the page-level `:::note Alpha Feature` banners or the `Enterprise Portal (New) (Alpha)` sidebar label. Those resolve to "Beta" at the EP v2 launch and belong to sc-138827. This PR only removes feature-level markers on SAML, which resolve to no marker at all — a blanket Alpha-to-Beta sweep would mislabel a GA feature as Beta.

Line-disjoint with sc-138827's PR by agreement: that PR touches lines 3-4 of the v2 pages, this one touches line 18 and below.

## Draft

Draft per team process. Content makes no Beta claim and is safe to merge before launch.


AmberAlston (approved):
Verified against sc-139273 Gap 1 and the market-api source. All five requirements land: Alpha/flag language stripped, domain-allowlist-gates-JIT documented, all four error codes in evaluation order, and the SP-initiated (needs invite) vs IdP-initiated (JIT-free) asymmetry.

Code checks that back the docs:
- Domain check runs inside `if !userExists` and before the invite check (saml_acs.go:240-259) — matches the documented ordering.
- `saml_auth_failed` is raised from both saml_initiate.go and saml_acs.go:404, so documenting it as covering both initiation and assertion is correct (and a good correction to the ticket's 'initiation-step' wording).

One non-blocking confirmation: the Classic-portal pages gain a new 'allowed domain list also governs SAML logins' claim. This checks out — there's a single SAML ACS handler shared by both portals — but since it's a new behavioral claim on a production page, worth a quick self-confirm that Classic exposes domain restrictions + SAML together as described. Approving on that basis. LGTM.

PR #4373, opened 2026-08-17
Refresh Security Center documentation

osc135:
## Shortcut 
https://app.shortcut.com/replicated/story/138787/docs-finalize-security-center-section-for-beta

## Summary

- Align the Vendor Portal documentation with the Incomplete scans filter and status refresh behavior.
- Document the New Enterprise Portal raw scan download setting and Fixed In release data.
- Correct the documented CVE detail fields and clarify KOTS and kURL coverage.

## Validation

- npx --no-install docusaurus build (passes with existing unrelated broken-link warnings)
- git diff --check

PR #4391, opened 2026-08-20
docs(ep-v2): add Portal Features page documenting Ask AI

Bishibop:
Part 3 of 3 for sc-139273. Ask AI is live and entirely undocumented. Parts 1 and 2 merged as #4389 and #4390.

## The provider is described but not named, deliberately

The data disclosure states what leaves the portal — the question, up to 10 prior messages, and the entitlement-filtered content — but calls the destination "a third-party AI provider" rather than naming it.

The provider is Groq, and it is absent from [Infrastructure and Subprocessors](https://github.com/replicatedhq/replicated-docs/blob/main/docs/vendor/policies-infrastructure-and-subprocessors.md), which today lists OpenAI and FireworksAI. Publishing a disclosure that names a processor missing from our own published list advertises the inconsistency. Adding a subprocessor is not a docs-team editorial call, and the listing is owned outside this PR with no timeline I can state. Rather than hold this page indefinitely — Ask AI is shipping to customers undocumented right now — the naming is deferred to a follow-up.

To be explicit about what that means for review: **this PR is complete and mergeable on its own.** The follow-up is not a loose end in this change, it is a separate change that becomes possible later. If the listing never happens, this page is still correct as written.

The model id is withheld for the same reason, and this part is not obvious: it reads as `openai/gpt-oss-120b`, and **OpenAI is on the subprocessor list**. Naming the model while withholding the provider would imply the processor is OpenAI, which is wrong. Provider and model are restored together or not at all.

The links to Infrastructure and Subprocessors are dropped for the same reason — a reader following one to learn who processes their AI data would find OpenAI and FireworksAI and draw the wrong conclusion.

**Follow-up, once Groq is listed:** restore the provider name and link, the model id, the sentence pointing at the subprocessor page, the Related topics entry, and `Groq` in `styles/config/vocabularies/ThirdPartyProducts/accept.txt` (Vale flags it as a spelling error otherwise). Roughly a five-line diff.

**What this PR does not do:** hide that data leaves. A vendor reading this page learns that enabling Ask AI sends their content and their customers' questions to a third party, and is told to check that against their own customer agreements. Only the identity is deferred.

## Why a new page

The **Portal Features** section of **Enterprise Portal > Content** holds three app-level settings and had no page describing it. sc-139271 needs a single **Learn more** link pointing at documentation covering all the toggles in the box, and #4373 documented the two Security Center toggles by topic on the Security Center page.

Rather than move that merged work, this page documents all three settings and links out to #4373's section for the Security Center detail. sc-139271 gets its single target; #4373 is untouched.

The page is a sibling of Customize Portal Content rather than a section inside it: that page is a content-authoring reference, and these are app settings.

## Changes

- New `docs/vendor/enterprise-portal-v2-portal-features.mdx`
- `security-center-enable-customer-access.mdx` — one reciprocal link (the only line changed outside the new file)
- `sidebars.js` — entry after `vendor/enterprise-portal-v2-content`

## No availability admonition, deliberately

This is currently the only page in the Enterprise Portal (New) section without a `:::note Alpha Feature` block. That is intentional, not an oversight.

sc-138827 is replacing those 16 blocks with `:::note Beta Feature` at the EP v2 Beta launch. Adding one here would introduce, in a brand-new file, the exact string that sweep exists to remove. Publishing a Beta banner now is also not an option: the process rule is that public docs announcing Beta merge at launch, not before.

Note that the earlier version of this PR argued this page would very likely merge *after* the sweep. With the subprocessor gate removed that is no longer a safe bet — this may well land first, and be the only page of 17 with no marker until the sweep runs.

**For whoever runs the sc-138827 sweep:** this page needs the "additional access" variant, not the plain one, since every capability it documents is gated (Security Center by team, Ask AI by a per-vendor toggle, and the Security Center settings by their own RBAC resource):

```
:::note Beta Feature
Features described on this page are Beta and subject to change. Some capabilities might require additional access.
:::
```

## Defaults, and how they were derived

Two of the three defaults are `0` in the schema and read as "Disabled". The third does not, and that is worth a reviewer's eye:

| Setting | Stored field | Stored default | Toggle shows |
|---|---|---|---|
| Enable Ask AI | `kots_app.ask_ai_enabled` | `0` | Disabled |
| Display only fixable CVEs in Security Center report | `enterprise_portal_security_settings.show_unfixable_cves` | `0` | **Enabled** |
| Enable raw CVE scan to be downloadable | `enterprise_portal_security_settings.allow_raw_scan_download` | `0` | Disabled |

The middle row inverts because the toggle is rendered as `checked={!securitySettings.showUnfixableCVEs}` (`PortalFeatureSettings.tsx:199`, label at `:191`). With `show_unfixable_cves` defaulting to false — in both the table schema and `DefaultSecuritySettings()` in `pkg/enterprise-portal/security_settings.go`, which also covers apps with no settings row — the toggle is **on** out of the box, so customers see only fixable CVEs until a vendor turns it off. The page says Enabled for that reason.

I had this backwards in an earlier revision of this PR. #4373 states a default only for the raw scan download, so this page was the only place asserting the other two and the error would have shipped here.

**A note for future edits:** this default is a derived claim, not a value read off a column. If anyone ever normalizes the toggle's polarity so the label matches the stored field, the documented default flips with no schema change to signal it.

## What the page documents

**All three settings**, with the conditions that hide some of them. The two Security Center settings render only for teams with Security Center enabled (`PortalFeatureSettings.tsx:54-57`), so a vendor without it sees one toggle and would otherwise conclude the docs are wrong. The Security Center settings also use a separate RBAC resource, so a team member can be able to change Ask AI and not those.

**Ask AI in full**, since it has no other home:
- Default off (`kots-app.yaml`, `ask_ai_enabled` default `"0"`), app-level, applies to all customers at once
- Header-button-and-right-panel delivery, with no supported way to move, rename, or hide it (sc-139279 is deferred)
- Answer scope, stated precisely. The assistant answers product questions only from the asking customer's entitlement-filtered content, and it *may* use general public knowledge about Kubernetes, Helm, Linux, Docker, networking, TLS, and DNS while refusing outside knowledge about the application (`ask_ai.go:350`). "Only your docs" is the natural summary and it is not quite right, so the page says both halves.
- The data sent to the provider: the question, up to the last 10 messages of the conversation, and the entitlement-filtered content. Not only the question.

## Verified against source

All claims were checked against `replicatedhq/vandoor` at `a56ae15d2`, not against UI copy:

- History cap is **10 messages**, not 10 turns — `historyLimit := 10` (`ask_ai.go:103`) counts user and assistant messages together, so roughly five exchanges. The page says messages.
- Provider and model come from `ask_ai.go:499` and the request target at `ask_ai.go:513`. Withheld from the page per the section above, recorded here so the follow-up does not have to re-derive them.
- The Vendor Portal's own helper text says the assistant is "trained on your documentation". It is retrieval over the content, not training, and the page describes it accurately rather than repeating that phrasing.

## Vale

Clean apart from four deliberate skips: three `Headings` sentence-case warnings, because **Portal Features** and **Ask AI** are UI proper nouns and every sibling EP v2 page H1 is title case and identically flagged; and one `PositionalLanguage` warning on the assistant panel's right-side placement, which is load-bearing description rather than navigation prose.

### DIFF
--- a/docs/vendor/enterprise-portal-v2-portal-features.mdx
+++ b/docs/vendor/enterprise-portal-v2-portal-features.mdx
@@ -12,9 +12,9 @@ The section contains the following settings:
 
 | Setting | Default | Description |
 |---|---|---|
-| **Enable Ask AI** | Disabled | Adds an AI assistant to the portal that answers customer questions from your portal content. See [Ask AI](#ask-ai) below. |
-| **Display only fixable CVEs in Security Center report** | Enabled | Limits the CVE report to vulnerabilities with an available fix. Disable it to let customers switch between all CVEs and fixable CVEs. See [Security Center display settings](#security-center-display-settings) below. |
-| **Enable raw CVE scan to be downloadable** | Disabled | Lets customers download the raw Grype scan JSON for a release. See [Security Center display settings](#security-center-display-settings) below. |
+| **Enable Ask AI** | Disabled | Adds an AI assistant to the portal that answers customer questions from your portal content. See [Ask AI](#ask-ai). |
+| **Display only fixable CVEs in Security Center report** | Enabled | Limits the CVE report to vulnerabilities with an available fix. Disable it to let customers switch between all CVEs and fixable CVEs. See [Security Center display settings](#security-center-display-settings). |
+| **Enable raw CVE scan to be downloadable** | Disabled | Lets customers download the raw Grype scan JSON for a release. See [Security Center display settings](#security-center-display-settings). |
 
 The two Security Center settings appear only when Security Center is enabled for your team. If **Portal Features** shows only **Enable Ask AI**, contact your Replicated account representative about Security Center.
 
@@ -22,9 +22,9 @@ Access to the settings also depends on your Vendor Portal permissions. The Secur
 
 ## Ask AI {#ask-ai}
 
-When you enable **Enable Ask AI**, the portal adds an **Ask AI** button to the header of every content page. The button opens an assistant panel on the right side of the page, where the customer can ask questions and receive streamed answers. The setting is disabled by default, and it applies to all of your customers at once.
+When you turn on **Enable Ask AI**, the portal adds an **Ask AI** button to the header of every content page. The button opens an assistant panel on the right side of the page, where the customer can ask questions and receive streamed answers. The setting is disabled by default, and it applies to all of your customers.
 
-The assistant is delivered in the page header only. It is not a navigation item, and there is currently no supported way to move it, rename it, or hide it from a subset of customers. The **Enable Ask AI** setting is the only control over whether customers see it.
+The assistant is delivered in the page header only. It is not a navigation item, and there is no supported way to move it, rename it, or hide it from a subset of customers. The **Enable Ask AI** setting is the only control over whether customers see it.
 
 ### What the assistant answers from
 
--- a/docs/vendor/enterprise-portal-v2-portal-features.mdx
+++ b/docs/vendor/enterprise-portal-v2-portal-features.mdx
@@ -4,7 +4,7 @@ This topic describes the app-level Enterprise Portal settings in the **Portal Fe
 
 ## About Portal Features
 
-**Portal Features** is a set of app-level settings that control optional capabilities in the Enterprise Portal. The settings apply to every customer of the application. They are separate from the per-customer settings on a customer's **Enterprise Portal access** tab, which control whether an individual customer has portal access at all.
+**Portal Features** is a set of app-level settings that control optional capabilities in the Enterprise Portal. The settings apply to every customer of the application. They are separate from the per-customer settings on a customer's **Enterprise Portal access** tab. Those control whether an individual customer has portal access at all.
 
 To open the settings, go to **Enterprise Portal > Content** in the Vendor Portal and find the **Portal Features** section. Changes are saved automatically.
 
@@ -32,17 +32,21 @@ Because the two groups use separate permissions, a team member can have access t
 
 When you turn on **Enable Ask AI**, the portal adds an **Ask AI** button to the header of every content page. The button opens an assistant panel on the right side of the page, where the customer can ask questions and receive streamed answers. The setting is disabled by default, and it applies to all of your customers.
 
-The assistant is delivered in the page header only. It is not a navigation item, and there is no supported way to move it, rename it, or hide it from a subset of customers. The **Enable Ask AI** setting is the only control over whether customers see it.
+The assistant appears only in the page header. It is not a navigation item, and there is no supported way to move it, rename it, or hide it from some customers. **Enable Ask AI** is the only control over whether customers see it.
 
 ### What the assistant answers from
 
-Each question is answered from the portal content the asking customer can see, for the version they are viewing, so two customers on different licenses can receive different answers to the same question. Content is excluded based on the `visible_when` conditions on your TOC entries and on page frontmatter, so the gating that you already configured for the portal also applies to the assistant.
+The assistant answers each question from the portal content the asking customer can see, for the version they are viewing. Two customers on different licenses can therefore get different answers to the same question.
 
-Separately, the request instructs the assistant not to mention entitlements, licenses, or feature flags, and not to bring up a fixed set of installation and recovery capabilities that the customer's license does not include. That instruction is a prompt to the model, not an enforced filter, so do not rely on it alone to keep a feature's existence confidential.
+The `visible_when` conditions on your `toc.yaml` entries and page frontmatter decide what the assistant receives. The gating you already configured for the portal applies to the assistant too.
+
+The request also instructs the assistant not to mention entitlements, licenses, or feature flags. It further tells the assistant not to bring up installation or recovery capabilities the customer's license excludes.
+
+Both are prompts to the model, not enforced filters. Do not rely on them alone to keep a feature's existence confidential.
 
 For questions about your application itself, the assistant uses only your portal content. If your content does not cover a topic, it says that it does not have information rather than inferring an answer. It does not draw on outside knowledge about your application, because published information about your application can be out of date or wrong. For dependencies and infrastructure — Kubernetes, the Helm CLI, Linux administration, Docker, cloud providers, networking, TLS, and DNS — it may also use general public knowledge.
 
-Long pages and large content sets are truncated to fit the request, so the assistant can miss a topic that your content does cover.
+Replicated truncates long pages and large content sets to fit the request. The assistant can therefore miss a topic that your content does cover.
 
 When an answer draws on a specific portal page, the assistant is instructed to cite that page as the source.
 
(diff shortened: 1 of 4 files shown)


AmberAlston (dismissed):

AmberAlston (commented):
Thanks for this. A few things before it lands, mostly readability plus one substantive dig into the RBAC section.

**On the AI provider (naming):** confirmed the provider is not named anywhere on the page, which is the right call given Groq isn't on the Infrastructure and Subprocessors list yet. Two small things: (1) the first commit message still says the disclosure "names Groq and openai/gpt-oss-120b" and "must not merge before Groq appears", which is now stale after you generalized it, worth a note so a reviewer skimming commits isn't confused; and (2) it's Groq (the inference provider), not Grok (xAI's model), just so it's right wherever it does eventually get named.

**On line 20 ("available to all teams"):** the whole Security Center block only renders when the `ep_security_center` flag is on (PortalFeatureSettings.tsx:70,84,96), which the migration defaulted on for all teams. As written it reads like a throwaway. Worth making it explicit that it's the gate, or dropping it.

Inline comments on the specific lines below. Happy to pair on the tightening pass if useful.

docs/vendor/enterprise-portal-v2-portal-features.mdx
```diff
@@ -0,0 +1,84 @@
+# Configure Portal Features
+
+This topic describes the app-level Enterprise Portal settings in the **Portal Features** section of the Vendor Portal, including the Ask AI assistant and the Security Center display settings.
+
+## About Portal Features
+
+**Portal Features** is a set of app-level settings that control optional capabilities in the new Enterprise Portal. The settings apply to every customer of the application. They are separate from the per-customer settings on a customer's **Enterprise Portal access** tab, which control whether an individual customer has portal access at all.
```
AmberAlston: Can we drop "new" here? Going into Beta let's call it plain **Enterprise Portal**:

> **Portal Features** is a set of app-level settings that control optional capabilities in the Enterprise Portal.

I'm fine keeping the "new" usages in the Security Center access section below, where they contrast with Enterprise Portal (Classic) and carry real meaning.

docs/vendor/enterprise-portal-v2-portal-features.mdx
```diff
@@ -0,0 +1,84 @@
...
+| **Enable raw CVE scan to be downloadable** | Disabled | Lets customers download the raw Grype scan JSON for a release. See [Security Center display settings](#security-center-display-settings). |
+
+The three Security Center settings are available to all teams.
+
+Access to these settings depends on your Vendor Portal permissions, and they are not all governed by the same one. **Enable Ask AI** and **Enable Security Center for all customers** use the application read and update permissions. The two Security Center display settings use a separate Security Center settings permission.
+
+Because the permissions are separate, a team member can be able to change **Enable Ask AI** and **Enable Security Center for all customers** while the display settings are unavailable to them. Without read permission, a setting is replaced by a message explaining that. Without update permission, it appears but cannot be changed. For more information, see [Configure RBAC policies](/vendor/team-management-rbac-configuring).
```
AmberAlston: This section is hard to follow ("not all governed by the same one", "a team member can be able to change"). I traced each toggle to its exact permission in vandoor (PortalFeatureSettings.tsx + pkg/policy/resources.go:777-782). Suggest replacing with a table so it's precise and self-sufficient:

> ### Permissions
>
> The **Portal Features** settings are not all controlled by the same permission:
>
> | Setting | Read permission | Update permission |
> |---|---|---|
> | Enable Ask AI, Enable Security Center for all customers | `kots/app/[:appid]/read` | `kots/app/[:appid]/update` |
> | Display only fixable CVEs, Enable raw CVE scan to be downloadable | `kots/app/[:appid]/enterprise-portal/security-settings/read` | `kots/app/[:appid]/enterprise-portal/security-settings/update` |
>
> Because the two groups use separate permissions, a team member can have access to one group and not the other. If you lack the read permission for a setting, the Vendor Portal replaces it with a message explaining that. If you lack the update permission, the setting appears but is read-only. For more information, see [Configure RBAC policies](/vendor/team-management-rbac-configuring).

Heads up: the `enterprise-portal/security-settings/read` and `/update` resources aren't in team-management-rbac-resource-names.md yet (that file has `access`, `branding`, `documentation`, `email-domain`, but not `security-settings`). So linking readers there for that permission is currently a dead end. Naming the strings here makes the page self-sufficient, and we should add the two entries to the RBAC reference in a follow-up.

docs/vendor/enterprise-portal-v2-portal-features.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+## Ask AI {#ask-ai}
+
+When you turn on **Enable Ask AI**, the portal adds an **Ask AI** button to the header of every content page. The button opens an assistant panel on the right side of the page, where the customer can ask questions and receive streamed answers. The setting is disabled by default, and it applies to all of your customers.
+
+The assistant is delivered in the page header only. It is not a navigation item, and there is no supported way to move it, rename it, or hide it from a subset of customers. The **Enable Ask AI** setting is the only control over whether customers see it.
+
+### What the assistant answers from
+
+Each question is answered from the portal content that the asking customer can see. The assistant receives content drawn from what that customer can see in the portal, for the version they are viewing, so two customers on different licenses can receive different answers to the same question. Content is excluded based on the `visible_when` conditions on your TOC entries and on page frontmatter, so the gating that you already configured for the portal also applies to the assistant.
```
AmberAlston: The first two sentences say the same thing twice ("the portal content that the asking customer can see" then "what that customer can see in the portal"). Tighten to one:

> Each question is answered only from the portal content the asking customer can see, for the version they are viewing, so two customers on different licenses can get different answers to the same question.

More broadly, the page leans on hedgy constructions throughout ("so that", "rather than", "does not by itself") that read as careful-legal rather than doc prose. The accuracy is valid so I wouldn't cut a single claim, but a light word-tightening pass would help readability a lot.

docs/vendor/enterprise-portal-v2-portal-features.mdx
```diff
@@ -0,0 +1,84 @@
...
+- **The capabilities the customer's license excludes.** The request names a fixed set of installation and recovery capabilities that the customer is not entitled to, so that the assistant does not offer them.
+
+Because enabling Ask AI sends your portal content and your customers' questions to a third-party provider, review the setting against your own agreements with your customers before you turn it on.
+
+### Disabling Ask AI
+
+Disable **Enable Ask AI** to remove the button from the portal header for all customers. Disabling the setting stops new requests. It does not affect conversations that have already been sent.
+
+## Security Center display settings {#security-center-display-settings}
```
AmberAlston: Ordering: the settings table at the top matches the UI toggle order (Ask AI, Enable Security Center for all customers, Display only fixable CVEs, Enable raw CVE scan), but the prose sections below don't. In the Vendor Portal the order is Ask AI, then Security Center access, then the two display settings (PortalFeatureSettings.tsx toggles at :193, :242, :294, :319). Here "Security Center display settings" comes before "Security Center access". Can we swap these two H2 sections so the section order matches the toggle order?

docs/vendor/enterprise-portal-v2-portal-features.mdx
AmberAlston: Side topic from this PR, i find this permission difference a bit odd. Something to follow up up

AmberAlston (approved):

PR #4418, opened 2026-08-24
Document safe credential handling in troubleshoot specs

banjoh:
## Summary

The docs had no example of `runPod` with `env`, let alone `env.valueFrom.secretKeyRef` — currently the only non-literal way to get a credential into a collector on KOTS.

Story: [sc-139336](https://app.shortcut.com/replicated/story/139336)

## Preview

- [Pass a credential to the pod](https://deploy-preview-4418--replicated-docs.netlify.app/vendor/support-bundle-examples#run-pod-credentials) — new subsection and example
- [Use a credential in a preflight check](https://deploy-preview-4418--replicated-docs.netlify.app/vendor/preflight-defining#credentials) — new section and example
- [Collect Redis and MySQL server information](https://deploy-preview-4418--replicated-docs.netlify.app/vendor/support-bundle-examples#collect-redis-and-mysql-server-information) — warning added, credentials replaced with placeholders
- [Add collectors and analyzers](https://deploy-preview-4418--replicated-docs.netlify.app/vendor/support-bundle-customizing#add-collectors-and-analyzers) — warning added
- [Check MySQL version using template functions](https://deploy-preview-4418--replicated-docs.netlify.app/vendor/preflight-examples#check-mysql-version-using-template-functions) — warning added
- [Helm template example](https://deploy-preview-4418--replicated-docs.netlify.app/vendor/preflight-sb-helm-templates-about#helm-template-example) — warning added
- [Example: Support bundle specifications by component](https://deploy-preview-4418--replicated-docs.netlify.app/vendor/support-modular-support-bundle-specs#component) — warning added, credential replaced with placeholder

## Changes

- **New examples** of `runPod` with `env.valueFrom.secretKeyRef`, for support bundles (`support-bundle-examples.mdx`) and preflights (`preflight-defining.mdx#credentials`).
- **Placeholders instead of real-looking credentials** in the Redis/MySQL and modular spec examples: `my-secret-pw`, `default:replicated`, `default:password` → `<PASSWORD>`.
- **Warnings where there were none**: a credential written into a collector `uri` ships in the release and sits at rest in the cluster, and Helm template functions and KOTS config options do not prevent that — they resolve to the live credential when the spec is rendered. The docs previously stated that a template function avoided exposing the credential; that sentence was removed in May 2026 and nothing replaced it.
- **New guidance on using a credential in a preflight check**: recommends a pre-existing Secret, since a Secret created by the application's own chart is not available when preflight checks run — including when annotated `helm.sh/hook: pre-install`. On Embedded Cluster v3, an extension chart can create the Secret, because `extensions.helmCharts[].values` supports Replicated template functions and a config screen password can be templated in. Embedded Cluster v2 has no template function support in its Config, so the recommendation there is a support bundle after the install completes.

## Not covered

The redactor reference still shows `values:` redaction without noting that the redactor spec is itself plain text, and `preflight-support-bundle-about` still lists passwords and database connection strings as automatically redacted without qualification. Both need the actual `redact.go` coverage confirmed before the claims are rewritten.

The underlying gap is that `troubleshoot.sh/v1beta2` has no non-literal credential form for the database collectors, which is a product issue rather than a docs one.

## Testing

`npm run build` passes, with no broken links or anchors on any changed page.


seanoseanohay:
docs/vendor/support-bundle-examples.mdx line 113:

"This pattern does not apply to preflight checks, which run before the application is deployed. For more information, see Define preflight checks (/vendor/preflight-defining)."

docs/vendor/preflight-defining.mdx line 77:

"To check a credentialed resource without exposing the credential, use a runPod collector and pass the credential to the Pod with env.valueFrom.secretKeyRef. The kubelet resolves the credential when the Pod starts, so the credential never appears in the spec."

To me one says the secretKeyRef pattern does not apply to preflights. The other is a how-to for using that same pattern in a preflight. Those two can’t both be true.

seanoseanohay:
> docs/vendor/support-bundle-examples.mdx line 113:
> 
> "This pattern does not apply to preflight checks, which run before the application is deployed. For more information, see Define preflight checks (/vendor/preflight-defining)."
> 
> docs/vendor/preflight-defining.mdx line 77:
> 
> "To check a credentialed resource without exposing the credential, use a runPod collector and pass the credential to the Pod with env.valueFrom.secretKeyRef. The kubelet resolves the credential when the Pod starts, so the credential never appears in the spec."
> 
> To me one says the secretKeyRef pattern does not apply to preflights. The other is a how-to for using that same pattern in a preflight. Those two can’t both be true.

Discussed offline and fixed.

xavpaice (approved):

PR #4442, opened 2026-08-27
docs(release-notes): clarify EC SELinux enforcing support across 2.1.0 and 2.8.0

AmberAlston:
Preview links: 
https://deploy-preview-4442--replicated-docs.netlify.app/release-notes/rn-embedded-cluster#280
https://deploy-preview-4442--replicated-docs.netlify.app/release-notes/rn-embedded-cluster#280

## Summary

Clarifies the Embedded Cluster release notes around SELinux enforcing mode. A field question surfaced that the release notes are easy to misread: the 2.1.0 entry says a preflight was added requiring SELinux **not** be in enforcing mode, and the 2.8.0 entry says SELinux environments are supported but does not explicitly say enforcing mode or that it supersedes the 2.1.0 restriction. A reader landing on 2.1.0 can reasonably conclude enforcing is still unsupported.

These are additive annotations to historical release notes, not rewrites of what shipped.

## Changes

`docs/release-notes/rn-embedded-cluster.md`:
- **2.8.0 improvements**: state explicitly that support includes SELinux in enforcing mode, and note that it removes the 2.1.0 preflight that previously required SELinux not be in enforcing mode.
- **2.1.0 improvements**: add a `:::note` after the SELinux preflight bullet pointing forward to 2.8.0, so a reader does not take the pre-2.8.0 behavior as current.

### DIFF
--- a/docs/release-notes/rn-embedded-cluster.md
+++ b/docs/release-notes/rn-embedded-cluster.md
@@ -820,7 +820,7 @@ Released on August 1, 2025
 
 ### Improvements {#improvements-2-8-0}
 * Adds host preflight checks to ensure nodes have sufficient storage to perform airgap installations when installing a new node and when joining nodes to an existing cluster.
-* Enables Embedded Cluster to install in SELinux environments by setting appropriate SELinux file contexts on the bin directory and restoring SELinux contexts for the data directory after creation.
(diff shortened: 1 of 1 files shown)


AmberAlston:
fyi @kriscoleman

PR #4444, opened 2026-08-27
Registry is accessed through proxy by default

divolgin:


AmberAlston:
Thanks for getting this started. Two things before it merges.

**1. The "older accounts" note needs specifics.** Right now it says the hostname "can be different for older accounts," which leaves people guessing. Let's say what differs, why, and how to change it. It's also misplaced inside the `--password` warning note, since it applies to every command on the page, not just the advanced login. I'd pull it out into its own note:

> :::note
> Newer accounts log in to `proxy.replicated.com` and pull from `proxy.replicated.com/registry/...`, routing all image and chart pulls through a single registry. Older accounts may instead see `registry.replicated.com` in the **Helm install instructions** dialog. Both work, so run the exact commands shown in your dialog. To switch an existing account to the single-registry commands, reach out to your Replicated account representative.
> :::

**2. The firewall/egress section still needs updating, and that's the bigger piece.** This PR swaps the command hostnames but leaves [the firewall section](https://docs.replicated.com/vendor/install-with-helm#firewall) listing both `registry.replicated.com` and `proxy.replicated.com`. The whole value of this change is that a proxy-only account only has to allowlist one registry, so that section is where the payoff shows up. This needs its own edit, not the command note copied over, because the *list of domains* changes and readers need a way to discover how to get onto the single-registry model. Concretely, I'd make it something like:

> | Domain | Description |
> |---|---|
> | `proxy.replicated.com` | Private images and Helm charts are pulled through the proxy registry. |
> | `registry.replicated.com` | Only required for accounts not yet on the single-registry model (see note below). |
> | `replicated.app` | Required if the Replicated SDK is included. Upstream application metadata is pulled from here. |
>
> :::note
> Newer accounts route all image and chart pulls through `proxy.replicated.com` and do not need to allow `registry.replicated.com`. Older accounts must allow both. To move your account to the single-registry setup and drop `registry.replicated.com` from your allowlist, reach out to your Replicated account representative.
> :::

docs/partials/install/_firewall-openings-helm.mdx
```diff
@@ -5,16 +5,20 @@
   </tr>
   <tr>
       <td>`replicated.app` &#42;</td>
-      <td><p>Upstream application YAML and metadata is pulled from `replicated.app`. The current running version of the application (if any), as well as a license ID and application ID to authenticate, are all sent to `replicated.app`. This domain is owned by Replicated, Inc., which is headquartered in Los Angeles, CA.</p><p>For the range of IP addresses for `replicated.app`, see [replicatedhq/ips](https://github.com/replicatedhq/ips/blob/main/ip_addresses.json#L60-L65) in GitHub.</p></td>
```
divolgin, line 8: restore "This domain is owned by Replicated, Inc"... etc

AmberAlston (approved):
Both items from my earlier comment are addressed (dedicated older-accounts note with the switch path, and the firewall section now conditional on the single-registry model with discoverability), and the domain ownership lines are restored. LGTM.

PR #4447, opened 2026-08-28
docs(ep-v2): document hidden: true page frontmatter

seanoseanohay:
## Summary

Document Enterprise Portal v2 page frontmatter `hidden: true` so vendors can keep a page in the content repo and `toc.yaml` without publishing it.

Implements docs for [sc-138380](https://app.shortcut.com/replicated/story/138380). Implementation: [vandoor#10449](https://github.com/replicatedhq/vandoor/pull/10449).

## Why

Vendors could omit a page from `toc.yaml` or gate it with `visible_when`, but there was no documented way to leave the nav slot in place and keep the page unpublished. `hidden: true` is unpublished content, not a license or entitlement gate. It takes precedence over `visible_when`.

## Changes

- `enterprise-portal-v2-content.mdx`
  - Distinguish omitting a page from `toc.yaml` (allowlist) from `hidden: true` (keep the TOC entry, unpublished)
  - Note that `overrides.home` cannot restore a hidden page, and `hidden: true` can hide the home landing page
  - Nested nav: a TOC node whose `page` is hidden is dropped with its subtree; title-only sections cannot use `hidden`
  - Add `hidden: true` next to frontmatter `visible_when`, with a page example and a matching `toc.yaml` entry
  - Asset access: a hidden page does not authorize downloads
- `enterprise-portal-v2-troubleshooting.mdx` — pages missing from navigation can be unpublished with `hidden: true`
- `enterprise-portal-v2-use.mdx` — customer-facing nav also respects unpublished pages

No changes to `enterprise-portal-v2-versioned-docs.mdx`. Hiding `main` as a draft branch is a different feature.

## Validation

- `git diff --check`
- Did not run Vale or a Docusaurus build locally

### DIFF
--- a/docs/vendor/enterprise-portal-v2-content.mdx
+++ b/docs/vendor/enterprise-portal-v2-content.mdx
@@ -105,10 +105,10 @@ The `toc.yaml` file at the root of your repo defines the sidebar navigation. Eac
 
 Every item also supports `icon` and `visible_when` (see [Visibility](#visibility) below).
 
-The `toc.yaml` acts as an allowlist: pages that exist in your repo but are not listed in `toc.yaml` are not reachable through the portal navigation or direct URL. This means you can hide a page by simply omitting it from `toc.yaml`.
+The `toc.yaml` acts as an allowlist: pages that exist in your repo but are not listed in `toc.yaml` are not reachable through the portal navigation or direct URL. Omitting a page from `toc.yaml` hides it entirely. To keep the page in the repo and listed in the TOC without publishing it, set `hidden: true` in the page frontmatter. See [Hide unpublished pages](#hide-unpublished-pages).
 
 :::note
-The `overrides.home` page bypasses the TOC allowlist and remains accessible even if it is not listed in `toc.yaml`. If you need to hide the home page, use `visible_when` in its frontmatter instead.
+The `overrides.home` page bypasses the TOC allowlist and remains accessible even if it is not listed in `toc.yaml`. To hide the home landing page, set `hidden: true` or `visible_when` in its frontmatter. `overrides.home` cannot restore a hidden page or a child of a hidden parent.
 :::
 
 Supported icon values: `rocket`, `download`, `cloud`, `settings`, `wrench`, `life-buoy`, `file-text`, `star`, `book`, `shield`, `package`, `refresh-cw`, `database`, `key`. If omitted, defaults to `book`.
@@ -182,6 +182,8 @@ navigation:
 In this example, **Configuration** expands to show **Required Values** and **Optional Values**. **Required Values** further expands to show **Authentication** and **Networking**. Each level is collapsible in the sidebar.
 
 Any item with `items` can also have its own `page`, `visible_when`, and `icon`. If all children of a parent are hidden by `visible_when`, the parent is also hidden automatically.
+
+If a TOC node has a `page` with `hidden: true`, the portal drops that node and its subtree, including nested `helm_chart` and `terraform_module` items. A section that is only a title (no `page`) cannot use `hidden`. For more information, see [Hide unpublished pages](#hide-unpublished-pages).
 
 ### External links {#external-links}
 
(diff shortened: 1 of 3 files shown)


AmberAlston (dismissed):

docs/vendor/enterprise-portal-v2-troubleshooting.mdx
```diff
@@ -38,6 +38,8 @@ See [Manage Content Versions](/vendor/enterprise-portal-v2-versioned-docs) for h
 - Ensure file path matches `toc.yaml` reference
 - Verify visibility rules (entitlements, channels)
 - Check frontmatter `visible_when` conditions
+- Confirm the page frontmatter does not set `hidden: true`. Hidden pages stay unpublished for every customer, including in local preview
+- Confirm the page frontmatter parses as valid YAML. Frontmatter that Enterprise Portal cannot parse is treated as unpublished, so the page is hidden from every customer, including in local preview
```
seanoseanohay: This is a bit broader than the implementation. Invalid YAML fail-closes as unpublished ([frontmatter.go:L58-L59](https://github.com/replicatedhq/vandoor/blob/sc-138380-hidden-frontmatter/pkg/enterprise-portal/parse/frontmatter.go#L58-L59)), but valid YAML with a type error (like `tags: draft`) does not hide the page unless `hidden: true` is set ([frontmatter_test.go:L23](https://github.com/replicatedhq/vandoor/blob/sc-138380-hidden-frontmatter/pkg/enterprise-portal/parse/frontmatter_test.go#L23)). Can we say "invalid YAML is treated as unpublished" instead of "frontmatter that Enterprise Portal cannot parse"?

docs/vendor/enterprise-portal-v2-content.mdx
```diff
@@ -720,6 +722,46 @@ The same `entitlements` conditions available in `toc.yaml` work in frontmatter,
...
 
+#### Hide unpublished pages {#hide-unpublished-pages}
+
+`hidden` controls whether a page is published, not who can see it. Use `visible_when` to show a page to some customers and not others based on their license. Use `hidden: true` to keep a page out of every customer's portal until you are ready to publish it, regardless of license.
+
+Set `hidden: true` in a page's YAML frontmatter when the page belongs in your content repo but is not ready for customers yet. For example, stage a product announcement, a new install guide, or a revised section next to your live docs, keep its entry in `toc.yaml` so local preview shows where it sits in the navigation, then publish it by removing the flag when you are ready. Draft pages stay under version control in the same repo and pull request flow as your published content, so nothing lives in a stray branch or an unwired file you have to remember later.
```
seanoseanohay: This says local preview will show the unpublished page in the nav, but [enterprise-portal-v2-content.mdx:L762](https://github.com/replicatedhq/replicated-docs/blob/sc-138380-ep-hidden-frontmatter-docs/docs/vendor/enterprise-portal-v2-content.mdx#L762) says the opposite. Preview prunes `hidden: true` the same way prod does ([run.go:L195-L199](https://github.com/replicatedhq/vandoor/blob/sc-138380-hidden-frontmatter/pkg/daemons/ep_preview_api/run.go#L195-L199)), so it won't show in the preview sidebar either.

Can we keep "leave it in `toc.yaml`" but change the reason? The TOC entry is so the nav slot is already wired when you unhide it, not so preview can display it. If we say that here, we can drop the duplicate sentence in the note.

docs/vendor/enterprise-portal-v2-troubleshooting.mdx
AmberAlston: Good catch, corrected in 9d8daeb6e. Reworded to "Invalid YAML is treated as unpublished" so it no longer implies a typed-parse failure like `tags: draft` hides the page.

docs/vendor/enterprise-portal-v2-content.mdx
AmberAlston: Fixed in 9d8daeb6e. Changed the reason to nav-slot pre-wiring, added a line that preview hides the page like production, and dropped the now-duplicate preview sentence from the note.

AmberAlston (approved):

PR #4486, opened 2026-09-08
docs: clarify EC v2 pod log storage

sgalsaleh:
## Summary
- document the default pod-log location for new Embedded Cluster v2 installations
- explain that pod-log usage counts toward the required data-directory space and kubelet disk-pressure accounting
- remove the worker-profile workaround that told vendors to set podLogsDir from the install data directory

## Note
This documentation relies on the product behavior in replicatedhq/embedded-cluster#3936. Merge this PR after that fix is released. Vendors only need to set podLogsDir when they deliberately want a different location.

## Validation
- git diff --check
- docs build was not run because dependencies are not installed locally

### DIFF
--- a/docs/partials/embedded-cluster/v2/_requirements.mdx
+++ b/docs/partials/embedded-cluster/v2/_requirements.mdx
@@ -14,7 +14,7 @@
 
 * The data directory used by Embedded Cluster must have 40Gi or more of total space and be less than 80% full. By default, the data directory is `/var/lib/embedded-cluster`. The directory can be changed by passing the `--data-dir` flag with the Embedded Cluster `install` command. For more information, see [install](/embedded-cluster/v2/embedded-cluster-install).
 
-   For new installations that use Kubernetes 1.30 or later, Embedded Cluster stores pod logs in `<data-dir>/k0s/pod-logs`.
+   For new installations created with Embedded Cluster 2.19.8 or later, Embedded Cluster stores pod logs in `<data-dir>/k0s/pod-logs`. Pod log volume counts toward the required space in the data directory. Storing pod logs in this directory lets kubelet account for their disk usage when it manages disk pressure and pod eviction.
 
    Note that in addition to the primary data directory, Embedded Cluster creates directories and files in the following locations:
 
@@ -36,6 +36,7 @@
       - `/var/log/calico`
       - `/var/log/containers`
       - `/var/log/embedded-cluster`
+      - `/var/log/pods` (installations created before Embedded Cluster 2.19.8)
       - `/usr/local/bin/k0s`
 
 * (Online installations only) Access to replicated.app and proxy.replicated.com or your custom domain for each
--- a/docs/partials/embedded-cluster/v2/_requirements.mdx
+++ b/docs/partials/embedded-cluster/v2/_requirements.mdx
@@ -14,7 +14,7 @@
 
 * The data directory used by Embedded Cluster must have 40Gi or more of total space and be less than 80% full. By default, the data directory is `/var/lib/embedded-cluster`. The directory can be changed by passing the `--data-dir` flag with the Embedded Cluster `install` command. For more information, see [install](/embedded-cluster/v2/embedded-cluster-install).
 
-   For new installations created with Embedded Cluster 2.19.8 or later, Embedded Cluster stores pod logs in `<data-dir>/k0s/pod-logs`. Pod log volume counts toward the required space in the data directory. Storing pod logs in this directory lets kubelet account for their disk usage when it manages disk pressure and pod eviction.
+   By default, new Embedded Cluster installations store pod logs in `<data-dir>/k0s/pod-logs`. Pod log volume counts toward the required space in the data directory. Storing pod logs in this directory lets kubelet account for their disk usage when it manages disk pressure and pod eviction.
 
    Note that in addition to the primary data directory, Embedded Cluster creates directories and files in the following locations:
 
--- a/docs/partials/embedded-cluster/v2/_requirements.mdx
+++ b/docs/partials/embedded-cluster/v2/_requirements.mdx
@@ -14,6 +14,8 @@
 
 * The data directory used by Embedded Cluster must have 40Gi or more of total space and be less than 80% full. By default, the data directory is `/var/lib/embedded-cluster`. The directory can be changed by passing the `--data-dir` flag with the Embedded Cluster `install` command. For more information, see [install](/embedded-cluster/v2/embedded-cluster-install).
 
+   For new installations that use Kubernetes 1.30 or later, Embedded Cluster stores pod logs in `<data-dir>/k0s/pod-logs`.
+
    Note that in addition to the primary data directory, Embedded Cluster creates directories and files in the following locations:
 
       - `/etc/cni`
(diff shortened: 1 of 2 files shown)


docs/partials/embedded-cluster/v2/_requirements.mdx
```diff
@@ -34,7 +36,6 @@
       - `/var/log/calico`
       - `/var/log/containers`
       - `/var/log/embedded-cluster`
-      - `/var/log/pods`
       - `/usr/local/bin/k0s`
```
AmberAlston, line 40: The sentence above is scoped to new installations, but the deletion is unconditional. Any cluster installed before 2.19.8 keeps writing here, since the k0s config is written once and this page says worker profiles cannot be changed or added on upgrade. Could we qualify it rather than drop it?

```suggestion
      - `/var/log/pods` (installations created before Embedded Cluster 2.19.8)
      - `/usr/local/bin/k0s`
```

AmberAlston (commented):
Good to have the `workerProfiles` replacement behavior written down, that one is easy to get bitten by. Four comments inline.

One thing that applies to both files. This shipped in 2.19.8, which only exists as `+k8s-1.34`, `+k8s-1.35`, and `+k8s-1.36`, so "that use Kubernetes 1.30 or later" cannot be false for anyone reading it. I would drop it from both sentences and let the version qualifier do the work on the `/var/log/pods` entry instead. Shout if I have the version wrong.

Also page-level. The current embedded-config page carries a general warning that overrides replace whole sections rather than deep-merge, using `workerProfiles` as its example, and the 2.0.0 page has none. This PR documents one instance of that hazard, so porting the general warning back might be worth doing at the same time.

docs/partials/embedded-cluster/v2/_requirements.mdx
```diff
@@ -14,6 +14,8 @@
 
 * The data directory used by Embedded Cluster must have 40Gi or more of total space and be less than 80% full. By default, the data directory is `/var/lib/embedded-cluster`. The directory can be changed by passing the `--data-dir` flag with the Embedded Cluster `install` command. For more information, see [install](/embedded-cluster/v2/embedded-cluster-install).
 
+   For new installations that use Kubernetes 1.30 or later, Embedded Cluster stores pod logs in `<data-dir>/k0s/pod-logs`.
```
AmberAlston: Could we add what this costs? Pod logs now live inside the data directory, and the bullet above requires 40Gi or more and under 80% full, so log volume counts against a budget it did not touch before. My understanding is that this was the point, since kubelet then measures disk pressure on the same filesystem and eviction accounts for pod logs. Both matter to someone provisioning a host, and this page is where they would look for it.

embedded-cluster_versioned_docs/version-2.0.0/embedded-config.mdx
```diff
@@ -298,6 +298,8 @@ You can configure the Kubelet to customize your worker nodes with Embedded Clust
 
 You can customize the Kubelet configuration settings by adding a _worker profile_ in the Embedded Cluster Config under `unsupportedOverrides.k0s.config.spec.workerProfiles[]`. When a worker profile is defined, Embedded Cluster uses the profile for every node in the cluster during initial installation and when joining nodes to the cluster.
 
+On new installations that use Kubernetes 1.30 or later, Embedded Cluster internally configures the `default` worker profile to store pod logs in `<data-dir>/k0s/pod-logs`. Defining `workerProfiles` replaces this internal profile and its pod log directory configuration. To use a different worker profile while storing pod logs in the data directory, set `podLogsDir` in that profile's `values`.
```
AmberAlston: "set `podLogsDir` in that profile's `values`" does not say what to set it to, so the reader has to work out `<data-dir>/k0s/pod-logs` and resolve `<data-dir>` themselves. Worth stating the value here, or adding it to the example further down so it can be copied.

embedded-cluster_versioned_docs/version-2.0.0/embedded-config.mdx
```diff
@@ -298,6 +298,8 @@ You can configure the Kubelet to customize your worker nodes with Embedded Clust
...
+
 The worker profile has the following required fields:
 * `name`: The name of the worker profile. Do not use the name `default` for your custom worker profile. `default` is reserved by k0s.
```
AmberAlston, line 302: This sits three lines under a paragraph saying Embedded Cluster configures the `default` worker profile, so the two read as a contradiction. Could the new paragraph pick this up, something like noting that we occupy `default` and a custom profile replaces it? That explains the reservation and the replacement behavior in one place.

embedded-cluster_versioned_docs/version-2.0.0/embedded-config.mdx
```diff
@@ -298,6 +298,8 @@ You can configure the Kubelet to customize your worker nodes with Embedded Clust
 
 You can customize the Kubelet configuration settings by adding a _worker profile_ in the Embedded Cluster Config under `unsupportedOverrides.k0s.config.spec.workerProfiles[]`. When a worker profile is defined, Embedded Cluster uses the profile for every node in the cluster during initial installation and when joining nodes to the cluster.
 
+On new installations created with Embedded Cluster 2.19.8 or later, Embedded Cluster internally configures k0s's reserved `default` worker profile to store pod logs in `<data-dir>/k0s/pod-logs`. Custom worker profiles must not use `default`; defining `workerProfiles` replaces this internal profile and its pod log directory configuration. To use a different worker profile while storing pod logs in the data directory, set `podLogsDir: <data-dir>/k0s/pod-logs` in that profile's `values`.
```
AmberAlston: The last sentence asks for something a vendor cannot write correctly. `--data-dir` is chosen by the end user at install time and only defaults to `/var/lib/embedded-cluster`, and this page's own Limitations say the Embedded Cluster Config does not support Go template functions. So `<data-dir>` can only be hardcoded, and any install that passes `--data-dir` gets a `podLogsDir` pointing at a different filesystem than the one the paragraph above says kubelet measures. That fails silently, and in the direction the change was meant to prevent.

Two ways out, your call. Either say the literal path only holds when end users accept the default data directory, or drop that sentence and state the consequence instead, that defining `workerProfiles` returns pod logs to the k0s default location. I lean toward the second, since the first documents a footgun as though it were a supported pattern.

AmberAlston (commented):
Thanks, the version framing, the `/var/log/pods` qualifier, the sizing and eviction sentence, and the `default` reservation all read well now. One new thing in the fix to my third comment, inline.

Not a docs issue, but worth knowing while you decide the wording. The ordering in `setPodLogsDir` is deliberate, the comment says "Set before overrides so a vendor- or end user-supplied worker profile still wins," so a vendor who sets `maxPods` currently has no supported way to keep the pod log placement as well. That may be worth a ticket rather than prose here.

embedded-cluster_versioned_docs/version-2.0.0/embedded-config.mdx
sgalsaleh: yeah, good catch. i agree the second option is better

AmberAlston: Reading this is making me wonder if we have a gap. A vendor writes the EC Config at release time, the end user picks --data-dir at install time, and the Config has no templating, so the literal is only correct when nobody overrides the default. When it is wrong, pod logs land outside the data directory and lose the eviction accounting the paragraph above promises. And a vendor who wants maxPods and the pod log placement has no supported way to get both, since their profile deliberately replaces EC's.

I do not think this needs to block the PR. Landing it with narrower wording and filing the gap separately seems right, but tell me if you read it differently.

AmberAlston (approved):

PR #4499, opened 2026-09-10
docs: document remove-node then reset flow for EC v3 node removal

banjoh:
The reset command no longer deletes the node from the cluster. Node removal is now done with remove-node run
from a controller. Update the manage-nodes procedures to match.

Cleanups:
- Make `reset` an optional follow-up to `remove-node`, for reusing the machine
- Attribute multi-node cluster deletion to both commands, not `reset` alone
- Move the etcd safety-check note from the worker step to the controller step
- Condense "Remove a multi-node cluster" to a paragraph that links to the node procedures
- Say reset after `remove-node` removes only Embedded Cluster, since the app is already drained
- Move "Limitations and best practices" after the procedures
- End cluster removal with an optional reset of each machine for reuse

Preview (read entire section): https://deploy-preview-4499--replicated-docs-upgrade.netlify.app/embedded-cluster/v3/embedded-manage-nodes

### DIFF
--- a/embedded-cluster/embedded-manage-nodes.mdx
+++ b/embedded-cluster/embedded-manage-nodes.mdx
@@ -163,7 +163,7 @@ To remove a node from a multi-node cluster, run the `remove-node` command from a
 Run the `remove-node` command from a controller node other than the node that you want to remove. The command performs the following steps:
 
 1. Drain workloads from the node so that they reschedule onto other nodes in the cluster
-1. For controller nodes, remove the node from the etcd cluster and delete the autopilot ControlNode object
+1. For controller nodes, remove the node from the etcd cluster
 1. Delete the node from Kubernetes
 
 The `reset` command removes Embedded Cluster and your application from the machine where it is run. This is useful for iteration, development, and when you make mistakes because you can reuse the machine instead of having to procure a new one.
@@ -195,7 +195,7 @@ To remove a node from a multi-node cluster:
     * `NODE_NAME` is the Kubernetes node name of the node to remove. To list the node names in the cluster, run `sudo ./APP_SLUG shell -c "kubectl get nodes"`.
 
     :::note
-    If the node is unreachable, include the `--skip-drain` flag to skip draining workloads from the node. If the node's Kubernetes object is already gone, `remove-node` skips directly to the etcd and ControlNode cleanup.
+    If the node is unreachable, include the `--skip-drain` flag to skip draining workloads from the node. If the node's Kubernetes object is already gone, `remove-node` skips directly to the etcd cleanup.
     :::
 
 1. (Optional) Reset the machine that you removed so that you can reuse it. See [Reset a machine](#reset-a-node) in this document.
--- a/embedded-cluster/embedded-manage-nodes.mdx
+++ b/embedded-cluster/embedded-manage-nodes.mdx
@@ -122,7 +122,7 @@ To create a multi-node cluster with HA:
 
 * During installation with Embedded Cluster, follow the steps in the Embedded Cluster UI to join a total of three `controller` nodes to the cluster. For more information about joining nodes, see [Join nodes](#add-nodes) on this page.
 
-   Embedded Cluster automatically converts the installation to HA when three or more controller nodes are present. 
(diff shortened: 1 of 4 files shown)


AmberAlston:
Nice cleanup, and the direction matches the code. I checked the claims against `reset.go` and `removenode.go` and the substance holds up. Two things that could look like regressions are actually corrections: `reset` never rebooted, it only prints that a reboot is recommended, so making that an explicit step is right. And the "reset won't remove a controller while workers exist" note described a check that doesn't exist in v3, where the only safety check is the etcd membership one, so dropping it is also right.

I'd like all of the below in this PR rather than follow-ups, since we're establishing the `remove-node` flow and I don't want half of it landing.

**On your changes**

The `remove-node` step list is missing the autopilot ControlNode deletion for controllers. Minor, but it's a real step in the help text.

There's no path for a node that's already unreachable, and I think that's the bigger gap. Removing a dead node is a main-line case, `remove-node` has `--skip-drain` specifically for it, and the command's own help calls it out with a dedicated example. As written, someone whose node is down follows the procedure and hits a drain that fails or hangs.

SeaweedFS is missing. The `remove-node` help says to follow the volume rebalancing procedures in the operator runbook after removal. For multi-node with object storage that's a data safety step, and Limitations and best practices seems like the right home for it.

Two small ones. "Reboot the machine" is written as required while the CLI says recommended, which is fine to be prescriptive about but worth knowing. And the section is now "Remove nodes and reset machines" with body text about resetting machines, while the heading below is still "Reset a node."

**Pre-existing on the reference pages, but needs fixing as part of enabling this flow**

None of this is from your change, but this PR routes people to `remove-node`, so it's about to send real traffic at a page with a dangerously wrong flag description.

On `embedded-cluster-remove-node.mdx`, `--force` is documented as "Force removal without additional confirmation." The flag is actually registered as "Force removal even if it breaks quorum (dangerous)", and `remove-node` has no confirmation prompt at all, so there is nothing to skip. Someone reading the current text would reach for it casually and could break etcd quorum. That one I'd treat as required.

Also on that page, `--skip-drain` says "Skip draining workloads from the node before removal" but omits when you'd use it, which is the useful half and ties back to the unreachable node gap above.

On `embedded-cluster-reset.mdx`, `--force` is described as "Ignore errors encountered when resetting the node." It also skips the destructive action confirmation prompt, which is the guardrail worth naming. Same wording is carried into the Limitations bullet on this page.

**Sequencing**

replicatedhq/ec#565 should land first. Until it does, worker reset still attempts the node delete, fails, and prints "Reset completed with warnings", which the new docs don't mention.

AmberAlston (changes requested):
See prior comment

banjoh:
Comments addressed. The only thing I left out is references to `ControlNode` and `autopilot`. EC v3 does not use k0s autopilot for cluster management unlike EC v2

embedded-cluster/embedded-manage-nodes.mdx
```diff
@@ -152,67 +152,87 @@ sudo ./APP_SLUG shell -c "kubectl get nodes"
...
+
+* If you need to remove one controller node in a three-node cluster, first join a fourth controller node to the cluster before removing the target node. This ensures that you maintain a minimum of three nodes for the Kubernetes control plane. You can add and remove worker nodes as needed because they do not have any control plane components.
+
+* When resetting a single node or deleting a test environment, you can include the `--force` flag with the `reset` command to skip the confirmation prompt and continue past any errors.
 
-1. After removing all the worker nodes in the cluster, SSH onto a controller node and run the `reset` command to remove the node.
+* When removing a multi-node cluster, remove the worker nodes before the controller nodes. For more information, see [Remove a multi-node cluster](#remove-a-multi-node-cluster) in this document.
 
-1. Repeat the previous step on the remaining controller nodes in the cluster.
+* If SeaweedFS is installed, follow the volume rebalancing procedures in the operator runbook after removing a node.
```
AmberAlston: This is the wording from the remove-node help text, but there's no operator runbook anywhere in the docs, so a customer following this has no clear next step

AmberAlston (approved):

PR #4500, opened 2026-09-10
Document EC v3 host configuration requirements and drop the private spec link

AmberAlston:
Preview links
* https://deploy-preview-4500--replicated-docs-upgrade.netlify.app/embedded-cluster/v3/embedded-overview#about-host-preflight-checks
* https://deploy-preview-4500--replicated-docs-upgrade.netlify.app/embedded-cluster/v3/installing-embedded-requirements#host-configuration

## What

Replaces the broken "full host preflight spec" link on the Embedded Cluster v3 overview with a pointer to the requirements page, and adds a **Host configuration** section to that page covering what the link was standing in for.

## Why

The overview linked to `replicatedhq/ec`, which is private, so every vendor following that link got a 404. It was also pinned to a commit SHA rather than a branch, so it was frozen wherever that commit landed. It is the only link into that repo anywhere in the docs.

The link existed because vendors preparing hosts were hitting blockers one at a time, and handing them the raw spec was a faster answer than writing the requirements down. That was a bandaid. There was also a customization rationale carried over from kURL, where vendors needed the YAML to know what to override, but that does not apply here: the same page states that vendors cannot modify the default host preflight checks or supply their own spec.

Publishing the raw spec would not be a good replacement either. It is 73 checks across three files, most of which restate already-documented requirements in preflight syntax, and the check names expose internal component and runtime detail that Embedded Cluster otherwise keeps out of the vendor-facing surface.

## What was already covered

Mapping the spec against existing docs, most of it is not new:

- the 14 port-availability checks are fully covered by the existing port requirements
- memory, CPU, cgroups v2, disk write latency, data directory space and permissions, and root access are in the system requirements
- API and proxy registry access are the firewall section

## What was missing

The host configuration a sysadmin has to set before installing, which is what this PR adds:

- kernel modules (`overlay`, `br_netfilter`, `nf_conntrack`) and the netfilter backend
- cgroup controllers
- required commands (`modprobe`, `mount`, `umount`)
- network `sysctl` parameters with their required values, including IP forwarding, bridge netfilter, ARP filter and ignore, and reverse path filtering
- inotify limits
- the 54-character hostname limit

Values are taken directly from the host preflight spec, so the checks and the docs describe the same requirements.

## Changes

- `docs/partials/embedded-cluster/v3/_host-config-reqs.mdx` (new)
- `embedded-cluster/installing-embedded-requirements.mdx` — adds the section between system requirements and port requirements
- `embedded-cluster/embedded-overview.mdx` — replaces the link, and updates the summary of what the checks cover so it matches the real groups

## Testing

`npm run build` succeeds. No headings were added or removed on existing pages, so no anchors changed. The two broken anchors the build reports near these pages are pre-existing and come from release notes.

### DIFF
--- a/docs/partials/embedded-cluster/v3/_host-config-reqs.mdx
+++ b/docs/partials/embedded-cluster/v3/_host-config-reqs.mdx
@@ -1,0 +1,53 @@
+Embedded Cluster verifies the following host configuration with host preflight checks during installation. Configuring these values before you install avoids failed installations.
+
+#### Kernel modules
+
+The following kernel modules must be loaded or loadable:
+
+* `overlay`
+* `br_netfilter`
+* `nf_conntrack`
+
+The host must also have a usable netfilter backend. This means that either the `ip_tables` module, for legacy iptables, or the `nf_tables` module, for nftables, is available.
+
+#### Cgroup controllers
+
+The following cgroup controllers must be available:
+
+`cpu`, `cpuacct`, `cpuset`, `memory`, `devices`, `freezer`, and `pids`
+
+#### Required commands
+
+The `modprobe`, `mount`, and `umount` commands must be available on the host.
+
+#### Network parameters
+
+The following `sysctl` parameters must be set as shown. To change a parameter, add it to `/etc/sysctl.conf` and then run `sudo sysctl -p`.
+
+| Parameter | Required value |
+|---|---|
+| `net.ipv4.ip_forward` | `1` |
+| `net.ipv4.conf.all.forwarding` | `1` |
+| `net.ipv4.conf.default.forwarding` | `1` |
+| `net.bridge.bridge-nf-call-iptables` | `1` |
+| `net.ipv4.conf.all.arp_filter` | `0` |
+| `net.ipv4.conf.default.arp_filter` | `0` |
+| `net.ipv4.conf.all.arp_ignore` | `0` |
+| `net.ipv4.conf.default.arp_ignore` | `0` |
+| `net.ipv4.conf.all.rp_filter` | `2` for loose mode, or `0` to disable |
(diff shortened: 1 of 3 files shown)


docs/partials/embedded-cluster/v3/_host-config-reqs.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+`cpu`, `cpuacct`, `cpuset`, `memory`, `devices`, `freezer`, and `pids`
+
+#### Required commands
+
+The `modprobe`, `mount`, and `umount` commands must be available on the host.
+
+#### Network parameters
+
+The following `sysctl` parameters must be set as shown. To change a parameter, add it to `/etc/sysctl.conf` and then run `sudo sysctl -p`.
```
sgalsaleh: hmm, idk where/how we should capture this, but EC does try to apply these configurations automatically. it has its own `/etc/sysctl.d/99-embedded-cluster.conf` config. it's best-effort though.

docs/partials/embedded-cluster/v3/_host-config-reqs.mdx
```diff
@@ -0,0 +1,53 @@
...
+| `net.ipv4.conf.all.arp_ignore` | `0` |
+| `net.ipv4.conf.default.arp_ignore` | `0` |
+| `net.ipv4.conf.all.rp_filter` | `2` for loose mode, or `0` to disable |
+| `net.ipv4.conf.default.rp_filter` | `2` for loose mode, or `0` to disable |
+
+Loose mode is preferred for reverse path filtering. Strict mode, which is the value `1`, drops pod traffic.
+
+#### Resource limits
+
+The following `sysctl` parameters must meet these minimums:
```
sgalsaleh: ditto

sgalsaleh (approved):

PR #4505, opened 2026-09-11
Document Troubleshoot CLI provenance verification

osc135:
## Summary

- add a Troubleshoot-specific page for verifying keyless SLSA provenance on `preflight` and `support-bundle` release archives beginning with version 0.134.1
- move the SBOM and image provenance pages from `/enterprise` to `/vendor` and add redirects for their published URLs
- repoint the existing Replicated SDK redirect directly to the relocated image provenance page
- update the security sidebar and cross-links for the vendor namespace
- clarify the provenance bundle `v` prefix, Windows `.zip` archives, placeholder substitution, and tested Cosign version

## Testing

- `npm run build`
- `git diff --check`
- `node --check sidebars.js`
- confirmed all three pages were generated under `/vendor`

The documentation build succeeds. It reports existing broken-link and broken-anchor warnings unrelated to these changes.

### DIFF
--- a/docs/enterprise/cli-provenance-validating.md
+++ b/docs/enterprise/cli-provenance-validating.md
@@ -1,12 +1,12 @@
-# Validate Troubleshoot CLI provenance
+# Validate CLI provenance
 
-This topic describes how to use Cosign to verify keyless Supply Chain Levels for Software Artifacts (SLSA) provenance for Troubleshoot CLI release archives.
+This topic describes how to use Cosign to verify keyless Supply Chain Levels for Software Artifacts (SLSA) provenance for Replicated CLI release archives.
 
-## About Troubleshoot CLI provenance
+## About CLI provenance
 
-Troubleshoot releases provide the `preflight` and `support-bundle` CLIs for multiple operating systems and architectures. A release that supports CLI provenance verification includes a `troubleshoot_VERSION_provenance.sigstore.json` bundle. The bundle contains signed SLSA provenance for all `preflight` and `support-bundle` archives in that release.
+Replicated CLI releases can include a Sigstore bundle containing signed SLSA provenance. The provenance associates each release archive's digest with the workflow and release tag that produced it. The keyless signature uses a short-lived certificate issued through the build system's OpenID Connect (OIDC) identity.
 
-The provenance associates each archive's digest with the Troubleshoot release workflow and release tag that produced it. The keyless signature uses a short-lived certificate issued through the GitHub Actions OpenID Connect (OIDC) identity.
+To verify a CLI archive, download the archive and its Sigstore bundle from the same release. Then, use Cosign to confirm that the archive's digest is included in the signed provenance and that the expected workflow identity produced the attestation.
 
 For information about validating Replicated container images, see [Validate image provenance](/enterprise/image-provenance-validating). For information about validating SBOM signatures, see [Validate SBOM signatures](/enterprise/sbom-validating).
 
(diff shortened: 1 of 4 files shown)


AmberAlston (changes requested):
Thanks for picking this up.

**1. These belong under `/vendor`, not `/enterprise`**

`docs/enterprise/` is specific to KOTS and kURL installations. Of its 42 files, everything is `installing-kurl*`, `snapshots-*`, `updating-app-manager`, `delete-admin-console`, and similar. It's tied to products we're retiring, so it isn't a namespace we plan to carry forward.

More importantly, it's the wrong audience. No enterprise ever reads these pages. A vendor reads them, either to write their own release instructions or to hand something to their customers. That's what `/vendor` is for, and it's the pattern the rest of our reference links already follow.

`sbom-validating.md` and `image-provenance-validating.md` are the only two files in `docs/enterprise/` that aren't KOTS or kURL docs. Dmitriy moved `image-provenance-validating.md` there on Aug 17 in #4375 and I didn't catch it at the time, but I don't want to keep the pattern going. Since you're already in these files and in this sidebar section, please move them as part of this PR rather than leaving a third page behind them.

Concretely:
- `git mv docs/enterprise/sbom-validating.md docs/vendor/sbom-validating.md`
- `git mv docs/enterprise/image-provenance-validating.md docs/vendor/image-provenance-validating.md`
- Land the new page under `docs/vendor/` as well
- Update the cross-link on line 11 of `sbom-validating.md`
- Update all four entries in the `sidebars.js` category so it's consistent
- In `netlify.toml`, add redirects from both old `/enterprise/` paths, and repoint the existing `/vendor/replicated-sdk-slsa-validating` rule directly at `/vendor/image-provenance-validating` rather than letting it chain through a redirect

The blast radius is small. There are only four internal references to those two pages in the whole repo, and no versioned-docs copies.

**2. Call this the Troubleshoot CLI, not the Replicated CLI**

To be direct about the ask, every generic mention of "CLI" or "Replicated CLI" on this page should say Troubleshoot CLI, or name `preflight` and `support-bundle` directly.

Here's why it matters. Replicated CLI is our product name for the `replicated` binary, which is the implementation of Vendor API v3 and the primary interface to the Replicated Platform. It has 143 reference pages under `docs/reference/replicated-cli-*` and its own install topic. So when a reader sees "Validate CLI provenance" in the security section, they will read it as that CLI, not as `preflight` and `support-bundle`.

That creates three problems. The claim isn't true of the thing it names, because the `replicated` CLI doesn't publish Sigstore bundles or SLSA provenance today, so a vendor would go looking for a bundle on a release that has none. The audience is inverted, because the Replicated CLI is a vendor tool authenticated with a vendor API token, while `preflight` and `support-bundle` are run by the vendor's customer inside their own environment. And it takes the title and URL we'd want later, when we do sign the `replicated` CLI.

What to change, specifically:

- **Filename:** `cli-provenance-validating.md` becomes `troubleshoot-cli-provenance-validating.md`, so the generic name stays free for the real Replicated CLI page later
- **Page title:** "Validate CLI provenance" becomes "Validate Troubleshoot CLI provenance"
- **First line:** "...for Replicated CLI release archives" becomes "...for Troubleshoot CLI release archives"
- **Section heading:** "About CLI provenance" becomes "About Troubleshoot CLI provenance"
- **First line of that section:** "Replicated CLI releases can include a Sigstore bundle" becomes "Troubleshoot releases include a Sigstore bundle." Drop "can," since every release from this version forward will have one
- **Second paragraph:** "To verify a CLI archive" becomes "To verify a Troubleshoot CLI archive"
- **Sidebar label** in `sidebars.js` should match the new title
- **The `## Validate Troubleshoot CLI provenance` heading** can then just be `## Validate a Troubleshoot CLI archive`, since the whole page is Troubleshoot now

The per-release detail in the existing Troubleshoot section is already written correctly. It's only the generic framing above it that needs rescoping.

**Smaller items**
- The release assets are inconsistent about the `v` prefix. Checksums publish as `troubleshoot_0.134.0_checksums.txt` with no `v`, while the provenance bundle will be `troubleshoot_v0.134.1_provenance.sigstore.json` with one. Your instructions are correct, but a parenthetical would save readers a failed attempt.
- Steps 3 and 4 should be a single step. Elsewhere we fold the substitution list into the step that runs the command. Also worth noting that VERSION appears twice, in the bundle filename and in `refs/tags/VERSION`.
- Windows archives are attested too. The example only shows `.tar.gz`, so one clause covering `.zip` would close that.
- "Cosign v3.1.3 or later" mirrors the version the release workflow pins for SBOM signing, which isn't the verification path. I'd phrase it as the version we tested rather than a hard floor.

AmberAlston (approved):
Updates look good thank you

PR #4519, opened 2026-09-15
docs: state where the persistent console stops and shell access begins in EC v3

AmberAlston:
Preview links
* https://deploy-preview-4519--replicated-docs.netlify.app/embedded-cluster/v3/embedded-persistent-console#console-scope
* https://deploy-preview-4519--replicated-docs.netlify.app/embedded-cluster/v3/embedded-cluster-shell

A vendor asked whether a customer with no shell access would still need kubectl
for Pod deletions and viewing Secrets. Neither the console page nor the shell
page could answer it, so it was answered by hand in Slack.

## The console page never stated its boundary

It listed four capabilities, and its Limitations section covered only deployment
constraints (single controller, not on a worker, TLS required).

It now explains the console as incremental capability rather than as a thing to
compare against the wizard. Install and upgrade already happen in a browser,
through the wizard the installer serves. Enabling the console adds a dashboard
that stays available afterwards, so an end user can come back later without
anyone opening a terminal. Upgrades and configuration changes started from the
console hand off to that same wizard rather than duplicating it.

It then states the boundary the original question was about. Neither surface
manages Kubernetes, and Pods, logs and Secrets need `shell`. Vendors who want to
give their customers cluster visibility are pointed at Helm chart extensions,
which is the standing recommendation and was previously only ever given in
support replies.

## The shell page was a flags table with no purpose

The page opened with "This topic describes the options available with the
Embedded Cluster `shell` command" and listed two flags. It never said this is the
supported route to kubectl, what the shell configures, or that it requires access
to a node.

Adds that, plus two examples. The specifics come from the command itself.
`KUBECONFIG` is set to the admin kubeconfig, `PATH` is extended with the bundled
tools, and the application namespace is the default context, so `kubectl get
pods` works without `-n`. Plain `kubectl` resolves because the wrapper is written
into the tools directory that is prepended to `PATH`.

## Verification

Console scope was read from the application's routes rather than from the
existing page, and against the released `3.12.0-beta.1` tag rather than `main`.
That distinction matters here: `main` carries two console features that are not
in any released build, and documenting from it would have described things
customers cannot see.


embedded-cluster/embedded-persistent-console.mdx
```diff
@@ -21,6 +21,24 @@ The initial application install and console setup require command-line access to
...
+
+* Check the installed version and channel, and open application links
+* See available updates and read release notes
+* Upload an air gap release archive
+* Generate a support bundle and download it
+* Start an upgrade, or change configuration values at the current version with **Redeploy**
+
+Upgrades and configuration changes started from the console hand off to the same wizard, so the console does not duplicate it. The practical difference is that nobody has to be at a terminal to begin.
+
+Neither surface manages Kubernetes. The console does not inspect or delete Pods, read container logs, or view Secrets. Those operations need command-line access to a node through [shell](embedded-cluster-shell).
```
sgalsaleh: hmm, i don't know if we can go as far as saying this, because the admin console interacts with our daemon systemd service that runs as root on the host, and that's how we upgrade the cluster and app or apply any changes. so technically, we have access to anything in the cluster, even host-level if we want to, via the daemon.

AmberAlston: Good catch, that was overclaiming. I'd conflated what the console exposes to a
user with what it's architecturally capable of — you're right that it drives the
daemon and could reach anything on the host.

Reworded to describe what it provides rather than what it can do:

"The console does not provide Kubernetes operations. There is no way to inspect
or delete Pods, read container logs, or view Secrets from it, so those need
command-line access to a node through shell."

The reader value I was after is just that a vendor shouldn't expect a Kubernetes
management UI, which that still carries without making a claim about the
daemon's reach. Pushed.

sgalsaleh (approved):

PR #4524, opened 2026-09-17
docs: document vendor redactors in Embedded Cluster v3

banjoh:
Vendor-supplied redactors are not a KOTS-only feature. Document them for Embedded Cluster v3.

- [Redactor](https://deploy-preview-4524--replicated-docs.netlify.app/reference/custom-resource-redactor) — drops the "KOTS Only" framing, adds where the resource goes per install method
- [Configure Embedded Cluster](https://deploy-preview-4524--replicated-docs.netlify.app/embedded-cluster/v3/embedded-using#redactors) — new "Add custom redactors" section: release-level and chart-packaged paths, template function syntax
- [Upgrade from Embedded Cluster v2](https://deploy-preview-4524--replicated-docs.netlify.app/embedded-cluster/v3/embedded-v3-migrate) — narrows "Support bundles are not affected" to spec versioning only
- [About Custom Resources](https://deploy-preview-4524--replicated-docs.netlify.app/reference/custom-resource-about) — redactors apply to support bundles, not preflight checks

Do not merge before the EC fix lands — the release-level path is documented as working.


docs/reference/custom-resource-redactor.md
```diff
@@ -25,9 +25,34 @@ spec:
...
+    <td>KOTS and Embedded Cluster v2</td>
+    <td>Add the Redactor resource to the release.</td>
+  </tr>
+  <tr>
+    <td>Embedded Cluster v3</td>
+    <td>Add the Redactor resource to the release, or package the spec in a labeled Secret or ConfigMap in your Helm chart. See [Add custom redactors](/embedded-cluster/v3/embedded-using#redactors) in <em>Configure Embedded Cluster</em>.</td>
+  </tr>
+  <tr>
+    <td>Helm</td>
+    <td>Package the spec in a Secret or ConfigMap in your chart's `templates/` directory. Label the resource `troubleshoot.sh/kind: support-bundle` and put the spec in the `redactor-spec` data key. The `kubectl support-bundle --load-cluster-specs` command discovers labeled resources in the cluster.</td>
```
AmberAlston: I'd recommend a Secret here rather than "a Secret or ConfigMap". Your own review round on https://github.com/replicatedhq/ec/pull/568 moved the release-level spec from a ConfigMap to a Secret because a rendered spec can carry password config values, and we have a release note recording ConfigMap storage as a fixed security issue in `docs/release-notes/rn-app-manager.md:65`. A Helm-templated redactor pulling from `.Values` has the same exposure. Same change applies to the Embedded Cluster v3 row on line 45.

embedded-cluster/embedded-using.mdx
```diff
@@ -258,10 +258,78 @@ This means bypassing a failed application check is a different decision from byp
...
+
+:::note
+Custom redactors apply to support bundles only. Preflight checks apply the built-in redactors and ignore Redactor resources.
+:::
+
+There are two ways to deliver a Redactor to an Embedded Cluster v3 installation:
+
+- **As a release-level resource:** Add the Redactor resource to your release alongside your HelmChart custom resources, the same way you did for installations with KOTS. Embedded Cluster publishes the spec into the cluster at install time and applies it to every support bundle. Releases that already define a Redactor need no change.
+
+- **From a Helm chart:** Package the spec in a Secret or ConfigMap in your chart's `templates/` directory. Label the resource `troubleshoot.sh/kind: support-bundle` and put the spec in the `redactor-spec` data key. Embedded Cluster discovers labeled resources in the cluster when it generates a support bundle. This is useful when a redactor belongs to a specific chart, or when you already ship support bundle specs this way.
```
AmberAlston: Same point as the Redactor reference page. I'd recommend a Secret and either drop ConfigMap or mark it as discouraged, since a rendered spec can carry credential values. The YAML example directly below already uses a Secret, so the prose is the only thing out of step.

embedded-cluster/embedded-using.mdx
```diff
@@ -258,10 +258,78 @@ This means bypassing a failed application check is a different decision from byp
...
+
+Redactors continue to use `v1beta2`. The `v1beta3` requirement applies only to preflight specs. See [Add preflight checks](#preflights) on this page.
+
+:::note
+Custom redactors apply to support bundles only. Preflight checks apply the built-in redactors and ignore Redactor resources.
+:::
+
+There are two ways to deliver a Redactor to an Embedded Cluster v3 installation:
+
+- **As a release-level resource:** Add the Redactor resource to your release alongside your HelmChart custom resources, the same way you did for installations with KOTS. Embedded Cluster publishes the spec into the cluster at install time and applies it to every support bundle. Releases that already define a Redactor need no change.
```
AmberAlston: "at install time" and "every support bundle" are both a little wide. The spec is republished on upgrade too, since `pkg/addons/upgrade.go` calls `EnsureVendorRedactorSecret` as well. And it applies only when the support bundle run can reach the cluster. `pkg/support/generate.go` always passes `--load-cluster-specs` but only adds `--kubeconfig` if one exists, so a host bundle collected against a broken cluster gets the built-in redactors only. Worth a sentence, since this is a redaction guarantee.

docs/reference/custom-resource-redactor.md
```diff
@@ -25,9 +25,34 @@ spec:
   redactors: []
 ```
 
-## Objects and Fields
+Redactors continue to use `troubleshoot.sh/v1beta2`. The `v1beta3` API version applies only to preflight specs in Embedded Cluster v3.
```
AmberAlston: This lands on a page read mostly by KOTS and Helm vendors, who have no frame for the v1beta3 migration. I'd move it into the Embedded Cluster v3 row of the table below.

docs/reference/custom-resource-redactor.md
```diff
@@ -1,20 +1,20 @@
...
 
-For Replicated KOTS installations, you can also add custom redactors to support bundles using the Redactor custom resource manifest file. For example, you can redact API keys or account numbers, depending on your customer needs. For more information about redactors, see [Redacting Data](https://troubleshoot.sh/docs/redact/) in the Troubleshoot documentation.
+You can also add custom redactors to support bundles using the Redactor custom resource manifest file. For example, you can redact API keys or account numbers, depending on your customer needs. For more information about redactors, see [Redacting Data](https://troubleshoot.sh/docs/redact/) in the Troubleshoot documentation.
 
-## Defining Custom Redactors
+## Defining custom redactors
```
AmberAlston: The Title Case to sentence case rewrite across this page is unrelated to the bug and makes the diff harder to read. There's no rule either way, since `.cursor/rules/docs-style-guide.mdc` is a stub and sibling reference pages are mostly Title Case. I'd drop it here or do it as its own pass.

AmberAlston (changes requested):
Checked the claims against the merged EC fix and troubleshoot. The mechanics are right. Secret name, namespace, label and key match `pkg/support/vendor_redactor.go`, the template rendering claim matches what KOTS does in `pkg/redact/app.go`, and the preflight statements hold up (`troubleshoot/pkg/preflight/collect.go:158` passes nil, and kots has no redact path in `pkg/preflight` at all).

Rest of my comments are inline. One item has no line here to anchor to, so it goes in this summary.

`docs/vendor/preflight-support-bundle-about.mdx` now contradicts this PR. Lines 31 to 39 frame redaction as a phase of both preflights and support bundles, and line 39 says custom redactors can be added for KOTS and Embedded Cluster installs, inside that shared section. That's the same defect you corrected on `custom-resource-about.md`, left standing on the page vendors hit first.

AmberAlston (approved):
Re-checked all five threads against a264052a. Everything is addressed.

The Secret-only guidance now carries the reason in both places, and there's no ConfigMap left in either file. The release-level bullet covers upgrade, and pulling the unreachable-cluster caveat out into its own callout after both bullets is better than what I asked for, since it applies to the chart-packaged path too. I confirmed `:::important` is registered in `docusaurus.config.js` and used widely elsewhere in the repo, and the caveat itself matches `pkg/support/generate.go`, which passes `--load-cluster-specs` unconditionally but only adds `--kubeconfig` when one exists.

The `preflight-support-bundle-about.mdx` rewrite reads well, and the link to the Redactor page is the first inbound prose link that page has had.

Approving.

PR #4525, opened 2026-09-17
docs(embedded-cluster): add SeaweedFS storage growth troubleshooting

banjoh:
Embedded Cluster v2 only. Documents the unbounded growth of the SeaweedFS filer change log under `/topics/.system/log` and the abandoned registry multipart uploads under `/buckets/registry/.uploads` on airgap HA installations.

Upgrading to a release with the maintenance script fix stops further growth, but SeaweedFS only reclaims the space that is already in use over time. The v2 page carries the one-time measure, purge, clean and vacuum remediation for existing clusters. v3 is in beta and will ship the corrected scripts before GA.

## Previews

- [Embedded Cluster v2 — Troubleshoot Embedded Cluster](https://deploy-preview-4525--replicated-docs.netlify.app/embedded-cluster/v2/embedded-troubleshooting#seaweedfs-storage-growth)


embedded-cluster_versioned_docs/version-2.0.0/embedded-troubleshooting.mdx
```diff
@@ -196,7 +196,81 @@ Reasons can include:
...
+kubectl exec -n seaweedfs seaweedfs-volume-0 -- du -sh /data
+```
+
+Where `/data` is the persistent volume claim that holds the volume files for that pod. Do not use `df` for this. The `openebs-hostpath` storage class does not enforce a quota, so `df` reports the usage of the entire node file system rather than the space that SeaweedFS uses.
+
+If you do not address the growth, the controller nodes eventually run out of disk space. This can make the cluster unhealthy and can block application upgrades.
+
+#### Cause
+
+There are two causes, and an affected cluster is usually accumulating data from both:
```
AmberAlston: This is stated harder than what we can support. The story has abandoned uploads as the leading candidate for the unexplained remainder but explicitly unconfirmed, says the collection cannot be determined from the bundle, and calls out `fs.du /buckets/registry/.uploads` as what would settle it. "Usually accumulating data from both" claims a prevalence nobody has measured. I'd frame uploads as a second contributor that may or may not be present, and keep the story's caution about not implying the purge recovers all of the reported growth.

embedded-cluster_versioned_docs/version-2.0.0/embedded-troubleshooting.mdx
```diff
@@ -196,7 +196,81 @@ Reasons can include:
...
+
+* **SeaweedFS filer change logs.** The SeaweedFS master runs a periodic maintenance cycle that includes `fs.meta.save` and `fs.meta.load`, which keep the metadata of the three filers in sync. Each run writes a complete metadata snapshot to the filer change log under `/topics/.system/log`. Because the cycle runs on a timer, the change log grows for the life of the cluster regardless of application activity. The amount written on each cycle scales with the number of files in the registry, so the growth rate increases as the registry grows.
```
AmberAlston: Optional. Worth giving the cadence here (`sleep_minutes = 17` in our values), since that's the number that lets an operator check their own growth rate against this explanation.

embedded-cluster_versioned_docs/version-2.0.0/embedded-troubleshooting.mdx
```diff
@@ -196,7 +196,81 @@ Reasons can include:
...
+
+* **Abandoned multipart uploads.** Image pushes to the built-in registry are multipart uploads. When a push is interrupted, the fragments that were already uploaded remain under `/buckets/registry/.uploads` and are never removed.
+
+Embedded Cluster releases that include the fix for this issue add `fs.log.purge` and `s3.clean.uploads` to the SeaweedFS maintenance cycle, which prevents further growth from both causes.
```
AmberAlston: Can we name the version here once it's cut? As written, a reader on an affected cluster can't tell whether they need to upgrade or are already on a fixed build, which is the first question this section raises.

embedded-cluster_versioned_docs/version-2.0.0/embedded-troubleshooting.mdx
```diff
@@ -196,7 +196,81 @@ Reasons can include:
...
+    :::
+
+    ```bash
+    kubectl exec -i -n seaweedfs seaweedfs-master-0 -- weed shell <<'EOF'
+    fs.log.purge -daysAgo 7
+    s3.clean.uploads -timeAgo=24h
+    EOF
+    ```
+
+1. Vacuum the volumes to return the space to the file system:
```
AmberAlston: The caution sits on the previous step, but this is the expensive one. At `-garbageThreshold 0.01` the vacuum rewrites essentially every volume carrying any garbage, under lock, and there's nothing here about how long that takes or what it costs in I/O. I'd move or duplicate the warning here. Related, the previous step justifies the 24 hour cutoff by saying pushes finish in minutes, which reads as though a concurrent push is in fact safe.

AmberAlston (changes requested):
Checked this against sc-139604 and both code fixes. The mechanics are right. Namespace and pod names match `pkg-new/constants/constants.go:5` and the `seaweedfs` release name, the `openebs-hostpath` and `df` caveat matches the dataDirs block in `values.tpl.yaml`, and both `fs.log.purge -daysAgo 7` and `s3.clean.uploads -timeAgo=24h` are in the shipped fix in both repos. "Upgrading stops further growth" holds because both fixes added the `maintenance-scripts-revision` pod annotation to force a master rollout, so AC 2 is covered in code.

Comments inline. The overstated cause is the one I'd hold on.

embedded-cluster_versioned_docs/version-2.0.0/embedded-troubleshooting.mdx
AmberAlston: Leaving this one open. The v2 fix (`2140a62d`) is on main but is not in any tag yet, and the newest release, `2.19.9+k8s-1.36`, predates it by a week, so there is nothing to name here. When can we expect to fill in the actual version number? I would like it in before merge, so the page is not telling operators to upgrade without saying to what.

AmberAlston:
Re-checked against 1452bb42. The cause hedge, the 17 minute cadence, and the vacuum cost warning are all addressed, and I've resolved those three threads. The vacuum rewrite is better than what I asked for, and moving the 24 hour cutoff so it explains why an upload is abandoned rather than why the step is safe cleared up the tension in that step.

One optional thing, not worth another round on its own. "the cluster holds a lock until it completes" describes the `weed shell` lock, which blocks other weed shell admin operations. The registry keeps serving during the vacuum, and I think a reader could take that sentence to mean the cluster is unavailable.

Left the version thread open, since there's nothing to name yet.

embedded-cluster_versioned_docs/version-2.0.0/embedded-troubleshooting.mdx
banjoh: We can cut a `2.19.10` release today

AmberAlston (approved):

PR #4527, opened 2026-09-17
EC v3: frame the v2 to v3 page for vendors deciding, not just migrating

AmberAlston:
Preview links
* https://deploy-preview-4527--replicated-docs.netlify.app/embedded-cluster/v3/embedded-v3-migrate
* https://deploy-preview-4527--replicated-docs.netlify.app/embedded-cluster/v3/embedded-config#example
* https://deploy-preview-4527--replicated-docs.netlify.app/embedded-cluster/v3/embedded-overview#ec-limitations

## Why

Three vendor questions this week that the docs have no good answer for:

- What do we have that shows customers the difference between EC v2 and v3?
  The best thing we have is the comparison section on the v2-to-v3 page, but
  it is written as "what's different," not "here's why you'd want to upgrade,"
  and it is vendor-facing release-authoring content that can't be forwarded to
  an end customer as-is.
- What does the `-beta` designation mean while an application release sits in
  the Stable channel? Asked by a vendor whose architecture and security teams
  pick up on it during review.
- Which EC version should we be on? Asked by a vendor pinned to
  `3.0.0-beta.1`, which is what our own Config examples still showed.

Docs are procedural, not marketing, and this PR keeps that line. The addition
is decision-support content, which is a reference question, not persuasion.
Every statement is a mechanism with a documented consequence and a link to the
page that proves it, and the section intro points at the release work required
so the gain is never stated without the cost beside it. No adjectives are doing
any work, and there are no new claims: everything in the new sections is
behavior already documented elsewhere in the v3 docs, pulled together and
framed for a reader making a decision rather than executing a migration.

## What

**Two new sections on the v2-to-v3 page, above the existing comparison.**

- *Benefits of Embedded Cluster v3.* Seven changes stated as mechanism and
  consequence. Four are facts already on this page or in the overview,
  currently filed under what has to change rather than what it buys you.
- *Changes for end users.* End-user visible behavior, split into one-time
  upgrade experience and ongoing v3 behavior so vendors don't describe a
  migration prompt as something customers hit every time. Says up front that
  vendors can adapt it for their own customer communications, which is the use
  case that prompted it.

Also removes the one-sentence `Installer wizard differences` subsection, now
covered in the end user list. Nothing linked to the anchor.

**Beta label linked to its definition.** The overview said only that v3 is Beta
and subject to change, which is the sentence that gets escalated in a security
review and answers nothing. The support lifecycle policy already defines Beta
as typically production-ready and following the same build and test processes
required for GA. Linked to it rather than restating it and risking drift. One
sentence added, no new commitments authored.

**Config examples no longer teach a stale version.** They pinned
`3.0.0-beta.1+k8s-1.34`, the first release of the v3 line, directly under a
migration step that says to update `version` to the latest. Replaced with an
`EC_VERSION` placeholder, matching the `APP_SLUG` convention already used in
the install docs, plus a pointer to the release notes. The placeholder does not
go stale. Also fixed the `+k8s` suffix illustration, which used a version
predating the beta line entirely, and a link on the v3 Config page that pointed
at the v2 release notes.

## Note for review

The end user section says the guided wizard "replaces the KOTS Admin Console
that handled these steps in v2." That is the sentence vendors will react to,
since some of them liked the always-on console. Worth a second opinion on the
wording from someone on the EC side.

`npm run build` passes. The three broken-link warnings are pre-existing on
pages this PR does not touch.


embedded-cluster/embedded-config.mdx
```diff
@@ -12,13 +12,15 @@ This topic is a reference for the Replicated Embedded Cluster Config custom reso
 
 ## Example
 
+In the following example, replace `EC_VERSION` with the version of Embedded Cluster to install. For the available versions, see [Embedded Cluster v3 release notes](/release-notes/rn-embedded-cluster-v3).
```
sgalsaleh: the problem with the release notes versions is that they do not include the `+k8s-1.34` suffix, and i'm sure some vendors are going to miss that.

embedded-cluster/embedded-v3-migrate.mdx
```diff
@@ -7,6 +7,60 @@ This page describes how to upgrade from Embedded Cluster v2 to Embedded Cluster
 
 It includes information about how to update your application release to support Embedded Cluster v3. It also describes how to upgrade existing installations from Embedded Cluster v2 to v3.
 
+## Benefits of Embedded Cluster v3
+
+Embedded Cluster v3 delivers the following improvements over v2. For the release changes required to adopt them, see [Comparison to Embedded Cluster v2](#comparison-to-embedded-cluster-v2).
+
+**One guided experience from the first step.** In v2, cluster infrastructure was set up on the command line before a separate UI handled the application. In v3, a single binary drives cluster configuration, preflight checks, node setup, and application deployment in one wizard. See [Embedded Cluster overview](embedded-overview).
+
+**The cluster and the application update together.** Embedded Cluster updates the cluster infrastructure alongside application updates, so end users keep both current without using `kubectl`. See [Perform updates in embedded clusters](updating-embedded).
```
sgalsaleh: this is technically true for EC v2 as well.

also, i don't understand what `so end users keep both current without using kubectl` means. we never required users to run kubectl to deploy the app, not even in kots. what are we comparing v3 to here?

embedded-cluster/embedded-v3-migrate.mdx
```diff
@@ -7,6 +7,60 @@ This page describes how to upgrade from Embedded Cluster v2 to Embedded Cluster
...
+
+**The upgrade is one-way and asks for confirmation.** The v3 binary detects the existing v2 installation and prompts the user to confirm before it proceeds. There's no path back to v2.
+
+**Configuration carries forward.** Values the customer already set, such as the application domain, are preserved, so nothing has to be re-entered.
+
+**Customers more than one Kubernetes minor version behind need an intermediate release first.** Kubernetes supports upgrading only one minor version at a time, and Embedded Cluster blocks an upgrade that would skip one. See [Increase the Kubernetes minor version](embedded-config#increment-k8s) for how to lead customers through the versions in order.
+
+**Port 30080/TCP must be open on the node.** The Embedded Cluster UI requires it, and it must also be reachable by nodes joining the cluster. Customers upgrading an existing v2 node may need to create a firewall opening for it. If the port is occupied, they can select a different port during installation. See [Port requirements](installing-embedded-requirements#port-requirements).
```
sgalsaleh: we can also mention that daemon api port, which is `30081` by default and can be overridden by `--api-port`.

sgalsaleh: it's probably only needed for multi-node clusters though

embedded-cluster/embedded-v3-migrate.mdx
```diff
@@ -7,6 +7,60 @@ This page describes how to upgrade from Embedded Cluster v2 to Embedded Cluster
...
+
+**An always-on console is available when you want one.** The [persistent admin console](embedded-persistent-console) runs as an opt-in system service, giving end users a browser dashboard without command-line access. From it they can see available updates, view release notes, trigger an upgrade, upload air gap releases, and generate support bundles. It isn't installed by default, and the initial install and console setup require command-line access.
+
+**How customers find out about updates depends on whether the console is installed.** With the persistent admin console, available updates appear in the console. Without it, customers get updates from the [Enterprise Portal](/vendor/enterprise-portal-about) or by running the `upgrade` command directly. See [Perform updates in embedded clusters](updating-embedded).
+
+**Multi-node clusters upgrade one node at a time.** After the first node upgrades, the wizard guides the user through upgrading each remaining node.
```
sgalsaleh: > the wizard guides the user through upgrading each remaining node.

the wizard will actually automatically upgrade the remaining nodes. it only shows guidance or manual commands if it fails to do so

embedded-cluster/embedded-config.mdx
kriscoleman: yea, my agents and vendor agents have stumbled around that too, and I've had to update our skills to nudge the agents about this. 

Vendors not using these skills with their agents hit this. If there were a clear call out in the docs about it it may correct that.

embedded-cluster/embedded-overview.mdx
```diff
@@ -11,7 +11,7 @@ This topic provides an introduction to Replicated Embedded Cluster.
 
 ## Limitations {#ec-limitations}
 
-* Embedded Cluster v3 is Beta. The limitations and features are subject to change.
+* Embedded Cluster v3 is Beta. The limitations and features are subject to change. For Replicated's definition of Beta, see [Support lifecycle policy](/vendor/policies-support-lifecycle).
```
kriscoleman, line 14: great addition, this comes up a lot lately

embedded-cluster/embedded-config.mdx
AmberAlston: Good catch, that would have bitten people. Fixed in 14f5a82 — the guidance now says the Kubernetes suffix is part of the value, shows `3.12.0-beta.1+k8s-1.36` as an example, and points at the per-release version table rather than just the release notes page. Same change in all three places the Config example appears.

embedded-cluster/embedded-v3-migrate.mdx
AmberAlston: You're right on both counts, so I dropped the bullet. It's true of v2, so it doesn't belong in a list of what v3 adds. And the "without using kubectl" clause was comparing v3 to nothing real — I lifted that sentence from the Embedded Cluster definition partial, where it's describing EC generally versus hand-managing a cluster, and it picked up a claim it wasn't making once I moved it into a v2-versus-v3 list.

embedded-cluster/embedded-v3-migrate.mdx
AmberAlston: Added, scoped to multi-node per your follow-up, with `--api-port` for overriding it. Also named `--port` for the UI port while I was in there.

30081 also wasn't in the port requirements partial at all, only on the External API page and in the install flag table, so I added it there too in bc90fbc. That's the page customers actually use to plan firewall openings.

One thing to check me on: I scoped it as multi-node needing openings between nodes, and single-node only needing it if you drive the API from off-host, since the External API page mentions headless installs and programmatic joins. If single-node never needs it exposed, I'll simplify.

embedded-cluster/embedded-v3-migrate.mdx
AmberAlston: Fixed, thanks — I'd described the failure path as the normal one. It now says the primary controller's daemon streams the upgrade to each remaining node, controllers first to preserve quorum, and that manual instructions only appear when a node can't be reached or predates streamed upgrades.

One question that affects the section above it. The docs say the orchestrator can't stream to nodes running EC earlier than 3.2. In a v2 to v3 upgrade every remaining node is earlier than 3.2, so does that case fall to the manual per-node path? The upgrade demo showed it that way. If so I'll say that explicitly under "During the upgrade from v2", since it'd be the opposite of what ongoing v3 upgrades do.

embedded-cluster/embedded-v3-migrate.mdx
sgalsaleh: i can't think of a use case where we need it for single-node, specially because the daemon listens on a unix socket on the host

kriscoleman: I haven't double checked this so I could be wrong, but is Port 30080 actually required?
Thinking that if automating with the new headless features, the UI isn't needed, so that port may not need to be open on the node. In which case the "Port 30080/TCP _must_ be open on the node" might be only true if actually using the UI.

sgalsaleh: for the ui port, i believe the flag is --installer-port

embedded-cluster/embedded-v3-migrate.mdx
```diff
@@ -7,6 +7,60 @@ This page describes how to upgrade from Embedded Cluster v2 to Embedded Cluster
...
+
+**The upgrade is one-way and asks for confirmation.** The v3 binary detects the existing v2 installation and prompts the user to confirm before it proceeds. There's no path back to v2.
+
+**Configuration carries forward.** Values the customer already set, such as the application domain, are preserved, so nothing has to be re-entered.
+
+**Customers more than one Kubernetes minor version behind need an intermediate release first.** Kubernetes supports upgrading only one minor version at a time, and Embedded Cluster blocks an upgrade that would skip one. See [Increase the Kubernetes minor version](embedded-config#increment-k8s) for how to lead customers through the versions in order.
+
+**Port 30080/TCP must be open on the node.** The Embedded Cluster UI requires it, and it must also be reachable by nodes joining the cluster. Customers upgrading an existing v2 node may need to create a firewall opening for it. If the port is occupied, they can select a different port during installation with `--port`. See [Port requirements](installing-embedded-requirements#port-requirements).
```
sgalsaleh:
```suggestion
**Port 30080/TCP must be open on the node.** The Embedded Cluster UI requires it, and it must also be reachable by nodes joining the cluster. Customers upgrading an existing v2 node may need to create a firewall opening for it. If the port is occupied, they can select a different port during installation with `--installer-port`. See [Port requirements](installing-embedded-requirements#port-requirements).
```

embedded-cluster/embedded-v3-migrate.mdx
```diff
@@ -7,6 +7,60 @@ This page describes how to upgrade from Embedded Cluster v2 to Embedded Cluster
...
+
+**Customers more than one Kubernetes minor version behind need an intermediate release first.** Kubernetes supports upgrading only one minor version at a time, and Embedded Cluster blocks an upgrade that would skip one. See [Increase the Kubernetes minor version](embedded-config#increment-k8s) for how to lead customers through the versions in order.
+
+**Port 30080/TCP must be open on the node.** The Embedded Cluster UI requires it, and it must also be reachable by nodes joining the cluster. Customers upgrading an existing v2 node may need to create a firewall opening for it. If the port is occupied, they can select a different port during installation. See [Port requirements](installing-embedded-requirements#port-requirements).
+
+### On any v3 installation
+
+These are the ongoing differences after the upgrade.
+
+**A guided wizard handles install and upgrade.** Install and upgrade operations launch a wizard that walks the user through configuration, preflight checks, node setup, and deployment, and closes when the operation finishes. It replaces the KOTS Admin Console that handled these steps in v2.
```
kriscoleman, line 52: I think this is fine, but I wanted to offer some food for thought: 

the wizards do close when the operation finishes; but I've noticed that k8s SMEs love it when I add "so unlike v2, the admin install/upgrade consoles are removed _so they don't take any resources from the host when they aren't needed_."

kriscoleman:
I'm not sure if it's ready to put in here yet, but I know DR comes up a lot when vendors ask us about EC v3. 

It would be great to add a section in here about the DR differences/improvements over EC v2 when the time comes.

embedded-cluster/embedded-v3-migrate.mdx
AmberAlston: accepted your suggested edit

AmberAlston:
> I'm not sure if it's ready to put in here yet, but I know DR comes up a lot when vendors ask us about EC v3.
> 
> It would be great to add a section in here about the DR differences/improvements over EC v2 when the time comes.

Agree. I'm waiting until this is available

embedded-cluster/embedded-v3-migrate.mdx
AmberAlston: Took this, thanks. I put it in the Benefits section rather than here, since it's a v2-versus-v3 gain rather than something an end user notices, and it folds into the footprint bullet that was already there:

"The install and upgrade wizard runs only while an operation is in progress, so no management console runs continuously in the cluster unless a customer opts into the persistent admin console."

I scoped it to the in-cluster console rather than host resources generally. The daemon does run persistently, since it serves the API on 30081, streams upgrades to other nodes, and reports lifecycle events, so "doesn't take any resources from the host when they aren't needed" would overstate it. The v2 comparison still holds for the part operators care about, which is that the console isn't sitting in their cluster between operations.

embedded-cluster/embedded-v3-migrate.mdx
AmberAlston: Answered this from the code so as not to hold up the PR.

Confirmed: a multi-node v2 to v3 upgrade does fall to the manual per-node path. `MinOrchestratedUpgradeVersion` is `3.2.0-beta.1`, and the comment in `pkg/upgradeorchestrator/capabilities.go` names the case directly, that v2 nodes whose EC version label is absent force manual upgrades. `classifyNode` treats an empty `ECVersion` as a blocker and `DecideMode` switches the whole run to manual as soon as one exists, which `capabilities_test.go` covers as the "v2-like" case.

So it's the guaranteed path for every multi-node migration rather than an edge case. Added a bullet under "During the upgrade from v2" in a7a30ca saying the wizard detects it and shows manual instructions, and that it stops applying once every node is on v3. Correct me if I've read that wrong.

AmberAlston: I think i self answered this from the code. Added https://github.com/replicatedhq/replicated-docs/pull/4527/changes/a7a30ca2360a62566bd57e1eb1dd3b6f29fb4d23#diff-363153e180631626803e9d38efd17f2fc622c7a0f6bdfd5f51eadd8974d7398aR42

sgalsaleh (approved):

PR #4529, opened 2026-09-17
Application slug alias docs

divolgin:


AmberAlston (approved):

docs/vendor/application-slug-aliases.md
```diff
@@ -14,10 +14,22 @@ Changing the active slug applies to all customers for the application. It does n
...
+**Helm**
+
+- Both the primary slug and an alias can be used to pull charts. Newly generated Helm install instructions use the active slug.
+
+**KOTS**
 
 - New installations can use an application slug alias with any version of KOTS.
-- Switching an existing installation from the primary slug to an alias requires KOTS v1.132.0 or later.
+- On KOTS v1.132.0 and later, existing installations pick up a new active slug automatically on their next license sync. The Admin Console URL for the application changes to use the new slug, and existing Admin Console URLs redirect to it.
+- On KOTS versions earlier than v1.132.0, existing installations are unaffected and continue to use the slug they were installed with. The new active slug is ignored rather than applied, and no error is reported.
```
divolgin: This is not entirely true. Previous releases fail license validation if application slug doesn't match.

AmberAlston: Thanks, that's my error. I read `Sync` and generalized from it, but I can see the airgap paths do fail on a slug mismatch, and #6061 dropping the `airgap.Spec.AppSlug != license.GetAppSlug()` check in `update.go` makes that concrete.

I've reworded it to say that earlier versions do not adopt the new slug and that operations comparing the slug against the license can fail.

